PR #157, opened 2013-09-04
Foreign Function Interface with Jasmine testing framework.

david-geo-holmes:


bnmnetp:
Doing some house cleaning, and I think we have left this PR in the past.

PR #237, opened 2014-06-04
Added ability to load libraries from an external source dynamically at compile time

bzwheeler:
Added loadExternalLibrary method to the import process which
will check an optional Sk.externalLibraries config for info on
where to find the library

Example usage would be:

``` js
Sk.externalLibraries = {
  pygal : {
    path : '/js/skulpt/pygal/__init__.js',
      dependencies : [
        '/components/highcharts/highcharts.js',
        '/components/highcharts/highcharts-more.js'
      ]
  },
  // ... other libraries
};
```

This allows us to dynamically load a library whenver the 'import ...' statement is compiled which means it does not have to be included in the standard library included with skulpt.

The path should specify the url at which the library can be found and can be either a .js builtin or a .py file which wil be compiled by skulpt after it loads.  You can also specify a depencies array which is a list of javascript libraries that must be loaded for the library to work.  These will also be dynamically loaded at the appropriate time.

As a shorthand in cases where there are no depenencies you can also have a format of:

``` js
Sk.externalLibraries = {
  pygal : '/js/skulpt/pygal/__init__.js',
  // ... other libraries
}
```


waywaaard:
just wanted to add that this works with my numpy and matplotlib/pyplot libs:

```

        Sk.externalLibraries = {
            numpy : {
                path: '/static/primeronoo/skulpt/external/numpy/__init__.js'
            },
            matplotlib : {
                path: '/static/primeronoo/skulpt/external/matplotlib/__init__.js'
            },
            "matplotlib.pyplot" : {
                path: '/static/primeronoo/skulpt/external/matplotlib/pyplot/__init__.js'
            }
        }
```

see pyplot must be included in quotes

PR #243, opened 2014-06-11
Improves time module

waywaaard:
Adds bunch of new methods to the time module. However the format functions are missing, so still wip.

Adds:
- struct_time
- localtime
- mktime
- asctime
- gmtime
- ctime
- sleep


waywaaard:
This still needs some more love and work. Have to figure out how to implement or mock the locale timezones and timezone formating. But I am currently working on the strftime function.

Todo:
- [ ] tzset
- [ ] tzname
- [ ] timezone
- [ ] strptime -  needs some research
- [ ] strftime - in work
- [x] clock - return current processor time using browser performance timer
- [ ] daylight
- [ ] altzone

waywaaard:
This needs some more work than I've expected, especially implementing ```struct_time``` correctly so that ````t.hours``` and etc. are working.

Jacco:
Oops didn't see this pullrequest before I added mine. Maybe we can combine some features?

waywaaard:
Of course, we can though I stopped there due to some problems and missing time. I think we should implement a proper named tuple. I know now much more about the inner of skulpt. Built-ins are not really subclassable currently -.-. So we may need to copy a few things from tuple or use javascript prototypes.

Jacco:
Ah ok. I just went with using plain tuples in my implementation. Shall I go ahead and pluck the cherries from your pull request? Or what do you think is the route to follow?

waywaaard:
My implementation followed the c implementation. But I think we should start implementing the named tuple. If you need a few things from the PR you can of course use them.

Jacco:
How would you approach that? Making named_tuple.js next to tuple.js? But you can't inherited from such a class right? How would you fix repr to show the correct classname etc?

waywaaard:
Well we could wait until builtins are subclassable (I guess this will take some time) or we use the same approach as Exceptions currently. 

```goog.inherits(mod.struct_time, Sk.builtin.tuple);```

and add a custom get|set attribute function for the specific struct time names.

Jacco:
Added this to tuple.js for now. Needs more work obviously, but do you think this is the right direction?
 
    Sk.builtin.namedtuples = {};

    Sk.builtin.make_namedtuple = function (module, name, fields) {
        var nm = module + "." + name;

        var cons = function nametuple_constructor(args)
       {
            var o;
            if (!(this instanceof Sk.builtin.namedtuples[nm])) {
                o = Object.create(Sk.builtin.namedtuples[nm].prototype);
                o.constructor.apply(o, arguments);
                return o;
            }
            Sk.builtin.tuple.apply(this, arguments);

            this.__class__ = Sk.builtin.namedtuples[nm];        
        };
        Sk.builtin.namedtuples[nm] = cons;

        goog.inherits(cons, Sk.builtin.tuple);
        cons.prototype.tp$name = nm;
        cons.prototype.ob$type = Sk.builtin.type.makeIntoTypeObj(nm, Sk.builtin.namedtuples[nm]);
        return cons;
    };
    goog.exportSymbol("Sk.builtin.make_namedtuple", Sk.builtin.make_namedtuple);

waywaaard:
Seems to be fine for me. I like that you actually tried to implement the named tuples. Maybe we can expose this function to the collections module.

rixner:
Don't named tuples belong in the collections module, not in tuple.js?  It shouldn't be that hard to do that (without having anything else from the collections module there).

Jacco:
Good idea. I added some more changes to the code above and have got it in a state where all unittests from cpython on struct_time work correctly. Shall I post (or edit) the code here so you can review it?

By collections you mean

https://github.com/skulpt/skulpt/blob/master/src/lib/collections.py

right?

Jacco:
tuple.js would indeed not be the right place. But it needs to be in a core js file I think. In python 2.7 they do it in collections.py using the following:

    namespace = dict(_itemgetter=_itemgetter, __name__='namedtuple_%s' % typename,
                     OrderedDict=OrderedDict, _property=property, _tuple=tuple)
    try:
        exec class_definition in namespace
    except SyntaxError as e:
        raise SyntaxError(e.message + ':\n' + class_definition)
    result = namespace[typename]

But I do not think we support that any time soon.

rixner:
They are implementing it as an ```OrderedDict```, which is also defined in collections:

https://github.com/python/cpython/blob/2.7/Lib/collections.py

So, I still think they both belong in the collections library, not in the core of Skulpt.  This also suggests that the right way to do this is to implement ```OrderedDict``` first.

bnmnetp:
replaced by #407

PR #276, opened 2014-08-11
Make stub files for all unimplemented stdlib modules

bnmnetp:
If you try to import a module that is in the 2.6 standard library you will now get a not implemented error on import.

All of the files were generated using the script stdlibstubs.py   This doesn't need to be part of the final PR, but you can see how all of these were generated.

This of course increases the size of skulpt-stdlib.js (by roughly 100kb)

Let me know what you think.

### DIFF
--- a/src/lib/test/__init__.py
+++ b/src/lib/test/__init__.py
@@ -1,0 +1,21 @@
+__author__ = 'bmiller'
+
+def testEqual(actual, expected):
+    if type(expected) == type(1):
+        if actual == expected:
+            print('Pass')
+            return True
+    elif type(expected) == type(1.11):
+        if abs(actual-expected) < 0.00001:
+            print('Pass')
+            return True
+    else:
+        if actual == expected:
+            print('Pass')
+            return True
+    print('Test Failed: expected ' + str(expected) + ' but got ' + str(actual))
+    return False
+
+def testNotEqual(actual, expected):
+    pass
+


mentalisttraceur:
I wish I had been here seven years ago to say:

Ahhh! Don't change the raised error type from `ImportError` to anything else! That will break literally any code which gracefully degrades in the absence of a module! (Unless you subclass `ImportError` and raise that subclass - that's harmless.)

The whole point of having import errors be typed as import errors is so that code can do stuff like this:

```python
try:
    from threading import local
     _threadlocal_state = local()
except ImportError:
    # degrade to works in single-threaded environments
    class _DummyO(object):
        pass
    _threadlocal_state = _Object()

...
```
or
```python
try:
    from reprlib import recursive_repr
    _recursive_repr_if_available = recursive_repr()
except ImportError:
    def _recursive_repr_if_available(decorated_function):
        return decorated_function

...

class SomeClass(object):
    ...
    @recursive_repr_if_available
    def __repr__(self):
        ...
```

If the error type is changed, all code already written to handle missing modules goes from working to unusably broken.

Because of this change, maximally portable code needs to catch two possible errors on every import instead of just one, if it wants to run without modification on Skulpt. Though I imagine almost every self-respecting developer solves that problem by deciding "my module just isn't going to work on Skulpt then, oh well" - I'm the only person I know neurotic enough to feel compelled to make sure users of my stuff can run it on Python implementations like Skulpt without modification.

Anyway, if it were me, I'd keep the very helpful error messages as-is, but change the error types back to `ImportError`. If someone actually has a use-case for _programmatically_ responding to these import errors differently than other ones, I'd `class ImportNotImplemented(ImportError): pass`. But since by now some Skulpt code out there might have been written to rely on catching the current error type, I'd `class ImportNotImplemented(ImportError, NotImplementedError): pass` - that way you restore compatibility with the standard and retain compatibility with how Skulpt's been doing it for seven years.

PR #310, opened 2014-11-07
Cpython unit tests

bnmnetp:
Create PR to merge the cpython unit test feature branch back to master.

Much work to do here!


eah13:
@bnmnetp friendly status check on this.  I added CPython unit tests to #400 so the concept won't get lost regardless.  

Will you take PRP on this or is it up for grabs?  If the latter, maybe worth documenting where you're at and/or merging what's usable.

bnmnetp:
Two things:

1.  I think we have a reasonable workflow/workaround to get the core elements of many of the tests working with just some wood chopping to do.

2.  Many other dependencies stand in the way of true compatibility.  Not small ones either.  Inheritance from builtin types, working decorators are two that immediately come to mind.

It is easy enough for someone implementing a new feature to use the core of the existing unit tests with some of simplification, that I think for now our efforts are better served on other things.

It would be interesting to do an analysis of the biggest bang for the buck improvements.  But my instinct is that the two things I mentioned (Inheritance and decorators) are the winners in terms of our ability to make a lot more of the pure python stdlib modules viable.  Followed closely by sys and a really good implementation of locals.

eah13:
Roger that.  I'll add decorators and locals to the list.  Anything else that belongs on #400 please feel free to add it directly.

Should this PR get closed then?

bnmnetp:
I think it is still a moderately useful long term goal, and it serves the purpose of uncovering these “advanced use cases” that need to be addressed.  So, I would say it is reasonable to keep it open.

PR #318, opened 2014-12-29
fixes turtle heading for negative radius turtle.circle() calls

bzwheeler:
When a negative radius is sent to `turtle.circle()`, the final heading of the turtle was being improperly calculated.  This PR properly accounts for the negative value.

### DIFF
--- a/src/lib/turtle/__init__.js
+++ b/src/lib/turtle/__init__.js
@@ -765,15 +765,19 @@ return (function() {
             w  = extent / steps;
             w2 = 0.5 * w;
             l  = radius * Math.sin(w/self._fullCircle*Turtle.RADIANS);
+
             if (radius < 0) {
                 l = -l;
                 w = -w;
                 w2 = -w2;
+                endAngle = angle - extent;
+            }
+            else {
+                endAngle = angle + extent;
             }
 
             promise = getFrameManager().willRenderNext() ? Promise.resolve() : new InstantPromise();
 
-            endAngle = angle + extent;
             angle += w2;
 
             for(i = 0; i < steps; i++) {


bnmnetp:
Hi @bzwheeler,

I'm working on integrating the new turtle back into the Runestone framework.  And I think, because of the way we move some elements around when the run button is clicked we are getting canvases that are overlapping but not in sync with one another.  Giving me a scene like this:

screenshot 2015-02-05 16 25 37

I updated the main skulpt index.html page today and had no such trouble, so I think it must have to do with our "canvas" always starting out hidden, and how we are positioning that canvas before running.  Or it may simply be the change from me supplying a canvas tag to your change over to just having a div and adding the canvas in Javascript.

Anyway, I'm hoping this might jog something in your memory that would help me get this fixed up quickly.

Thanks,

Brad

bzwheeler:
Hmm... Not sure if this will help but in our trinkets the output area is initially hidden and we configure it on-demand by using an onBeforeImport like this:

```js
Sk.onBeforeImport = function(library) {
  if (library === "turtle") {
    if (!Sk.TurtleGraphics) {
      Sk.TurtleGraphics = {};
    }
    Sk.TurtleGraphics.width  = 400;
    Sk.TurtleGraphics.height = 400;
    // tell turtle to create canvases in the #graphic-output div
    Sk.TurtleGraphics.target = $("#graphic");
  }
}
```

If your code is running on some environment that I can access, I'd be happy to try and help out more by diving in and playing around with the code.

bnmnetp:
I can work on making the environment available to you.  In the meantime I have discovered the following:

1.  I need to give the div some size.  400x400 otherwise the canvases do not show at all!??
2.  Somewhere along the way, some bit of code decides that the canvases should be 320 by 320 even using your hint from above to set height and width to 400.
3.  The second canvas has a margin of -400

However,  If I change the size of the div to 320x320 and the hints to 320x320 then everything lines up again.

bnmnetp:
OK, I tracked this down to some CSS, I think I have a handle on this now.  

Thanks.

bzwheeler:
Sorry about that! Glad you got it all worked out.

PR #319, opened 2015-01-05
String functions

gerbal:
Implemented:

```
str.expandtabs()
str.isspace()
str.istitle()
str.splitlines()
str.swapcase()
str.title()
str.zfill(width)
```

based largely on [pypy implementations](https://github.com/woodrow/pyoac/blob/master/pypy/objspace/std/stringobject.py).  Unittests are based on cPython [buffer unittests](https://github.com/python/cpython/blob/master/Lib/test/buffer_tests.py). 

Still need to implement string.format, and get Skulpt to pass cpython [string_tests.py](https://github.com/python/cpython/blob/master/Lib/test/string_tests.py) but that's a whole other project.

### DIFF
--- a/src/str.js
+++ b/src/str.js
@@ -773,7 +773,14 @@ Sk.builtin.str.prototype["isspace"] = new Sk.builtin.func(function (self) {
     return Sk.builtin.bool(true);
 });
 
+
 Sk.builtin.str.prototype["expandtabs"] = new Sk.builtin.func(function (self, tabsize) {
+    var input = self.v;
+    var expanded = "";
+    var split;
+    var spacestr = "";
+    var spacerem;
+
     Sk.builtin.pyCheckArgs("expandtabs", arguments, 1, 2);
     if ((tabsize !== undefined) && !Sk.builtin.checkInt(tabsize)) {
         throw new Sk.builtin.TypeError("integer argument exepcted, got " + Sk.abstr.typeName(tabsize));
@@ -783,10 +790,25 @@ Sk.builtin.str.prototype["expandtabs"] = new Sk.builtin.func(function (self, tab
     }else{
         tabsize = Sk.builtin.asnum$(tabsize);
     }
+    split = input.split("\t");
 
-    var spacestr = goog.string.repeat(" ", tabsize);
+    for(var x = 0; x<split.length; x++){
+        if (split[x].match('/\n|\r/g') !== undefined){
+            spacerem = tabsize - ((split[x].length - (Math.max([split[x].lastIndexOf('\r'),split[x].lastIndexOf('\n')]))+2) % tabsize);
+        } else{
+            if(Math.max([expanded.lastIndexOf('\r'),expanded.lastIndexOf('\n')]) !== -1){
+                spacerem = tabsize - ((expanded.length - (Math.max([expanded.lastIndexOf('\r'),expanded.lastIndexOf('\n')]))+2) % tabsize);
+            } else {
+                spacerem = tabsize - (expanded.length % tabsize);
+            }
+        }
+        spacestr = goog.string.repeat(" ", spacerem);
+        expanded += split[x] +  spacestr ;
 
-    return new Sk.builtin.str(self.v.replace(/\t/g, spacestr));
+    }
+    
+
+    return new Sk.builtin.str(expanded);
 });
 
 Sk.builtin.str.prototype["swapcase"] = new Sk.builtin.func(function (self) {
@@ -898,8 +920,32 @@ Sk.builtin.str.prototype["isupper"] = new Sk.builtin.func(function (self) {
 });
 
 Sk.builtin.str.prototype["istitle"] = new Sk.builtin.func(function (self) {
+    var input = self.v;
+    var cased = false;
+    var previous_is_cased = false;
+    var pos;
+    var ch;
     Sk.builtin.pyCheckArgs("istitle", arguments, 1, 1);
-    return Sk.builtin.bool(self.v.length && self.v === goog.string.toTitleCase(self.v));
+
+    for(pos = 0; pos<input.length; pos++){
+        ch = input.char(pos);
+        if(ch === ch.toUpperCase()){
+            if(previous_is_cased){
+                return Sk.builtin.bool(false);
+            }
+            cased = true;
+        }else if(ch === ch.toLowerCase()){
+            if(!previous_is_cased){
+                return Sk.builtin.bool(false);
+            }
+            cased = true;
+        } else {
+            previous_is_cased = false;
+        }
+    }
+
+
+    return Sk.builtin.bool(cased);
 });
 
 Sk.builtin.str.prototype.ob$type = Sk.builtin.type.makeIntoTypeObj("str", Sk.builtin.str);
--- a/src/str.js
+++ b/src/str.js
@@ -766,7 +766,7 @@ Sk.builtin.str.prototype["isspace"] = new Sk.builtin.func(function (self) {
     newstr = goog.string.normalizeWhitespace(self.v);
     for (i = 0; i < newstr.length; i++) {
         ch = newstr.charAt(i);
-        if (!goog.string.isSpace) {
+        if (!goog.string.isSpace(ch)) {
             return Sk.builtin.bool(false);
         }
     }
--- a/src/str.js
+++ b/src/str.js
@@ -702,18 +702,41 @@ Sk.builtin.str.prototype["replace"] = new Sk.builtin.func(function (self, oldS, 
 
 Sk.builtin.str.prototype["zfill"] = new Sk.builtin.func(function (self, len) {
     var newstr;
+    var width = len;
+    var data = self.v;
+    var buf;
+    var start;
+    var middle;
     Sk.builtin.pyCheckArgs("zfill", arguments, 2, 2);
     if (!Sk.builtin.checkInt(len)) {
         throw new Sk.builtin.TypeError("integer argument exepcted, got " + Sk.abstr.typeName(len));
     }
     len = Sk.builtin.asnum$(len);
     if (self.v.length >= len) {
-        return self;
-    } else {
-        newstr = Array.prototype.join.call({length: Math.floor(len - self.v.length) + 1}, "0");
-        return new Sk.builtin.str(newstr + self.v);
+        return new Sk.builtin.str(self.v);
+    } 
+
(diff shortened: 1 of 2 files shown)


src/str.js
```diff
@@ -716,6 +755,162 @@ Sk.builtin.str.prototype["isdigit"] = new Sk.builtin.func(function (self) {
...
 
+Sk.builtin.str.prototype["isspace"] = new Sk.builtin.func(function (self) {
+    var ch;
+    var i;
+    var newstr;
+    Sk.builtin.pyCheckArgs("isspace", arguments, 1, 1);
+    if (self.v.length === 0) {
+        return Sk.builtin.bool(false);
+    }
+    newstr = goog.string.normalizeWhitespace(self.v);
```
albertjan: why not just use: `self.v.replace(/\xa0|\s/g, ' ');` I don't think we use the closure lib anywhere else.

src/str.js
```diff
@@ -700,6 +700,45 @@ Sk.builtin.str.prototype["replace"] = new Sk.builtin.func(function (self, oldS,
...
+
+    for(var i=start; i<middle; i++){
+        buf[i] = "0";
+    }
+
+    for(var j=middle; j<width; j++){
+        buf[j]=data[start];
+        start = start+1;
+    }
+    // newstr = Array.prototype.join.call({length: Math.floor(len - self.v.length) + 1}, "0");
```
albertjan: remove?

src/str.js
```diff
@@ -716,6 +755,162 @@ Sk.builtin.str.prototype["isdigit"] = new Sk.builtin.func(function (self) {
...
+    var spacestr = "";
+    var spacerem;
+
+    Sk.builtin.pyCheckArgs("expandtabs", arguments, 1, 2);
+    if ((tabsize !== undefined) && !Sk.builtin.checkInt(tabsize)) {
+        throw new Sk.builtin.TypeError("integer argument exepcted, got " + Sk.abstr.typeName(tabsize));
+    }
+    if (tabsize === undefined) {
+        tabsize = 8;
+    }else{
```
albertjan: should be:
```js
}
else {
```

albertjan:
Very cool thanks! 

I added a few comments. there are some more coding style issues, that I think should have been caught with jshint. I'm going to check why that didn't happen :smile:

albertjan:
Allright I found it! Apparently the jshint guys don't want jshint to check indentation any more. See: https://github.com/jshint/jshint/issues/1677 . So I think we're going to have to use a separate js style checker.

src/str.js
```diff
@@ -716,6 +755,162 @@ Sk.builtin.str.prototype["isdigit"] = new Sk.builtin.func(function (self) {
...
+        }
+        strs_w.push(new Sk.builtin.str(slice));
+    }
+
+    return new Sk.builtin.list(strs_w);
+});
+
+Sk.builtin.str.prototype["title"] = new Sk.builtin.func(function (self) {
+    var input = self.v;
+    var buffer = new Array(input.length).join(" ").split("");
```
albertjan: What happens here?

src/str.js
```diff
@@ -716,6 +755,162 @@ Sk.builtin.str.prototype["isdigit"] = new Sk.builtin.func(function (self) {
...
+    var sol = 0;
+    var slice;
+    Sk.builtin.pyCheckArgs("splitlines", arguments, 1, 2);
+    if ((keepends !== undefined) && !Sk.builtin.checkBool(keepends)) {
+        throw new Sk.builtin.TypeError("boolean argument expected, got " + Sk.abstr.typeName(keepends));
+    }
+    if (keepends === undefined){
+        keepends = false;
+    } else {
+    keepends = keepends.v;
```
albertjan: indent

src/str.js
```diff
@@ -716,6 +755,162 @@ Sk.builtin.str.prototype["isdigit"] = new Sk.builtin.func(function (self) {
...
+Sk.builtin.str.prototype["swapcase"] = new Sk.builtin.func(function (self) {
+    var i;
+    var letters;
+    var newletters = "";
+    var ch;
+    Sk.builtin.pyCheckArgs("swapcase", arguments, 1, 1);
+    letters = self.v;
+
+    if (letters.length === 0){
+        return new Sk.builtin.str("");
```
albertjan: Why return a new string here and not just `self`

src/str.js
gerbal: Populating an empty array without using map and an anonymous function. It's also quite a bit faster with very large strings.

src/str.js
gerbal: fixed

src/str.js
gerbal: fixed throughout

src/str.js
gerbal: fixed througout

src/str.js
gerbal: removed.

src/str.js
gerbal: good suggestion, change in latest commit.

albertjan:
@bnmnetp lgtm! :shipit:

src/str.js
```diff
@@ -700,6 +701,46 @@ Sk.builtin.str.prototype["replace"] = new Sk.builtin.func(function (self, oldS,
     return new Sk.builtin.str(self.v.replace(patt, replacer));
 });
 
+Sk.builtin.str.prototype["zfill"] = new Sk.builtin.func(function (self, len) {
```
bzwheeler: I am a little curious about the complexity of the zfill implementation.  After reading over the python docs for this method I am wondering if there is a simpler/clearer solution?

```js
// figure out how many zeroes are needed to make the proper length
var zeroes = len - str.length;
// offset by 1 if there is a +/- at the beginning of the string
var offset = (str[0] === "+" || str[0] === "-") ? 1 : 0;
var pad = "";
// exit early if no zeroes are needed
if (zeroes <= 0) return str;
// create a string of as many zeroes as are needed
while(zeroes--) pad += "0";
// combine the string and the zeroes
return str.substr(0, offset) + pad + str.substr(offset);
```

src/str.js
```diff
@@ -716,6 +757,164 @@ Sk.builtin.str.prototype["isdigit"] = new Sk.builtin.func(function (self) {
     return Sk.builtin.bool(true);
 });
 
+Sk.builtin.str.prototype["isspace"] = new Sk.builtin.func(function (self) {
```
bzwheeler: Couldn't we replace this with a single regexp?

```js
Sk.builtin.pyCheckArgs("isspace", arguments, 1, 1);
return Sk.builtin.bool(/^\s+$/.test(self.v));
```

albertjan: :smile: yes

src/str.js
```diff
@@ -716,6 +757,164 @@ Sk.builtin.str.prototype["isdigit"] = new Sk.builtin.func(function (self) {
...
+        ch = newstr.charAt(i);
+        if (! goog.string.isSpace(ch)) {
+            return Sk.builtin.bool(false);
+        }
+    }
+    return Sk.builtin.bool(true);
+});
+
+
+Sk.builtin.str.prototype["expandtabs"] = new Sk.builtin.func(function (self, tabsize) {
```
bzwheeler: This also seems a bit complicated.  I'd suggest something simpler like:

```js
var spaces = (new Array(tabsize) + 1)).join(" ");
var expanded = self.v.replace(/([^\r\n\t]*)\t/g, function(a, b) {
  return b + spaces.slice(b.length % tabSize);
});
return Sk.builtin.str(expanded);
```

src/str.js
```diff
@@ -716,6 +757,164 @@ Sk.builtin.str.prototype["isdigit"] = new Sk.builtin.func(function (self) {
...
+            expanded += split[x] + spacestr;
+        }
+        else {
+            expanded += split[x];
+        }
+    }
+    return new Sk.builtin.str(expanded);
+});
+
+Sk.builtin.str.prototype["swapcase"] = new Sk.builtin.func(function (self) {
```
bzwheeler: again a regexp replace should suffice here:

```js
self.v.replace(/[a-z]/gi, function(c) {
  var lc = c.toLowerCase();
  return lc === c ? c.toUpperCase() : lc;
});
```

src/str.js
```diff
@@ -716,6 +757,164 @@ Sk.builtin.str.prototype["isdigit"] = new Sk.builtin.func(function (self) {
...
+        slice = data.slice(sol, eol);
+        if (! keepends) {
+            slice = slice.replace(/(\r|\n)/g, "");
+        }
+        strs_w.push(new Sk.builtin.str(slice));
+    }
+    return new Sk.builtin.list(strs_w);
+});
+
+Sk.builtin.str.prototype["title"] = new Sk.builtin.func(function (self) {
```
bzwheeler: Maybe I like regex replace too much but again I would use it here:

```js
self.v.replace(/[a-z][a-z]*/gi, function(str) {
  return str[0].toUpperCase() + str.substr(1).toLowerCase();
})
```

src/str.js
```diff
@@ -742,6 +941,35 @@ Sk.builtin.str.prototype["isupper"] = new Sk.builtin.func(function (self) {
     return Sk.builtin.bool(self.v.length && !/[a-z]/.test(self.v) && /[A-Z]/.test(self.v));
 });
 
+Sk.builtin.str.prototype["istitle"] = new Sk.builtin.func(function (self) {
```
bzwheeler: Might be simpler to just compare the incoming value against the result of calling ```title()``` on that value?

gerbal:
great suggestions @bzwheeler, I'll take a look at using regex where possible.

bzwheeler:
great, thanks @gerbal, I definitely can see where these implementations came from when I look at the pypy implementations but I think the language features of javascript warrant a different approach.  Thanks for contributing!

bnmnetp:
Hi guys,

Am I correct in my understanding that this is all working and could be accepted, except for some differences of opinion on how things are implemented?

Brad

albertjan:
@bnmnetp it works, some function could be made simpler with the regex suggestions done by @bzwheeler.

gerbal:
I did a little bit of testing and the Regex are probably faster too, but have some less predictable behavior I haven't had time to debug.

gerbal:
@bzwheeler I made a bunch of the changes you suggested. Regex is the way to go in almost every case.

bnmnetp:
This looks good to me.  I'm ready to accept it unlessI hear from someone else in the next day.

PR #324, opened 2015-02-04
Asynchronous (suspendable) loading of imported modules

meredydd:
This change allows imported modules to be loaded asynchronously.

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -304,6 +304,23 @@ Compiler.prototype._jump = function (block) {
     }
 };
 
+Compiler.prototype._checkSuspension = function(e) {
+    var retblk;
+    if (this.u.canSuspend) {
+
+        retblk = this.newBlock("function return or resume suspension");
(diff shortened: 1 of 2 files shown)


bnmnetp:
@bzwheeler have you checked this out with the Sk.externalLibraries functionality?  Any problems?

meredydd:
There is one issue I've discovered, and will be fixing this morning - turns
out the strategy of just attaching a traceback property to whatever gets
thrown through is not robust to things like, eg, throw(undefined);.  I'm
going to modify it to wrap everything thrown through it in an Sk.Exception
in order to attach a meaningful stack trace. Hope that's OK.

Meredydd

meredydd:
(sorry, I made a comment here that belonged on the wrong pull request. As far as I'm concerned, this one is ready to merge.)

bnmnetp:
Seems to work fine with Sk.externalLibraries for me.

PR #325, opened 2015-02-04
Provide stack traces for exceptions

meredydd:
Replace the global Sk.curr{LineNo,ColNo,Filename} with local vars in each stack frame, allowing us to generate proper stack traces for exceptions.

This required modifying one test, as it relied on exceptions reporting incorrect line numbers.

This PR merges nicely with PR #324, and they can be applied in either order.

### DIFF
--- a/HACKING.rst
+++ b/HACKING.rst
@@ -726,3 +726,13 @@ Build the distribution files for skulpt:
 
 * skulpt.min.js  -- This is a minified version of the core interpreter files.
 * skulpt-stdlib.js  -- This is an unminified version of library functions.  This file may contain javascript that implements a module, such as turtle or math, or it may contain pure python.
+
+
+
+
+Getting stack traces from an exception
+--------------------------------------
+
+Sk.builtin.Exception objects have a property called 'traceback'. This property contains an Array of objects with 'filename', 'lineno' and (optionally) 'colno' properties, each representing a stack frame. The array is ordered from innermost to outermost frame.
+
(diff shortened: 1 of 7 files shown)


waywaaard:
:+1:

bnmnetp:
Very cool.  It seems to work just fine for me.  the .toString() function on the error works as it always has.

Could you provide a little documentation on the error and the traceback information itself.  How would I go about giving the user a full traceback, and how could I access the local variables within a stack frame if I wanted to display them as well?

Brad

meredydd:
Yeah - I initially put it in the str(), but I decided to nod to the Python
convention that stringifying exceptions doesn't give you much, and you have
to inspect them properly to get a traceback.

Exceptions (indeed, any object that is thrown through Skulpt) receive a
"traceback" property, containing an Array of objects with the keys
"filename", "colno" and "lineno".

Where would you like this documentation in the source tree?

bnmnetp:
If you could just append that information to the end of the HACKING.rst document that would be great.  That is my repository for all kinds of stuff I want to document about skulpt.

meredydd:
Done :)

meredydd:
I've just pushed an update that changes how we deal with non-Skulpt exceptions (ie thrown objects that aren't instances of Sk.builtin.Exception). We now wrap them in an Sk.builtins.ExternalError, so that we can safely deliver weird things that are entirely legitimate to throw in Javascript (eg integers, or even 'undefined') into Python with tracebacks and other luxuries.

I've also added a commit to this PR to remove a few instances where the Skulpt runtime was throwing strings rather than Sk.builtins.Exceptions - I assume this was old code that predated the existence of errors.js.

meredydd:
Thanks :)

PR #330, opened 2015-02-15
adding str.format() support

### LINKED ISSUE #90: format is not implemented
"str {0}".format(something) fails to run properly.

### LINKED ISSUE #321: .format() is not working
.format is a standard function so I don't understand why it doesn't work: example:

print "The answer is {}".format(42)

does not work.

ERROR = AttributeError: 'str' object has no attribute 'format'

gerbal:
str.format is almost 100% supported. Fixes #90, fixes #321 

Some issues remain in repr support and I have no idea how to handle kwargs, any advice is welcome.

I'm sure I have overlooked some edge cases of various combinations of formatting specifications.

### DIFF
--- a/src/formatting.js
+++ b/src/formatting.js
@@ -62,6 +62,9 @@ Sk.builtin.str.prototype["format"] = new Sk.builtin.func(function (self) {
 			field_name = repsplit[0];
 	    	format_spec = repsplit[1];
 	    }
+	    else {
+	    	field_name = replacement_field
+	    }
 
 
 		parse_field_name(field_name);
--- a/src/formatting.js
+++ b/src/formatting.js
@@ -1,821 +1,908 @@
 Sk.builtin.str.prototype["format"] = new Sk.builtin.func(function (self) {
 
-	// implementation of pythons str.format() function
-	// step1: pull out {}'s
-	// step2: process {}'s
-	// step3: replace {}'s
+    // implementation of pythons str.format() function
+    // step1: pull out {}"s
+    // step2: process {}"s
+    // step3: replace {}"s
 
-	var replacement_field;
-	var input;
-	// var arguments;
-	var args;
-	var fields;
-	var field_dict = {};
+    var replacement_field;
+    var input;
+    // var arguments;
+    var args;
+    var fields;
+    var field_dict = {};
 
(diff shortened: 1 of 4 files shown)


bnmnetp:
Thanks for the PR. 

If you haven't looked at it already, the HACKING.rst file has some information on doing kwargs.

Brad

bnmnetp:
One problem is that **kwargs does not work properly in Skulpt.  This is a separate issue that should not keep this PR from getting accepted.  

```python
coord = {'latitude': '37.24N', 'longitude': '-115.81W'}

def x(latitude=1, longitude=1):
   print latitude, longitude

x()
x(**coord)
```

fails on line 7.

gerbal:
Good to know about **kwargs. I disabled the test that cause Travis failures on **kwargs.

I'm still not sure why sk.builtin.repr() is failing in the remaining unittest. Any insight?

src/formatting.js
```diff
@@ -0,0 +1,363 @@
...
+            precision = parseInt(precision, 10);
+        }
+
+        formatFormat = function(value){
+            var r;
+            if(conversion === undefined || conversion === ""){
+                return value;
+            }
+            else if( conversion == "r"){
+                r = Sk.builtin.repr(value);
```
bnmnetp: The problem may be that by the time you are here, value is a raw javascript string.  It is no longer a python object.  ``repr`` does not deal with javascript objects, hence it returns unknown.

gerbal: Thanks! With that cleared up, all test are passing.

bnmnetp:
I added the following test to test_strformat.py, however it failed.

```
    def test_width(self):
        self.assertEqual('         2,2',"{0:10},{0}".format(2))
        self.assertEqual('foo bar baz ',"{0:4}{1:4}{2:4}".format("foo","bar","baz"))        
```

Its odd how real python adds space in the right for strings, but the left for numbers.

gerbal:
That's an odd behavior. and not one specified in PEP 3101. I added the spacing tests. Now, I have more more work to do.

bnmnetp:
I commented out the tests not covered by PEP 3101, and accepted this PR.

I would rather have a mostly working version of format for us to use and improve upon than to have this languish as an unaccepted PR.

Brad

PR #340, opened 2015-02-20
Update to allow a pixel to store the x and y location

ericsonga:
I had to fix the code to match Brad's changes.


bnmnetp:
That all looks like it works.

It would be cool to now integrate the new Suspensions feature with the image module so you could show the original image and then the student could see pixel by pixel updates as the image is transformed.

PR #342, opened 2015-02-22
Wrapped turtle module in a closure tied to DOM

### LINKED ISSUE #341: Cannot have more than one turtle app on a page
Ben,

I ran across one more regression with the new turtle.  I have always been able to keep the results of multiple turtle programs visible on the page, so I could go look at the output from one or more programs just by scrolling around.  However in the new world, when I run a second turtle program on the same page the previous turtle output disappears.   

Is there a way to prevent the previous canvases from getting destroyed? 

Thanks,

Brad

bzwheeler:
Updated turtle graphics module to be encapsulated in a closure that is tied
to the currently configured dom element set by Sk.TurtleGraphics.target.

This means that multiple turtle apps can be running simultaneously.

fixes #341

### DIFF
--- a/src/lib/turtle/__init__.js
+++ b/src/lib/turtle/__init__.js
@@ -2,12 +2,41 @@ var $builtinmodule = function (name) {
 "use strict";
 
 // See if the TurtleGraphics module has already been loaded
-if (Sk.TurtleGraphics && Sk.TurtleGraphics.module) {
-    Sk.TurtleGraphics.reset();
-    return Sk.TurtleGraphics.module;
+// for the currently configured DOM target element.
+var currentTarget = getConfiguredTarget();
+
+if (!currentTarget.turtleInstance) {
+    currentTarget.turtleInstance = generateTurtleModule(currentTarget);
+}
+else {
+    currentTarget.turtleInstance.reset();
 }
 
-return (function() {
+Sk.TurtleGraphics.module = currentTarget.turtleInstance.skModule;
+Sk.TurtleGraphics.reset  = currentTarget.turtleInstance.reset;
+Sk.TurtleGraphics.focus  = currentTarget.turtleInstance.focus;
+Sk.TurtleGraphics.raw = {
+    Turtle : currentTarget.turtleInstance.Turtle,
+    Screen : currentTarget.turtleInstance.Screen
+};
+
+return currentTarget.turtleInstance.skModule;
+
+function getConfiguredTarget() {
+    var selector, target;
+
+    selector = (Sk.TurtleGraphics && Sk.TurtleGraphics.target) || "turtle",
+    target   = typeof selector === "string" ?
+        document.getElementById(selector) :
+        selector;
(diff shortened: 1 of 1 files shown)


bnmnetp:
Wow, Thanks Ben!

In the meantime, I have one other minor thing for you to look at.  Try this in a trinket:

```
import turtle

wn = turtle.Screen()
t = turtle.Turtle()

#wn.setworldcoordinates(-10,-10, 10, 10)
t.goto(-wn.window_width()/2, -wn.window_height()/2)
```
Note the turtle position when you run this.  Now, if you uncomment the call to setworldcoordinates you will see that the turtle goes all the way to the corner as you would expect.

Brad

bzwheeler:
So oddly enough this is actually the "correct" behavior, despite it not being terribly intuitive.  The methods `window_width()`, `window_height()` and `screensize()` don't really work in the way one might expect.  The previously mentioned methods actually just dictate how large, in pixels, the canvas should be and don't reflect the bounding box coordinates of the turtle world.  There doesn't seem to be a "getter" for the actual world coordinates as set by `setworldcoordinates`.

Check out [this trinket](https://trinket.io/python/6c13dce9b9), it might clarify things.

bnmnetp:
That is odd.  I've always assumed that the initial world coordinates were essentially in pixels.

bzwheeler:
So with nothing overridden, you are correct.  The default size is 400x400 with the world set to -200,200,-200,200.  However, with trinkets in particular, the window world coordinates stay -200 to 200 but the screen size is dependent on the size of the embedded trinket.

bnmnetp:
Ah,

So, if Sk.TurtleGraphics.{width,height} are set would it make sense to use those values to calculate the parameters to the setupWorld function in the Screen function?  

I’m trying to find a way to make this behave as close to real python as we can.

Brad


Brad Miller
Associate Professor, Computer Science
Luther College
Founder, Runestone Interactive
Blog: http://reputablejournal.com <http://reputablejournal.com/>

bnmnetp:
@bzwheeler, I made this simple change to the Screen function, and now I get the behavior I expect with respect to using `window_width` without calling `setworldcoordinates` or `setup`

Does the change make sense to you?

```
    function Screen() {
        var w,h;
        this._frames    = 1;
        this._delay     = undefined;
        this._bgcolor   = "none";
        this._mode      = "standard";
        this._managers  = {};
        this._keyLogger = {};
        if (Sk.TurtleGraphics.height && Sk.TurtleGraphics.width) {
            w = Sk.TurtleGraphics.width/2;
            h = Sk.TurtleGraphics.height/2;
        } else {
            w = 200;
            h = 200;
        }
        this.setUpWorld(-w,-h,w,h);
    }
```

bzwheeler:
@bnmnetp Yeah, that makes sense.  It would be best to change the `Sk.TurtleGraphics.width/height` references to `_config.width/height` so that each Turtle App can be configured independently.  Looks great other than that!

PR #344, opened 2015-02-24
Fix __getattr__ and __setattr__

meredydd:
```__getattr__``` and ```__setattr__``` support was broken (wrong lookup order), and pretty difficult to use because ```object.__setattr__``` was not available, so if you implemented a ```__setattr__``` there was no way to set *any* attributes in your object dictionary!

I've fixed this, and added a test. ~~In addition, these calls are now suspendable, so they can block.~~


meredydd:
I've withdrawn the suspendability commits from this PR, due to a bug that wasn't caught by the tests, but the rest should still be good to go.

(It turns that the way we fall back from something that might have been a numeric bin-op but isn't (eg foo.bar += "hi") is to re-walk the attribute/item chains again(!) inside an if(){} statement. This isn't really suspension-compatible. Expect a separate PR bringing some sanity to that mechanism and then going for suspendability again.)

bnmnetp:
@bzwheeler, @albertjan Could one or both of you take a quick look at these?  I would feel a lot more comfortable about accepting them if we could get at least 3 of us (including @meredydd ) to agree that this and #343 is ready to accept.

Thanks,

Brad

albertjan:
@bnmnetp this one looks good to me! 

@meredydd I can't believe no one ever ran into this before. :smile: 

I don't see why the `chain` function in misceval is in this PR. I understand it's a useful function, but it has very little to do with `__getattr__` and `__setattr__`. It was probably meant as a part of the suspend-ability of these function.

meredydd:
It was indeed included for the sake of the suspendability stuff (also the base for some other work that I hope to PR in the next couple of weeks).

Working with suspensions in JS without something like this is like pulling teeth, so if it doesn't offend your sensibilities I'd prefer to leave it in, but feel free to rebase it out (or ask me to, although there will be a one-day lag on that as I'm away from my desk today).

albertjan:
I'm not at all against leaving it in. We should really make up our minds on javascript testing. So we can test these kinds of functions.

bzwheeler:
@bnmnetp Looks good to me!

Issue #219, opened 2014-05-10
raw_input() does not work if the parameter is omitted


csev:
Hi all, This should be an easy one-liner and I would even fix it if I had a skulpt dev environment on my laptop - but while 

x = raw_input('prompt') works

x = raw_input() fails with 

"TypeError: Cannot read property 'v' of undefined"

![TypeError](https://cloud.githubusercontent.com/assets/1197222/2937385/bf89023a-d89e-11e3-8d07-459e20aa1afd.png)

PR #345, opened 2015-02-25
first try at adding suspensions to updatePixel

bnmnetp:
In an attempt to really get my head around suspensions, I thought I would try to add suspensions to the image module.  I have a couple of goals in mind:

1.  I would like to be able to use updatePixel in a way that when a single pixel is changed, that change would be reflected in the image in the browser window, this would give sutdents a much better visual understanding of nested iteration patterns.

2.  I would like to be able to grab an image from an arbitrary URL and bring it in for working on.  Right now the module is limited to working with images that are already working on the page.  I figure once I have this fully understood and working this knowledge can be used to overhaul our file handling in the main body of skulpt.

My first naive attempt to add suspensions to the updatePixel function looks like the following.

```javascript
	    $loc.updatePixel = new Sk.builtin.func(function (self, pixel){
            var susp = new Sk.misceval.Suspension();
            susp.resume = function() { return Sk.builtin.none.none$; }
            susp.data = {
                type: "Sk.promise",
                promise: new Promise(function(resolve) {
                    var x = Sk.builtin.asnum$(Sk.misceval.callsim(pixel.getX, pixel));
                    var y = Sk.builtin.asnum$(Sk.misceval.callsim(pixel.getY, pixel));
                    var index = (y * 4) * self.width + (x * 4);
                    self.imagedata.data[index] = Sk.builtin.asnum$(Sk.misceval.callsim(pixel.getRed, pixel));
                    self.imagedata.data[index + 1] = Sk.builtin.asnum$(Sk.misceval.callsim(pixel.getGreen, pixel));
                    self.imagedata.data[index + 2] = Sk.builtin.asnum$(Sk.misceval.callsim(pixel.getBlue, pixel));
                    self.imagedata.data[index + 3] = 255;
                    resolve();
                })};
            return susp;
	    });
```
I took this basically by following the example in time.  But turtle and others do things much differently.  In particular the chaining that @meredydd just added seems like it might be relevant here.  So, I'm looking for comments and discussion to make sure I get this right, and to make sure I understand what I'm doing so I can better follow other Suspension related PR's.

Brad


meredydd:
Does it work? Looks sane to me.

This is an interesting application, as it's not so much a blocking operation as "just let the browser event loop run for a moment so the user can see the intermediate state on the screen" - correct? (Although if you're relying on that for your user experience, people with faster/slower browsers might not get the full benefit...)

The chain() function in my PR is designed for handling things that have to happen *after* any suspension resumes. (For example, if we're calling back into user code, then we have to do something with its return value in JS before we return our final result. ```__getattr__``` and ```import``` are obvious examples.) I don't think it applies here, though.

LeszekSwirski:
I think a good example of `chain` would be if the pixel getters (`getX` etc) were potentially suspendable -- it looks like this is never the case here, but one could technically imagine allowing the user to pass in a user-defined `Pixel` subclass with complex getters.

bnmnetp:
Yeah, the Suspensions make a lot more sense for the Image constructor where I would like to be able to provide a map, or allow the user to put in a URL to the image.  I implemented this today, and it is working in prototype form.  Minus the fact that I need to get around the `Unable to get image data from canvas because the canvas has been tainted by cross-origin data.` error.

My first thought was that the stuff that should come after the image was successfully loaded should be placed in the  `susp.resume` function.  but that does not appear to ever get executed.  What is the purpose of that function?  When should it be used?

```javascript
   var image = function ($gbl, $loc) {
        $loc.__init__ = new Sk.builtin.func(function (self, imageId) {
            //self.image = document.getElementById(imageId.v);
            self.image = null;
            if (self.image == null) {
                var susp = new Sk.misceval.Suspension();
                susp.resume = function() {
                    // Should the post image get stuff go here??
                    console.log("resuming");
                };
                susp.data = {
                    type    : 'Sk.promise',
                    promise : new Promise(function(resolve) {
                            var newImg = new Image();
                            newImg.onload = function() {
                                self.image = newImg;
                                console.log('got goldy')
                                self.width = self.image.width;
                                self.height = self.image.height;
                                self.canvas = document.createElement("canvas");
                                self.canvas.height = self.height;
                                self.canvas.width = self.width;
                                self.ctx = self.canvas.getContext("2d");
                                self.ctx.drawImage(self.image, 0, 0);
                                self.imagedata = self.ctx.getImageData(0, 0, self.width, self.height);
                                resolve();
                            };
                            newImg.src = "http://interactivepython.org/runestone/static/thinkcspy/_static/goldygopher.png";
                        }
                    )
                };
            }

        });
```

Thanks, for the comments and advice.

Brad

Brad

bnmnetp:
@meredydd or @LeszekSwirski  one more question on this.  How can I report errors back out the Promise in the example above.  For example lets say the image does not load.  So I could have an img.onerror callback, but how would I get that out of the Promise/Suspension??  It seems like having the function have a reject parameter might be the right way but that doesn't seem to do anything.

Also, If I have a mistake in my my own javascript code, and it throws an exception, the exception just gets swallowed up somewhere and never gets reported to the console, much less back to the main asyncToPromise call that calls my importMainWithBody.

Thanks,

Brad

LeszekSwirski:
Here we have an interesting lesson on the interaction between JS promise, suspensions, and Python. A Promise object is passed two callbacks, `resolve` on success and (optionally) `reject` on error. If you look at how Promise suspensions are handled in [`asyncToPromise`](https://github.com/skulpt/skulpt/blob/4b063e7bb9b109f2b7cb5858323d76c611295d0a/src/misceval.js#L842), the `promise` data field of the Suspension, which is your Promise, is passed two handlers: [`resumeWithData`](https://github.com/skulpt/skulpt/blob/4b063e7bb9b109f2b7cb5858323d76c611295d0a/src/misceval.js#L811) for success and [`resumeWithError`](https://github.com/skulpt/skulpt/blob/4b063e7bb9b109f2b7cb5858323d76c611295d0a/src/misceval.js#L819) for failure. Looking at the definition of these, they fill in the Suspension's `result` or `error` data fields, respectively, and then call the Suspension's `resume` method, passing it back into `handleResult`. Note that the Suspension's `resume` method is distinct from a Promise `resolve` callback.

At this point, you are in the Suspension's `resume` method, whose default behaviour is [to call `resume` on the Suspension's child](https://github.com/skulpt/skulpt/blob/4b063e7bb9b109f2b7cb5858323d76c611295d0a/src/misceval.js#L21). This is what allows suspensions to bubble up the call stack, the default behaviour when a generated function receives a Suspension instead of a value is to [create its own Suspension with the former Suspension as the child](https://github.com/skulpt/skulpt/blob/4b063e7bb9b109f2b7cb5858323d76c611295d0a/src/compile.js#L905). In this case, you don't want this bubbling behaviour, so you have a custom `resume` method, which has access to the `result` and `error` data fields of the Suspension. It is now up to you how to convert these into Python objects, by returning a value or throwing an exception. For example, `time.sleep` always returns `None`, and assumes there can never be an error.

Hope that helps.

bnmnetp:
Trying to follow along here:

Did you mean the suspensions `data` and `error` fields?  I've not seen any example or documentation about the `error` field or the `result` field.  If it is `data` then I think I am beginning to make sense of whats going on.

bnmnetp:
Deleted my previous response.

So, I am making progress.   I think the answer to my question is that in my  "Promise function" If there is some kind of error in loading the image I should set `susp.error` to some value to indicate that the image was not loaded correctly.  THEN in my `susp.resume` function I should check that value and throw an exception.  Since this is part of a Python init method, I don't need to return any value for the result.

HOWEVER, if this were something like file.open then I would want to return a file object from `susp.resume`

meredydd:
Yes, I think that's right.

LeszekSwirski:
Not entirely right -- you should add a `reject` callback argument to your Promise object along with your `resolve` callback, and on error, pass the error into `reject`. Then, read off `this.data["error"]` (equivalent to `susp.data["error"]`) in `susp.resume`. Your suggested method would work too btw, @meredydd isn't wrong, but I believe that this would be the idiomatic way.

Come to think of it, there should probably be a helper method for creating these promise suspensions out of a promise object that includes a good default `resume` implementation.

bnmnetp:
Thanks guys,  I'm feeling very good about my latest commit.  Using reject properly I believe.

It does feel a bit awkward to check this.data["error"] in the resume function.  For some reason in my mind it feels like there should be a susp.error or susp.reject function that gets called when reject is called rather than resume.  But I also assume you had a reason for doing it the way you did.

This exercises has helped a lot, and has really filled in the gaps that are not covered in the time example.

Thanks.

LeszekSwirski:
As I recall, there's some complication with Suspensions inheriting their children's data, and this the reason for the awkward data check. I don't entirely remember though, I'd have to refresh my memory to explain this in more detail.

LeszekSwirski:
Having said that, I'm sure there's a way of wrapping it to be less awkward.

bnmnetp:
Or maybe as you said, a good default implementation where resolve(my return value) would return the given Python value, and  reject(Error(my error message)) would throw a Python exception.  That might cover a big fraction of the use cases, and eliminate the need for someone like me to even worry about writing the resume function.

bnmnetp:
I am very happy with how this little experiment has turned out.  Suspensions are awesome, and I see how I could very easily do something now to improve the way we imitate opening/writing files.

This is also making me think about the execution limits.  I do think we need to keep them, especially for programs that do not use suspensions.  But for animated programs like image processing, turtle programs, and other longer running algorithms, I'm wondering if you either of you could see a way to inject some kind of interrupt.  I'm thinking it would be nice to have a button that when clicked would set a flag that could be checked whenever a suspension resolves in order to stop the entire program.  A virtual Ctrl-C.

LeszekSwirski:
> I'm thinking it would be nice to have a button that when clicked would set a flag that could be checked whenever a suspension resolves in order to stop the entire program.

That is a exactly what a couple of my PRs (#219, #292, #305) were for.

Consider a Suspension handler -- it receives a Suspension, and returns a Promise which resumes the execution of the program. This Promise, however, doesn't have to ever resolve, which would effectively end the execution of the program (or rather, indefinitely suspend it, which is equivalent once the Suspension is garbage collected). Or, it could always reject, passing an `InterruptedException` to the reject handler. Either way, you could write a suspension handler which conditionally on some flag returns this Promise rather than one which resumes the program.

Thus, the difficulties are: handling all Suspensions, resuming the program correctly, and ensuring that Suspensions happen often enough for this to be meaningful. These are solved by the above PRs:

  1. Handling all Suspensions solved by #292, where you can define a handler for `"*"` rather than individually for `"Sk.promise"`,`"Sk.debug"` etc., and it gets executed for every Suspension.
  2. Resuming correctly is solved by #291, where returning `false` from a handler executes the default resuming behaviour -- this means that you can conditionally either return your "always reject" Promise if your interrupt flag is true, or return `false` to pass through into default resuming behaviour.
  3. Ensuring that Suspensions happen is solved by #305, which introduces a "soft timeout" suspension. This is similar to the execution limits, but instead of throwing an exception, it forces the program to return a `"Sk.yield"`

PR #347, opened 2015-02-26
Suspendable __(get|set)(attr|item)__

meredydd:
So, here's an updated version of the work I omitted from my previous PR, which makes ```__getattr__```, ```__setattr__```, ```__getitem__``` and ```__setitem__``` suspendable.

I've significantly modified how we handle augmented assignments (```+=``` and friends), so that we only walk attributes or items once. (That is, when you do ```x.y.z += 4``` or ```x.y["z"] *= 7```, ```x``` and ```x.y``` are only looked up once.) Without the recursive ```vexpr()``` calls in the ```AugStore``` handling (also in ```vexpr()```), suspending becomes straightforward.

Aside: In preparing this PR, I've become aware of a missing/forgotten/bit-rotted feature: There is a provision, in the way that AugStores are implemented, for ```Sk.abstr.numberInplaceBinOp()``` (itself a bit of a misnomer - it handles all binops) to return ```undefined```, signalling that the binop was done in-place, so there was no need to perform any more assignments. (This would mean, for example, that when executing ```x.y += 1```, ```Sk.abstr.sattr()``` would not need to be executed if ```y``` was a number that could therefore be updated in-place. Probably a worthwhile speed boost.)

However, ```Sk.abstr.numberInplaceBinOp()``` has no code path that can return ```undefined```, so this short-cut is never taken (I double-checked by running all our tests with an assertion). I have, however, left the short-cut in place in the ```AugStore``` handling, along with an explanatory comment. This is because (a) this would be a useful feature and I want to leave the door open to it, and (b) leaving the short-cut in there makes it more obvious why I've implemented AugStores the way I have, and prevents someone from accidentally making it harder to use that short-cut later.


meredydd:
(I just tweaked this to make augmented assignments suspendable as well, and add them to the tests. I promise I'm done now :) )

PR #348, opened 2015-03-02
non-read opens proof of concept

brianpmarks:
@bzwheeler Proof of concept to open up a discussion about the best way to incorporate this into skulpt. I imagine this implementation isn't merge-able upstream and likely needs to be a new configurable option.

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -610,9 +610,9 @@ Sk.builtin.open = function open (filename, mode, bufsize) {
     if (mode === undefined) {
         mode = new Sk.builtin.str("r");
     }
-    if (mode.v !== "r" && mode.v !== "rb") {
-        throw "todo; haven't implemented non-read opens";
-    }
+    //if (mode.v !== "r" && mode.v !== "rb") {
+    //    throw "todo: haven't implemented non-read opens";
+    //}
     return new Sk.builtin.file(filename, mode, bufsize);
 };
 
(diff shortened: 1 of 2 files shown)


brianpmarks:
I opened this on the wrong repo. Sorry about that.

bnmnetp:
@brianpmarks, @eah13,
I’ve just finished coding an upgrade to the image module that allows me to load images by URL (using Suspensions)  rather than having them embedded in the page.  This is something I’ve wanted to do with files as well, and I think my solution for opening an image file is a good step.   I will probably have to run an image proxy to get around Same Origin restrictions in the browser.   However, I’ve been thinking that for intro students it would be nicer for the instructor to tell the students to use image X and we could provide a map from X to the URL for X behind the scenes.    If you guys are going to be overhauling file handling, lets talk and collaborate on it right away.

brianpmarks:
@bnmnetp Our current plan isn't quite as ambitious. We're essentially looking at firing an event any time a file is opened in write or append mode and then fire another event when, for example, `file.write` is called. We're thinking this would be an option passed in to configure and then left to the caller to listen for those events. Since trinkets have file tabs, we'll create a new tab if it doesn't exist and write to it if it does.

That being said, we actually run a simple proxy server for handling `urllib.request` requests. This is for getting around Same Origin restrictions as well as handling non-https URLs since trinket is over https.

Would be happy to talk over either of these.

PR #355, opened 2015-03-09
Adds Screen method aliases to turtle module

### LINKED ISSUE #352: extra turtle appears after hideturtle call when using events
@bzwheeler 

This little program illustrates the problem:

```
import turtle

# Create our turtle and screen objects
t = turtle.Turtle()
wn = turtle.Screen()
wn.setup(300,300)
t.hideturtle()

# define the callback
def gotopoint(x, y):
    t.goto(x,y)

# connect the callback with the click event and wait
wn.onclick(gotopoint)
turtle.mainloop()
```

The turtle is hidden initially but then reappears but stays put after the first click.  I'm guess it has something to do with the multi-canvas approach for animating the turtle separately from the picture, but just guessing.

bzwheeler:
Explicitly adds screen method aliases without requiring creation of an anonymous turtle

fixes #352

### DIFF
--- a/src/lib/turtle/__init__.js
+++ b/src/lib/turtle/__init__.js
@@ -1017,28 +1017,6 @@ function generateTurtleModule(_target) {
         };
         proto.$shape.minArgs = 0;
 
-        proto.$window_width = function() {
-            return this._screen.$window_width();
-        };
-
-        proto.$window_height = function() {
-            return this._screen.$window_height();
-        };
-
-        proto.$tracer = function(n, delay) {
-            return this._screen.$tracer(n, delay);
-        };
-        proto.$tracer.minArgs = 0;
-
-        proto.$update = function() {
-            return this._screen.$update();
-        };
-
-        proto.$delay = function(delay) {
-            return this._screen.$delay(delay);
-        };
-        proto.$delay.minArgs = 0;
-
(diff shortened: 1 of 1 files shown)


src/lib/turtle/__init__.js
```diff
@@ -1017,28 +1017,6 @@ function generateTurtleModule(_target) {
...
 
-        proto.$window_width = function() {
-            return this._screen.$window_width();
-        };
-
-        proto.$window_height = function() {
-            return this._screen.$window_height();
-        };
-
-        proto.$tracer = function(n, delay) {
```
bnmnetp: @bzwheeler, This appears to have the unintended consequence of removing the tracer method from turtle.  Even though I've been running this locally with the latest build I didn't notice this until I moved this latest set of changes to the live site!  Ugh!   t.tracer(10,30)  Now throws an error that tracer is not a method of turtle.

bnmnetp: As an experiement, I just added all of these back into the source at the spot where they were removed.  All appears well.  Although there may be a different problem that I overlooked this is certainly an easy solution.

Brad

bzwheeler: Hmm, yeah, that should work, was indeed an unintended consequence.

PR #359, opened 2015-03-19
Skulpt processing: added bezier

dsblank:
Here is my first contribution to Skulpt. Just making sure that this is an appropriate style, and that I took care of all of the necessary details. Let me know. Then I'll go through the rest of the processing lib and see what else is missing.

### DIFF
--- a/src/lib/processing/__init__.js
+++ b/src/lib/processing/__init__.js
@@ -79,6 +79,15 @@ var $builtinmodule = function (name) {
         mod.processing.triangle(x1.v, y1.v, x2.v, y2.v, x3.v, y3.v);
     });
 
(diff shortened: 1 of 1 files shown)


bnmnetp:
I would be happy for you to rearrange the code.  My work was more along the lines of proof of concept, so if there are better ways to do things then go ahead.

I am spending the spring semester in Malta as the director of Luther's Malta and the Mediterranean study away program.  Malta is GMT+1, so its late morning here.

Brad

bnmnetp:
You are an early riser if you are in the US...

dsblank:
Thanks! We'll at least document what is missing, and hopefully fill in the rest.

Beautiful area of the world!

Yes, early rise today... couldn't sleep.

Issue #361, opened 2015-03-24
Processing mouse x/y


eah13:
Looks like the mouse x/y implementation is returning a non-int, preventing it from working as designed?  

https://trinket.io/library/trinkets/a10acb232a

bnmnetp:
Strange, in either case I get:  ``AttributeError: 'Mouse' object has no attribute 'x' `` when I try to run it in my test page.   I'm thinking that @bzwheeler  has made some changes that haven't been pushed back, since I also remembered/verified that the implementation in the repo expects a ``canvas`` and not a ``div`` as its output element.

bnmnetp:
Very interesting.  Many of my old examples are broken because in each case I get a problem with an attribute such as mouse.x  or Environment.frameRate not found.  In every case I had to update the ``__getattr__`` method to remap the key parameter back to a Javascript string.  I know this worked fine before, but I have no idea when it stopped.  Here is what I did. So, I have no idea how either version is working for you @eah13.

```javascript
        $loc.__getattr__ = new Sk.builtin.func(function (self, key) {
            key = Sk.ffi.remapToJs(key);
            if (key === "x") {
                return mod.processing.mouseX;
            }
            else if (key === "y") {
```

Tagging @dsblank for this discussion because I know he is working on adding giant chunks of functionality into the module.

Doug, I'm pretty sure I can push this bug fix without causing merge conflicts for you...

bnmnetp:
OK, now that I've cleaned up the problems with getattr I see what the problem is in your example

```python
from processing import *

def draw():
    ellipse( mouse.x, mouse.y, 50, 50)  # doesn't work on Chrome :(
    #ellipse(int(mouse.x), int(mouse.y), 50, 50)  # this works - where's the rub? 
    
    print (mouse.x, mouse.y)

run()
```

the problem is that ``__getattr__`` is being bad and returning a "javascript int" NOT a "python int".  Whereas elllipse expects to get a "python int".

So, the processing module was written in a time when we were much less thoughtful about always returning proper python data types.  @dsblank as you wrap things this is important for you as well.

I'll fix up the return types on my branch so you can see what I'm referring to.

Brad

dsblank:
I think this is fixed now too. Can be closed.

Francyfx:
why mouseX and mouseY couldn`t useful

PR #366, opened 2015-03-26
Add most of missing ProcessingJS bridge

dsblank:
Needs to have some testing done, and needs some missing constants to 100%.


bnmnetp:
Great progress Doug.

Going back to the discussion around the gist and the turtle from a day or two ago, I think we need to make turtle, image, and canvas all consistent about what they expect in terms of a div versus a canvas. Turtle and Image now both construct a canvas as a child of the provided div element.  Can you make that change as a part of this PR?

Just for quick testing purposes, could you also check that the proctest.html file in the doc/static directory is also up to date and working with your changes?

bnmnetp:
So I spent a few minutes this morning trying to track down why I couldn't even get the simple example in #361 to work.  I have a fix for that.   so now all of the programs in ``proctest.html`` work.

I also mentioned the div versus canvas issue.  I know what do do there, and I could also go ahead and fix that up this afternoon.

Doug are you comfortable with rebasing your own stuff If I push up those changes?  Otherwise I could hold off.

The challenge is that our Easter break is starting tomorrow and I head out to Morocco with the students for a week, and then another week of R&R in Spain (no students).  I'll have internet access and an iPad but no macbook for hacking.

bnmnetp:
OK, I've pushed up a branch that contains my changes that way you can grab the branch and see what I've done and merge as you see fit.

dsblank:
Thanks @bnmnetp ! I was able to incorporate your changes into this PR. I was wondering why some of the lookups on the classes weren't working. We might later be able to remove mouseX, mouseY, pmouseX, and pmouseY in favor of mouse.x, mouse.y, mouse.px, and mouse.py. But we can discuss when you get back.

Love the change about making the canvas if one is needed! Great!

bnmnetp:
Just pushed my changes so that getattr returns proper python ints or bools in some cases for things like mouse.x and mouse.y

You are right.  This needs a bunch of testing now.

dsblank:
How to proceed?

1. I can push to your proctest branch
2. I can incorporate your last changes into my PR (I have everything but the return values in the key lookups)

bnmnetp:
I'm fine with either strategy.  If @eah13 and/or @bzwheeler  and others are going to do some testing and possibly contribute some fixes then the branch would make sense as a common development area.  

However, I suspect that you may be the only one contributing code over the next few days and so keeping this PR in place would be easy and preserves the conversation.

dsblank:
Ok, I'll bring in the changes regarding proper Python types (I had noted that in my PR too) into this PR. And I think I see a few other fixes I can make over the next couple of weeks. I'll put those in the PR.

Thanks!

bnmnetp:
Sounds good.   Feel free to add to proctest.html if it makes sense to have a few more strenuous examples/tests on that page.

I got an email from someone yesterday, I think, about switching to p5.  The person said that processing.js was at a dead end as far as development?   I haven't really kept up, and I'm assuming p5 probably has a different API than processing.  Do you know about p5?  Any opinions on whether that should be a long term goal?

dsblank:
Regarding testing, I'm planning to make a series of Jupyter notebooks that can be used both as documentation and testing. I don't know if they could be worked into Travis, but they will be available.

BTW, ProcessingJS + Java->JS converter is not as useful as Skulpt + ProcessingJS because Skulpt has good error messages. ProcessingJS + Java converter is nearly useless if there is a bug. So, not good in the intro classroom.

After doing this transformation, I feel like a lot could be done in automatic Javascript bridging. I know about p5 (and many other great Javascript libraries) that could be used in this same manner as Processing. 

I suggest a project that looks at either the automatic bridging, or perhaps a human-guided semi-automatic bridging. IronPython has a similar technique for being able to import DLLs on the fly and make them look like Python libraries.

bnmnetp:
An automated bridging project would be awesome.  I gave it some thought when I created the document module to bridge to some DOM functionality, but ultimately gave up and just did the work of bridging common DOM functions.   In retrospect that was probably a horrible place to start.  Better to look at some examples that are not quite so tangled.

The notebooks are under your Calysto project?

I'm a big fan of iPython notebooks.  I use them for making class notes and 'handouts' in most of my upper level courses.  I need to take a day or two and dig in to this Jupyter/iPython split.

I don't know if you are aware, but you may also be interested to know that @waywaaard has done some great work in partially recreating matplotlib and numpy.

dsblank:
An automated ability to import Javascript might be very hard when going from no types to a typed language. But maybe Skulpt could be adapted to better deal with native Javascript values? That sounds like part of a bigger project... maybe a grant proposal...

Yes, we'll put the notebooks under Calysto. 

I wondered what the relationship with @waywaaard 's repo and Skulpt. I will probably include modules like numpy and matplotlib. My plan is to use Skulpt as the visualization and widget components of notebooks: students will do their hard-core CPython-based number crunching, but still have interactive, running code in static notebooks via Skulpt. IPython has new "widgets" but they are useless in static (rendered) notebooks. Skulpt could fill that niche.

bnmnetp:
Have you seen the section on skulpt.org about third party modules?   The crew at trinket.io have done a great job of wrapping pygal for charts too.

tildebyte:
Some comments:

1. I've committed a few of the "official" Processing examples ported to Processing.py, so I feel safe in saying I'm pretty familiar with it ;)  One of the things I love about P5.py is being able to drop in most any Java library with ease, so the idea of automagically wrapping JS libs for use with Skulpt is awesome.
1. Given the above, most of my testing will be with existing Python examples from the P5.py repo.
1. IMHO, Processing.js is waning (although in fairness, the first commits in something like a year have gone in over the last week), and p5.js is the future. p5.js is on the Processing home page, along with Processing.py, and both are directly supported by the Processing Foundation. P5JS never had that honor.

From the p5.js [FAQ](https://github.com/processing/p5.js/wiki/Frequently-Asked-Questions):

> The main goal of Processing.js is to execute Processing files in HTML5, but not necessarily to write native HTML5. It supports a mixed syntax of Processing and JavaScript, where the JavaScript is not really meant to be consumed by the end-user. 

and

> with p5 we are reimagining Processing's original goals in native JavaScript, in a way that is intended to be transparent and intuitive. It is easy to translate a sketch from Processing to p5 and the process of doing so begins to teach people the basics of JS. 

So (again, IMHO), P5JS has always been a bit of a toy. p5.js is much more aligned with the goals of the Processing Foundation: teach artists programming, but allow them to have *very* powerful tools once they've learned the basics.

dsblank:
@tildebyte I agree with your points. I have been working on Processing-based systems for a few years. We have a Processing.DLL that can be imported by a block language (Jigsaw), IronPython, C#, and other languages. My main focus is easily teaching computer science, using whatever language and motivations are useful. 

I use ProcessingJS because it straddles web use and Java. But because it doesn't give good error messages, it is not a good tool for the whole first semester. If I didn't need the connection to Java, I would probably use a Python-based Processing, or Javascript with P5. (I have a student looking to see if we can make better error messages). I agree that a Javascript-based language is the future. If we can make Skulpt directly import P5, then we're done!

dsblank:
Ok, I think you can merge this. There are issues that we want to address, but it may be better to deal with each of those separately.

bnmnetp:
Huh, this is now showing that we have merge conflicts.  I'm not sure how we could have gotten there. I'll have to take a look at this after lunch.

Brad

bnmnetp:
Actually it looks like the conflict is on turtle??

dsblank:
Thanks! Here are some test Jupyter notebooks: http://jupyter.cs.brynmawr.edu/hub/dblank/public/Examples/SkulptPython 

I know that there are some issues with images, both in the preloading, and with accessing the pixels in the array. But this module should be in fairly good shape moving forward.

bnmnetp:
Do we still have a problem with mouse events when there is more than one processing program on the same page?  If so we should open up a separate issue to track that.  And any other bugs known to you.

villares:
I've been dreaming of a P5JS.org  (instead of the ProcessingJS) solution wrapped in Skulpt or Brython. Here is a proof of concept using P5JS.org with Brython: http://jsfiddle.net/villares/g2sL3eza/

PR #372, opened 2015-04-09
Initial pass at divmod

jamesmallen:


waywaaard:
Thanks for the contribution, however, we need to consider a few other cases.

```Built-in function divmod( x,y ) calls x .__divmod__( y ). __divmod__ should return a pair```

For custom classes that implement the ```__divmod__``` magic method, this method should be called. You can find the appropriate places in the python source: https://hg.python.org/cpython/file/76ed5454b09d/Objects/floatobject.c#l578 

Additionally, you can check with ```a.tp$getattr("__divmod__")``` if an argument has this method and then call it with ```Sk.misceval.callsim();```. Though, there are some edges when dealing with built-in types. You may need to check for ```a.__divmod___``` or as you already did with above functions.

I hope I could give you some pointers.

**Comment by jamesmallen** [comment]
Thanks! I didn't realize __divmod__ was also a magic method. I re-implemented the builtin function to just use that as best as I could tell. I have tested it out with a variety of situations, including custom classes with the magic method and situations that cause errors, and the output appears to be functionally identical to actual Python. Test code is below.

```python
class FooString(object):
    def __divmod__(self, other):
        return 'Boo!'
    def __rdivmod__(self, other):
        return 'Hoo!'


foo = FooString()


print "testing..."
print divmod(foo, 1)
print divmod(1, foo)

print divmod(13, 5)
print divmod(1.6, 0.5)
try:
    print 'a'.__divmod__('b')
except Exception as e:
    print e
#print divmod(3, float('infinity'))
try:
    print divmod('a', 2)
except Exception as e:
    print e
try:
    print divmod(2, 'a')
except Exception as e:
    print e
try:
    print divmod(3, 0)
except Exception as e:
    print e
print "done"
```

waywaaard:
seems to be good, maybe we can get the ```binop``` tests to work.

https://hg.python.org/cpython/file/08adaaf08697/Lib/test/test_binop.py

bnmnetp:
Thanks for your work on this.  The PR does need some test cases added before it can be accepted.

The new style(such as what @waywaaard linked to) is preferred to the old numbered tests.  If you look at the tests in test/unit you will see how to adapt the standard Python unittest tests to our system.

Brad

jamesmallen:
I've added a modified version of test_binop - the only tests that work right now are the divmod tests, though, so I've commented out the others for the time being with notes about what seems to be the problem. It seems like those fixes should be addressed in another pull request, though!

albertjan:
```
test/unit/test_binop.py
ran 1467 tests, passed: 1467 failed: 0
```

- [x] some tests added :smile:

Nice work

jamesmallen:
The way the tests are counted in this implementation of unittest is different than the way they are counted in the cpython implementation - specifically, every single assert statement is considered a "test" in the skulpt testing framework, vs each method counting as a test in official unittest. Really there are just three tests currently in there. :smiley:

waywaaard:
I've tried implementing ```__float__```, ```__int__``` more correctly though there are a few edges like ```__trunc__```. Same applies for the complex type.

bnmnetp:
OK, I added a couple of tests to test divmod with floats and they work fine.

I don't think we need to make this about making all the tests in binop pass before accepting this PR.

Comments?  If not I'll accept it later today.

Brad

Issue #384, opened 2015-04-24
NotImplemented not implemented


waywaaard:
I can't find the ```NotImplemented``` that is required for the ```richcompare``` functions. Maybe we should add this.

bnmnetp:
Sorry, I have no idea what you are asking for.  the NotImplemented exeption is in errors.js but I don't think that is what you are asking for.

waywaaard:
No there is a constant ```NotImplemented```: https://docs.python.org/2/library/constants.html#NotImplemented

PR #385, opened 2015-04-24
quick stab at adding a NotImplemented constant

bnmnetp:
In reponse to #384 I put this together.  

I would like others to have a look and make sure it seems right.


waywaaard:
Seems good to me, next step would be to look where ```richCompare``` would return this constant besides in ```complex```

Issue #369, opened 2015-03-27
float builtin does not use __float__...


waywaaard:
I am currently implementing the complex number type and just wondered how to do this properly.

In CPython the complex type does implement ```nb_float```, ```nb_int```. However, skulpt does not regard those and therefore we can't also implement our own specific ```__float__``` and ```__int__```

The same for ```__round__```. Would it be sufficient to upgrade the builtins and implement ```__float__``` on complex?

Or do we need a proper mapping of ```nb$float``` to ```__float__```.

Regards

waywaaard:
The current implementation of the builtin types is completely off standard. The builtins ```int``` and ```float``` should call the appropriate magic methods and not do all in the ```constructor```.

brianpmarks:
@waywaaard Thanks for looking into this. We've got some users that are really interested in complex number support. Anything we (at trinket) can do to help out?

waywaaard:
Currently, I am in the testing phase and discovered a few edge cases while parsing complex strings, e. g., ```"-inf+nanj"```. I am looking forward fixing those soon.

waywaaard:
can be closed.

PR #387, opened 2015-04-25
Complex Type and Number improvments

waywaaard:
This PR adds a bunch of my recent additions to skulpt:
Currently passing tests: ```ran 107559 tests, passed: 107559 failed: 0``` (we have to change the way how they are counted)

Just look at the unit test ```test/uni/test_complex.py``` in my branch (https://github.com/waywaaard/skulpt/tree/epy). It also takes a few seconds (30?) for the all the complex unit tests to run.

Adds:
- [x] adds better ```Sk.python3``` support and ```print()``` as a function (this gets only called when the flag is set to ```true```
- [x] support for negative zeros in ```float```
- [x] complex builtin and type
- [x] all complex operations and methods (see ToDo)
- [x] required changes in ```ast.js``` and ```compile.js```
- [x] proper parsing of complex strings (regex FTW)
- [x] many passing unit tests
- [x] improved builtins: float/int call special methods now (se issue #369)
- [x] fixed builtin abs for complex
- [x] made appropriate adjustments in promoting number types for complex
- [x] working attributes on complex ```c.real```
- [x]  added unit tests for int
- [x]  added unit tests for dict
- [x] fixes a few issues and methods for dict

ToDo:
- [ ] keyword arguments are currently not working (sth like ```complex(real=2, imag=1)```
- [x] complex does not call ```_float__``` if argument is an object and implements this method -> we need to implement the proper PyFloatAsDouble https://github.com/python/cpython/blob/aa8ea3a6be22c92e774df90c6a6ee697915ca8ec/Objects/floatobject.c#L200 (or different name)
- [x] can't really subclass exceptions and therefore I disabled a few tests -> applies to all builtins currently, should be a own PR
- [ ] weird bug for repr/str, though it is not really affecting anything
- [x] divmod is supported though it is deprecated and could be removed -> I guess we can just keep it
- [x] look into ```test_richcompare_boundaries``` because those are failing, could be related to skulpt float impl. -> is caused by rounding and javascript float boundaries
- [x]  ```nb$coerce``` is not implemented (this builtin is not implemented and are not essential (see docs)) -> discard as it is not implemented in skulpt
- [x] use constant ```NotImplemented``` in ```richCompare``` 

While implementing the complex type, I improved a few builtin functions and some other areas. There might be a few bugs, but it should work for most cases (at least the unit tests are working).


bnmnetp:
I was able to resolve the merge conflicts pretty easily, but I get the following error on a bunch of unittests:

```
test/unit/simpleMath.py
undefined:56: ParseError: bad input on line 54
misceval.print_(err.toString() + '\n'); return $loc; } else { throw err; } } }
```

waywaaard:
I had also some unexpected behavior after merging the latest changes. I could get around them by replacing the unittest/__init__.py with my previous version which also included the required fixes. I am on it though I have a paper deadline -.-

bnmnetp:
If I comment out the ``from __future__ import print_function`` line at the top of unittest.py  everything works fine for me.

waywaaard:
I need to have a look into this future stuff again, maybe I am going to remove it from this PR and make another.

bnmnetp:
Yeah, I wouldn't mind if you separated them.  I feel like the complex numbers implementation is sufficiently done, and other things are holding us up.

waywaaard:
Sorry for the delay, I am going to remove the future stuff, for now. After merging the recent PRs, I think we are ready to go.

The ```textiowrapper``` can be made in another PR, though I made some progress there.

bnmnetp:
I would not worry about getting the kwargs to work right away.  That is part of the larger problem with builtins and kwargs.  We should figure out the solution to the larger problem first rather than working around kwargs in different ways now, and then coming back again later.

If the repr/string thing is kind of obscure then lets work on testing this and accepting it.

waywaaard:
I am currently making a new PR, and apply only the required changes. I do not want to spam the git history with reverted commits.

eah13:
Roger that.  For posterity, the new PR is #443

waywaaard:
closed in favor of #443

PR #389, opened 2015-05-05
retainglobals broken

bzwheeler:
The retainglobals flag seems to have been broken by the introduction of suspensions.  Before the breaking commit, `$loc` was being assigned to `$gbl` AFTER `Sk.retainGlobals` was checked and `Sk.globals` was applied.  After the commit, this assignment no longer took place and `$loc` remained as an empty object.  This commit simply adds the assignment of `$loc` back in the case of `Sk.retainGlobals` being true.


bzwheeler:
Reopening as a new PR with a clean history

Issue #377, opened 2015-04-15
Skulpt.org is broken in two ways


dsblank:
1. You can never run the turtle demo after clicking the `clear` link
1. Running:
```python
>>> x = 1
>>> print x
```
fails, with `NameError: name 'x' is not defined on line 1`

But this works:
```python
>>> x = 1; print x
```

waywaaard:
confirmed
1 case:
After clearing the ```turtle demo``` does not work anymore. 

2. case:
And after the turtle demo has finished and you try to use the ```interactive mode```, it fails with the error message @dsblank reported.

bnmnetp:
Just fixed case 1.  The clear button hid the canvas.  The turtle used to show the canvas but does not any more.  So we need to handle that better.

Case 2:  Fails in the way you report, but only after running something the demo above it.

@albertjan Any idea why this might be happening?  Maybe this has always been the case.

Brad

albertjan:
No it shouldn't be failing and it isn't in my browser:

```python
    Python 2.6(ish) (skulpt, Wed Apr 15 2015 10:10:02 GMT+0200 (W. Europe Daylight Time))
    [Mozilla/5.0 (Windows NT 6.3; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2272.118 Safari/537.36] on Win32
    Don't type "help", "copyright", "credits" or "license" unless you've assigned something to them
>>> x = 1
>>> print x
    1
```

@dsblank could you post the message too? So I can see which browser and version you are using,

albertjan:
@dsblank nevermind. I know what the issue is.

albertjan:
It has to do with the `Sk.retainGlobals` setting. Because every statement in the repl is a seperate skulpt program. We need to retain the global variables between runs. This is done by setting the aptly named `retainGlobals` :smile:. But because the first example and the repl share the same skulpt instance. And retainglobals is only set for the repl. 

I think the solution is to put the repl in an iframe and isolating it that way. Or maybe even make it a @trinketapp repl. Which is afaik the same code. But allready runs in an iframe.

bnmnetp:
It seems like there might be a bug with retain globals then, because I checked and Sk.retainGlobals is set to true, but after ``x=1`` there is not an x in Sk.globals.   After running an example from the first instance.

albertjan:
But we don't want to retain globals between runs in the first example.

bnmnetp:
[...]

PR #390, opened 2015-05-05
Reassign $loc to $gbl if retainGlobals flag is set

bzwheeler:
The retainglobals flag seems to have been broken by the introduction of suspensions. Before the breaking commit, $loc was being assigned to $gbl AFTER Sk.retainGlobals was checked and Sk.globals was applied. After the commit, this assignment no longer took place and $loc remained as an empty object. This commit simply adds the assignment of $loc back in the case of Sk.retainGlobals being true.


albertjan:
Oh good one I couldn't find it. This will fix the issue @dsblank raised for the website too. #377

dsblank:
Great; thanks!

PR #394, opened 2015-05-06
Adds 'Date' as a reserved name

### LINKED ISSUE #391: cannot create class named `Date`
The following code throws a `RangeError: Maximum call stack size exceeded`

```python
class Date:
  pass
```

bzwheeler:
As suggested by @waywaaard in Issue #391

 * Added Date to the _reservedNames object
 * Updated Compiler.prototype.cclass to use 'fixReservedNames' method to protect class names

Fixes #391

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -211,7 +211,8 @@ var reservedNames_ = {
     "unwatch": true,
     "valueOf": true,
     "watch": true,
-    "length": true
+    "length": true,
+    "Date": true
 };
 
 function fixReservedNames (name) {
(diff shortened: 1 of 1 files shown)


bnmnetp:
Should we go ahead and add Math, Regex, document, window, and others while we are at it?

src/compile.js
```diff
@@ -1809,7 +1810,7 @@ Compiler.prototype.cclass = function (s) {
     entryBlock = this.newBlock("class entry");
 
     this.u.prefixCode = "var " + scopename + "=(function $" + s.name.v + "$class_outer($globals,$locals,$rest){var $gbl=$globals,$loc=$locals;";
-    this.u.switchCode += "return(function " + s.name.v + "(){";
+    this.u.switchCode += "return(function " + fixReservedNames(s.name.v) + "(){";
```
bzwheeler: The more I think about this, maybe the easiest/safest solution is to not name the function at all.  As best as I can tell, the name isn't necessary.  If we want a name for stack-trace/debugging purposes, maybe we name it something descriptive yet "safe" like `"$" + s.name.v + "$_closure"`

waywaaard: Yeah I had the same thought, therefore I haven't opened a PR yet. While tinkering around it seems that class names like Number or String do no cause any trouble. As @bzwheeler stated we should use a descriptive name.

Maybe we need to have a look at properties and functions defs too?

bnmnetp:
Are you guys happy with the current solution in this PR?  If so I will accept it.

waywaaard:
I think we should use bzwheelers suggestion and use ```"$" + s.name.v + "$_closure"```

bzwheeler:
@bnmnetp - updated to use the `$` and `$_closure`  fencing, good to go from my perspective.

PR #407, opened 2015-05-18
Implemented more of the time module features

Jacco:
Implemented are:
- localtime([secs])
- asctime([t]) (works, but needs some error handling)
- ctime([secs])
- mktime(t)
- timezone
- daylight
- tzname
- clock()
- struct_time type (structseq)
- gmtime()

Not implemtented yet:
- accept2dyear
- strftime()
- strptime()
- tzset()

Also added some unit tests for this module.

This is my first pull request, so I might need some help getting things right. I am also working on the datetime module but doing time first might be a bit easier. I added the python 2.7 documentation as comments to the __init__.js. Will those be stripped from the dist files?


Jacco:
I will :-) How do you add commits to an existing pull request?

waywaaard:
just commit and push to that PR branch

Jacco:
One of the functions was throwing a unit test error because the build server is in a timezone I did not test it in. I hope I fixed it; this timezone stuff is confusing some times. It was the tm_yday field that was set to 0 in stead of 1 when calculating gmt(0). But when this PR it merged with skulpt the same spooky error might surface again probably somewhere around midnight.

Jacco:
I reworked the namedtuple I had before to structseq (which is more correct according to the cpython sources). Structsec does not surface to users but is used twice throughout the lib-code. nametuple has a few other features and is created using templating (very nice if would could do this kind of metaprogramming though!!) and "exec str in dict". Where the dict will become a real class if I am correct.

Jacco:
For namedtuple the same route could be followed though, and maybe some code could be shared between the two.

src/abstract.js
```diff
@@ -242,12 +242,12 @@ Sk.abstr.numOpAndPromote = function (a, b, opfn) {
     }
 
     if (a.constructor === Sk.builtin.lng) {
-//      if (b.constructor == Sk.builtin.nmber)
-//          if (b.skType == Sk.builtin.nmber.float$) {
-//              var tmp = new Sk.builtin.nmber(a.tp$str(), Sk.builtin.nmber.float$);
-//              return [tmp, b];
-//          } else
-//              return [a, b.v];
+        //if (b.constructor == Sk.builtin.nmber)
```
bnmnetp: I would prefer that you revert these changes.  It is so confusing when the only differences are with whitespace.   It makes it difficult to understand what is true implementation of the PR and what your editor has messed up.

Jacco: I did these changes because the build server was not giving green light. I can revert them easily because they are in separate commits.

Jacco: Should I re-do the changes in that commit concerning error.js?

bnmnetp:
Thanks,

Do you know that you can run the closure compiler tests locally using ``skulpt.py dist`` ?  This is really the best practice as you don't have to wait around for Travis.

I'm happy with what I see here.  It seems that @waywaaard  is happy too, and once I accept this PR I will simply close the old one.   I'm assuming that what you have implemented is enough to let you move forward  with datetime, and so you don't have immediate plans to do the functions on your not implemented list?

Jacco:
I will work on both modules later. The string formatting of datetime hard that of time as well.

rixner:
This pull request seems to have failing tests.  I now see the following:

test/unit/test_structseq.py
Fail: expected time.struct_time(tm_year=1970, tm_mon=1, tm_mday=1, tm_hour=0, tm_min=0, tm_sec=0, tm_wday=3, tm_yday=365, tm_isdst=0) to be equal to   time.struct_time(tm_year=1970, tm_mon=1, tm_mday=1, tm_hour=0, tm_min=0, tm_sec=0, tm_wday=3, tm_yday=1, tm_isdst=0) 
Tests failed in  StructSeqTest.test_repr

bnmnetp:
@rixner, I just reran all of the unittests when I saw this message and everything passes for me.

bnmnetp:
@rixner are you still seeing this as an open PR?   I merged it this morning.

rixner:
I am not seeing this as an open pull request.  But, when I checkout skulpt and run the tests in master, I get this error.  The problem started occurring for me after this pull request was merged.

Jacco:
This is probably caused by a timezone bug in my code I am afraid.

rixner:
I'm in CDT in the US, which is UTC -5.

bnmnetp:
Since you are UTC-5 your date was getting calculated to be December 31 resulting in daynum=365 instead of January 1.  Since I am UTC+2 right now, I was on the right day.

PR #408, opened 2015-05-18
Removes special treatment of SystemExit

LeszekSwirski:
Previously, `SystemExit` exceptions were being caught at the module level,
presumably to not crash an interpreter which did not expect this
exception. However, this had several problems:

  1. Importing a module which calls `exit()` in its body would not
     correctly exit the importing module.

  2. The additional `try`-`catch` logic required for the special treatment
     of `SystemExit` had a noticeable performance penalty.

  3. Special casing `SystemExit` on a module level is inconsistent with
     Python's behaviour, which treats it like any other exception, and
     where it is up to the interpreter (not the module) to catch this
     exception and deal with it as desired.

  4. Additionally, `SystemExit` was incorrectly set to inherit from
     `Exception` (instead of `BaseException`), which meant that except
     blocks catching `Exception` would also catch `SystemExit`

This patch fixes the above issues by:

  1. Removing the special check for `SystemExit`

  2. Changing `SystemExit` to inherit from `BaseException` (and, while I was
     at it, change other exceptions to inherit from `SystemError`)

  3. Updating the tests to special case `SystemExit` as required

  4. Adding new-style unit tests for testing the above issues

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -927,7 +927,7 @@ Compiler.prototype.outputSuspensionHelpers = function (unit) {
         }
     }
 
-    output +=  "try { $ret=susp.child.resume(); } catch(err) { if (!(err instanceof Sk.builtin.Exception)) { err = new Sk.builtin.ExternalError(err); } err.traceback.push({lineno: currLineNo, colno: currColNo, filename: '"+this.filename+"'}); if($exc.length>0) { $err=err; $blk=$exc.pop(); } else { throw err; } }" +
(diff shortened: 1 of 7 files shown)


albertjan:
I remember adding the flag for the repl here so the repl wouldn't stop running when exit was called on the webpage but it would when you run in a console using `./m repl`. Maybe late but does this still work?

albertjan:
I don't know how many actually use the cli repl.

albertjan:
@LeszekSwirski @bnmnetp this broke the build.

PR #411, opened 2015-05-19
Start of collections module

rixner:
This is an attempt to begin to implement the collections module.  It currently has implementations of ```defaultdict``` and ```Counter```.

We subclassed the builtin dict by using it as the prototype for these objects.  This does not perfectly replicate Python, but it yields working classes.  We decided this was the best approach for now, given that actually subclassing builtins is not really possible in Skulpt at this point.

We have been working on this for a bit, but the history got corrupted, so we have unfortunately had to squash the whole thing into a single commit.


bnmnetp:
This is great.  

I can accept this and then @jacco can move his namedtuple code into here.

Jacco:
This has changed a bit (see my PR). In cpython the structseq and test_structseq (which actually tests the struct_time) were used so I moved the code there. structseq and namedtuple are different in a few ways (constructor + methods + fields). The named tuple factory of cpython is actually pure python using "exec str in dict" where dict becomes a real class.

bnmnetp:
This is great, and I will accept this today, so that we can build more of the collections classes.

My only request, and this applies to #407 and others in general. I'm happy to have these partial implementations, but I think it would be good for us to use the NotImplemented exception to indicate that we are aware that there are certain functions or classes that we are not adding at this time.  It makes for a much clearer error message to the end user.

Brad

bnmnetp:
I added empty implementations of OrderedDict, deque, and namedtuple that throw NotImplemented before I merged the PR.

Jacco:
I will update the time module with NotImplemented errors.

bnmnetp:
I already did, when I accepted it this morning.

rixner:
@bnmnetp, I think we both missed the fact that there is a collections.py file still sitting in the lib directory that raises a ```NotImplementedError```.  This is in conflict with the collections directory this pull request adds.  Should I create another pull request just to get rid of that file?

bnmnetp:
I don't know how others will feel but it seems like a waste of time to do that.  When any of us could just ``git rm`` it and push.  But it doesn't seem to be causing me any trouble. so I suppose there is no urgency and setting a good example....

rixner:
It's not an issue, because I think it finds the new module first.  If the import code is changed, it could become an issue.  While it seems a little silly, I set the good example and made the pull request.  I will refrain from accepting it myself, too, even though it's trivial.

PR #412, opened 2015-05-19
correct closure comments

albertjan:
fixes the broken build after #408


bnmnetp:
OK, that is the last time I accept a PR using CodeHub on my iPad.  I could not see that it was breaking the build.

PR #415, opened 2015-05-21
Jscs style fixes

bnmnetp:
Somewhere, somehow, my development environment version of jscs has started to whine about the indentation of comments.   This has not yet infected the travis-ci builds, so things are failing for me locally, but passing the automated builds.  @albertjan as the resident expert do you have any thoughts?

This PR cleans up the indentation and makes jscs happy again.


bnmnetp:
@jacco This looks like the same set of files I was whining at you about yesterday.  Sorry...

albertjan:
I think this has to do with the saying:
> "life is like `npm install` you never know what you're gonna get". 

I'll do a pr with a `package.json` file where we can fix the jscs and jshint version so we won't be suprised again by updates to them :smile:. I was also planning on adding some more rules to jscs, to enforce the coding style more strictly. Because jshint is becoming less and less concerned with style.

PR #241, opened 2014-06-10
Test unary ~ operator for ints and longs (failing)

niklasf:
The unary operator ~ should also be supported for longs:

```
JavaScript debugger enabled
Python 2.6(ish) (skulpt, Wed Jun 11 2014 01:50:19 GMT+0200 (CEST))
[v8: 3.19.18.18] on a system
Don't type "help", "copyright", "credits" or "license" unless you've assigned something to them
>>> ~1L
TypeError: bad operand type for unary ~: 'long' on line 1
1>: evaluationresult = ~1L
2 : if not evaluationresult == None: print evaluationresult
>>> 
```

PR #416, opened 2015-05-21
fix for #241  invert longs

bnmnetp:
This should be merged with #241, as it provides the fix for the provided failing tests.


albertjan:
LGTM. I don't believe it's necessary to pull in the tests from #241 the tests from the unit test provide much better coverage. Do we want to add a bug for the imaginary numbers? :smile:

PR #420, opened 2015-05-21
Timemodule

Jacco:
Improved both time and structseq
Structseq is a bit hacky but behaves OK-ish.

Before merge I want to polish a bit more:
- use generic pad function (as @waywaaard suggested)

Next steps
- create an os environment setting for the current timezone (and maybe locale?)
- implement the time/datetime formatting using window.Intl (there is a polyfill so we should be OK)
- more tests


src/lib/time/__init__.js
```diff
@@ -40,10 +135,16 @@ var $builtinmodule = function (name) {
...
+        "Delay execution for a given number of seconds.  The argument may be\n" +
+        "a floating point number for subsecond precision.";
+
     function isLeapYear(year) {
         if((year & 3) != 0) return false;
         return ((year % 100) != 0 || (year % 400) == 0);
-    };
+    }
 
     function getDayOfYear(date) {
```
bnmnetp: I think you want to add an optional utc flag to this function.  Then from your gmtime function you can tell it to get you the day of the year using UTC.  I *think* this will fix your timezone bug and the test that is failing for @rixner back in CDT.

bnmnetp: Yep, I just verified it.  I changed the function to:

```
    function getDayOfYear(date,utc) {
        utc = utc || false;
        var dayCount = [0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334];
        var mn = utc ? date.getUTCMonth() : date.getMonth();
        var dn = utc ? date.getUTCDate() : date.getDate();
        var dayOfYear = dayCount[mn] + dn;
        if(mn > 1 && isLeapYear(utc ? date.getUTCFullYear() : date.getFullYear())) dayOfYear++;
        return dayOfYear;
    };
```
I tried it out on my machine back in Iowa and got the right behavior.

bnmnetp:
Actually it would be nice to do this as a bugfix PR before you go on to implement the formatting functions.  Otherwise the failing gmtime test is going to cause other people problems locally in their testing.

src/lib/time/__init__.js
```diff
@@ -5,19 +5,106 @@
...
+ ['__doc__', '__file__', '__name__', '__package__', 'accept2dyear', 'altzone', 'asctime', 'clock', 'ctime', 'daylight', 'gmtime', 'localtime', 'mktime', 'sleep', 'strftime', 'strptime', 'struct_time', 'time', 'timezone', 'tzname', 'tzset']
  */
 
 var $builtinmodule = function (name) {
     var mod = {};
 
-    var struct_time_fields = ['tm_year', 'tm_mon', 'tm_mday', 'tm_hour', 'tm_min', 'tm_sec', 'tm_wday', 'tm_yday', 'tm_isdst'];
+    mod.__file__ = "/src/lib/time/__init__.js";
+
+    mod.__package__ = Sk.builtin.None;
```
rixner: Python ```None``` is the singleton ```Sk.builtin.none.none$```.

src/lib/time/__init__.js
```diff
@@ -269,25 +391,71 @@ var $builtinmodule = function (name) {
...
+        "(eg. 'US/Eastern', 'Europe/Amsterdam'). Unknown timezones will silently\n" +
+        "fall back to UTC. If the TZ environment variable is not set, the local\n" +
+        "timezone is set to the systems best guess of wallclock time.\n" +
+        "Changing the TZ environment variable without calling tzset *may* change\n" +
+        "the local timezone used by methods such as localtime, but this behaviour\n" +
+        "should not be relied on.";
+
+    function strptime_f()
+    {
+        Sk.builtin.pyCheckArgs("strptime", arguments, 1, 2);
```
rixner: Given that this is not yet implemented, this will throw a ```TypeError``` if the wrong number of arguments are passed, rather than a ```NotImplementedError```.  It seems like this should be commented out until the function is implemented.

src/lib/time/__init__.js
rixner: @Jacco can you please fix this quickly and create a pull request just for the fix that I can accept?  I would like to have a working master again as soon as possible.

Jacco: I can try to fix it tonight. Quickfix would be to disable the unittest for now.

bnmnetp: I think it is a quick fix, as I've already tested it.  in UTC+ and UTC- timezones and it passes all the tests in both locations.

Jacco: Created the PR to fix it

src/lib/time.js
```diff
@@ -129,30 +170,7 @@ var $builtinmodule = function (name) {
...
-
-    /*
-    Convert a tuple or struct_time representing a time as returned by gmtime() or localtime() to a 24-character 
-    string of the following form: 'Sun Jun 20 23:21:05 1993'. If t is not provided, the current time as returned 
-    by localtime() is used. Locale information is not used by asctime().    
-    */
-    function asctime(time) {
+    function asctime_f(time) {
         if (!time || Sk.builtin.checkNone(time))
         {
```
waywaaard: Mixed bracket styles.

src/lib/time.js
```diff
@@ -270,25 +275,42 @@ var $builtinmodule = function (name) {
...
-        throw new Sk.builtin.NotImplementedError("accept2year is not implemented")
-    };
 
-    // strftime()
-    mod.strftime = function strftime() {
-        throw new Sk.builtin.NotImplementedError("strftime is not implemented")
-    };
+    function strftime_f(format, t) {
+        throw new NotImplementedError("time.strftime() is not yet implemented");
+        Sk.builtin.pyCheckArgs("strftime", arguments, 1, 2);
```
waywaaard: dead code?

src/structseq.js
```diff
@@ -56,15 +66,24 @@ Sk.builtin.make_structseq = function (module, name, fields) {
     cons.prototype.__getitem__ = new Sk.builtin.func(function (self, index) {
         return Sk.builtin.tuple.prototype.mp$subscript.call(self, index);
     });
+    cons.prototype.__reduce__ = new Sk.builtin.func(function (self) {
+        return new Sk.builtin.str("oh no");
```
waywaaard: ??

src/lib/time.js
Jacco: I can remove it. Its actually the start of the implementation.

src/structseq.js
Jacco: LOL, oops, I'll remove

src/lib/time.js
Jacco: Wow, you're thorough. Why does jslint tool not warn about that?

src/lib/time.js
waywaaard: I think we should not ship dead code, even when it is the start of an implementation. Better we ship here a clean impl. of the working functions and add the others later.

rixner: I disagree.  It depends on how useful the existing code is.  If it's a reasonable start, why make someone else start over from scratch?  As long as there are clear comments about what is there and what is not, I think there are situations where it could be okay...  @Jacco I'm not sure if this applies here or not.

waywaaard: It is, but he plans to add the code in another PR so I think we can remove it from here. I am wondering anyways why the closure compiler, jshint or jscs does not complain about unreachable code.

waywaaard:
Any updates on this one? @Jacco

meredydd:
Hi @Jacco - if you've got a moment to revise this, it'd be great to see it move forward. I keep bumping my head on this missing feature! :)

waywaaard:
@bnmnetp Is it possible to pull this in. Or should I make a new PR and merge it with the latest master branch?

albertjan:
@waywaaard we could pull it into a feature branch on skulpt

albertjan:
moved to #517

PR #421, opened 2015-05-22
Remove not implemented placeholder collections module.

rixner:
This should have been removed in #411.


bnmnetp:
This one definitely needs do further review.

Issue #400, opened 2015-05-08
Master Coordination Issue for Sprints


eah13:
Who's working on what and how do we prioritize specific features?  How will new contributors understand what would be most helpful to work on? This issue, that's how :sunglasses: 

Building off of the project management goodness of issues like #290 and #297, this is a list of scopes of work linking off to issues and/or pull requests dealing with those items. It's an index of sorts. 

If you're working on these features or interesting in working on them, please comment below or tag this issue so I can keep it updated. Thanks to everyone who contributed to the Google Doc that this started life as, including @LeszekSwirski @waywaaard @gerbal and @albertjan.

There shouldn't be too many collisions or duplicated work.  If you're worried about that, request that @bnmnetp assign you to an issue if you want people to know you're working on it. Better yet, open that pull request early so others can see what you're doing!

### Built-ins
- [ ] High: built-in class inheritance.  Issue #213
- [ ] complex()
- [ ] dict() does not support **kwargs
- [ ] eval() 
- [ ] .format()
- [ ] help(). Discussion: #249
- [ ] iterable methods. Issues: #282 and possibly #338, #335, 
- [ ] max/min don't support key arguments. Issue: #393
- [ ] open() modes. 
- [ ] property() #326 
- [ ] reversed()
- [ ] Try/except/finally
- [ ] type() Issue #97
- [ ] unichr()
- [ ] variable operators check
   - [ ] It seems some variable operators are behaving strangely (see #278). Worth a check
- [ ] With statement

### Stdlib
- [ ] calendar
- [ ] cmath. #387 may lay groundwork
- [ ] collections
- [ ] __future__. Discussion: #313
- [ ] itertools
- [ ] keyword
- [ ] math (math.degrees and math.sqrt are missing.  module needs an audit.) #math
- [ ] sys, os & argparse compatibility/overrides
   - [ ]  __platform__ (prereq to sys). Notes: #397
   - [ ] sys Issue: #300, #269 Notes:  #398
- [ ] urllib2
- [ ] urlparse  https://docs.python.org/2/library/urlparse.html
- [ ] textwrap Issue: #261
- [ ] time & datetime Issue: #243
- [ ] xml Issue: #289 

### Third Party Modules
- [ ] finish/add to numpy & matplotlib Notes: #399
- [ ] pandas
- [ ] TKinter  

### Performance/under-the-hood
- [ ] More Pervasive use of Suspensions
  - [ ] More performant 'debug mode'
[...]

PR #423, opened 2015-05-22
Implementation of OrderedDict

rixner:
This adds the ```OrderedDict``` type to collections.

I probably will not add ```deque``` and ```namedtuple``` any time soon, so others should feel free to go after those.

Trying to run the CPython tests exposes that our ```dict``` implementation is missing a lot of features that are probably not as commonly used as the core we have implemented.

### DIFF
--- a/src/lib/collections/__init__.js
+++ b/src/lib/collections/__init__.js
@@ -301,9 +301,194 @@ var $builtinmodule = function(name)
 
 
     // OrderedDict
-    mod.OrderedDict = function OrderedDict(items) {
-        throw new Sk.builtin.NotImplementedError("OrderedDict is not implemented")
+    mod.OrderedDict = function OrderedDict(items)
+    {
+        if (!(this instanceof mod.OrderedDict))
+        {
+            return new mod.OrderedDict(items);
+        }
+
+        this.orderedkeys = [];
+
+        Sk.builtin.dict.call(this, items);
+
+        return this;
     }
+
+    mod.OrderedDict.prototype = Object.create(Sk.builtin.dict.prototype);
+
+    mod.OrderedDict.prototype.tp$name = 'OrderedDict'
+
+    mod.OrderedDict.prototype.ob$type = Sk.builtin.type.makeIntoTypeObj('OrderedDict', mod.OrderedDict);
+
+    mod.OrderedDict.prototype['$r'] = function()
+    {
+        var v;
+        var iter, k;
+        var ret = [];
+        var pairstr;
+        for (iter = this.tp$iter(), k = iter.tp$iternext();
+             k !== undefined;
+             k = iter.tp$iternext()) {
+            v = this.mp$subscript(k);
+            if (v === undefined) {
+                //print(k, "had undefined v");
+                v = null;
+            }
+            ret.push("(" + Sk.misceval.objectRepr(k).v + ", " + Sk.misceval.objectRepr(v).v + ")");
+        }
+        pairstr = ret.join(", ");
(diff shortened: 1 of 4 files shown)


bnmnetp:
I have a branch on my own repo with a ``namedtuple`` implementation.

bnmnetp:
@rixner since you have just looked at the list, maybe you can file an issue with any of the missing methods you think are candidates to add.

Do you think that `move_to_end` should be part of this PR?   I don't think I've ever used OrderedDict so I don't have any use cases in mind.

rixner:
I didn't implement ```move_to_end``` because it wasn't in the documentation:

https://docs.python.org/2/library/collections.html#collections.OrderedDict

It's just in the tests.  I guess it's a "secret" undocumented method?

bnmnetp:
Ah, my mistake.  I was looking at the 3.4 documentation where it is exposed.

This is good to go then as far as you are concerned?

rixner:
Yes, I think it has everything it can have given what's in the core of Skulpt.

rixner:
@bnmnetp, the file ```dict.js``` already has comments identifying the missing dictionary methods.  I haven't really missed any of them in all of the years I've been using Skulpt.  I can file an issue, if you'd like, but I'm not sure how to prioritize which ones are actually important to add, since I haven't missed any of them...

I could open an issue with other things that cropped up from the CPython tests:

* keyword arguments to the constructor
* ```reversed``` is not implemented
* ```iter``` is not implemented
* builtins cannot be subclassed
* ```copy```/```deepcopy``` are not implemented
* Skulpt can't handle recursive data
* ```with``` statements do not work

I'm not sure of the priority of these, though.  Some of them (like the last 2 in particular) seem pretty low priority.  Maybe @eah13 should just add them to #400?

eah13:
Happy to. 

In general, unless someone will be actively working on them soon I think adding to that list is cleaner than opening more issues. But I'll follow y'all's lead on that so let me know what's helpful. 

For dictionaries specifically we have had users try to use most of the dictionary methods. They're very useful in intermediate level programs of all sorts. Copy and deep copy are the most recent request I can remember, for easily duplicating state objects.

rixner:
I am sure that all of the methods are used sometimes by some people.  It's just a question of how common that use is and how much value those methods actually add...

I agree that ```copy```/```deepcopy``` would have high value, though.  I would like those, too.  However, they are pretty complicated to implement, as they touch almost everything and require a lot of care.  In other contexts, I've even found bugs in the CPython ```deepcopy``` implementation.  Handling recursive structures that have references to themselves buried deeply is pretty complicated.  So, I don't think this is a minor task.

I don't think there is anything missing on dictionaries that can't be easily accomplished in a different way.  But, I think it would make sense to go through and actually throw a ```NotImplementedError``` on methods that are missing.  I may try to create a pull request for that.

src/lib/collections/__init__.js
```diff
@@ -301,10 +301,195 @@ var $builtinmodule = function(name)
...
+
+    mod.OrderedDict.prototype.mp$del_subscript = function(key, w)
+    {
+        var idx = this.orderedkeys.indexOf(key);
+        if (idx != -1)
+        {
+            this.orderedkeys.splice(idx, 1);
+        }
+
+        return Sk.builtin.dict.prototype.mp$del_subscript.call(this, key, w);
```
waywaaard: Why do you pass here the w parameter? I've added checks on the dict object, that checks the number of parameters. ```del_subscript``` only expects the key and the third ```w``` is not used at all.

PR #424, opened 2015-05-22
Bug fix for timemodule

Jacco:
Hope this fixes it. Followed @bnmnetp advice


Jacco:
I think it should be solved differently though. But I need to investigate.

The date created should have a timezoneOffset of 0 to begin with. Making the utc parameters unnecessary.

The localtimes date should be created with a timezoneOffset coming from os.environ

bnmnetp:
But from what I have read, you cannot make a javascript Date object other than one your browser thinks you are currently in.

bnmnetp:
As a bugfix PR, I am happy for @rixner to accept this.

Issue #326, opened 2015-02-10
property() - builtin


waywaaard:
I was fiddeling with the property() builtin and tried to implement them in a pure python way. However it seems that the custom __get__ does not get called.

```python
class Property(object):
    "Emulate PyProperty_Type() in Objects/descrobject.c"

    def __init__(self, fget=None, fset=None, fdel=None, doc=None):
        self.fget = fget
        self.fset = fset
        self.fdel = fdel
        if doc is None and fget is not None:
            doc = fget.__doc__
        self.__doc__ = doc

    def __get__(self, obj, objtype=None):
        if obj is None:
            return self
        if self.fget is None:
            raise AttributeError("unreadable attribute")
        return self.fget(obj)

    def __set__(self, obj, value):
        if self.fset is None:
            raise AttributeError("can't set attribute")
        self.fset(obj, value)

    def __delete__(self, obj):
        if self.fdel is None:
            raise AttributeError("can't delete attribute")
        self.fdel(obj)

    def getter(self, fget):
        return type(self)(fget, self.fset, self.fdel, self.__doc__)

    def setter(self, fset):
        return type(self)(self.fget, fset, self.fdel, self.__doc__)

    def deleter(self, fdel):
        return type(self)(self.fget, self.fset, fdel, self.__doc__)

class test():
    def h(self):
        return 4
    h = Property(h, doc='''returns 4''')

t1 = test()
print(t1.h) # should return 4
```

Otherwise it would be pretty straight forward to implement the property() builtin and would allow us to port a few other modules as python modules

bnmnetp:
It looks to me like __get__ and __set__ should be handled in object.js  (see around line 30) or type.js (around line 245).

I'm not totally really up to speed on descriptors, but at first look it seems like they ought to be treated as a special case of __getattr__ ??

Jacco:
I have it working with this code in object.js (in Sk.builtin.object.prototype.GenericGetAttr):

    // otherwise, look in the type for a descr
    if (descr !== undefined && descr !== null && descr.ob$type !== undefined) {
        f = descr.ob$type.tp$descr_get;
        if (!(f) && descr["__get__"])
        {
            f = descr["__get__"];
            return Sk.misceval.callsim(f, descr, this);
        }
    }

Don't know if that is the proper place though

Jacco:
At least it doesn't break any test :P

Jacco:
And with the following additions to compile.js in buildcodeobj:

    //
[...]

PR #426, opened 2015-05-22
Decorators & descriptors

Jacco:
Implementation of decorators/descriptors

- compile.js adds code to call 1 decorator (first one) on function/method (FunctionDef)
- object.js GetGenericAttr/SetGenericAttr handle descriptor __get__/__set__
- tests added

To do:
- create builtin @property/@staticmethod/@classmethod decorator/descriptor classes
- handle multiple decorators
- handle decorators on classes/types (ClassDef)

#326 
#18


Jacco:
Getting the staticmethod to work is somewhat harder:

    class staticmethod(object):
        def __init__(self, f):
            print "__init__"
            self.f = f

            print "get"
            return self.f

    class testclass:    
        @staticmethod
        def val(p1,p2):
            print str(p1)
            print str(p2)

    x = testclass()
    x.val('1', '2')
    testclass.val('3','4')

It should output:

    __init__
    get
    1
    1
    get

The problem is that to make this behavior work we can't rely on getattr but the decriptor.\_\_get\_\_ needs to be called from somewhere else.

PR #428, opened 2015-05-23
Decorators

bnmnetp:
Making a PR for the decorators feature branch back to master.  This PR can serve as the main general discussion area for this work.

This feature is a work in progress.  contributions toward the feature should be made as PRs against the decorators branch.

To do:
- [ ] create builtin @property/@staticmethod/@classmethod decorator/descriptor classes
- [ ] handle multiple decorators
- [ ] handle decorators on classes/types (ClassDef)

#326 
#18


bnmnetp:
Moving this comment from @Jacco  from the now closed PR that started the work on this feature branch

Getting the staticmethod to work is somewhat harder:

    class staticmethod(object):
        def __init__(self, f):
            print "__init__"
            self.f = f

            print "get"
            return self.f

    class testclass:    
        @staticmethod
        def val(p1,p2):
            print str(p1)
            print str(p2)

    x = testclass()
    x.val('1', '2')
    testclass.val('3','4')

It should output:

    __init__
    get
    1
    1
    get

The problem is that to make this behavior work we can't rely on getattr but the decriptor.\_\_get\_\_ needs to be called from somewhere else.

waywaaard:
Maybe we should note or maintain a list which modules (python ones) should be overhauled. I know that at least the unit tests use decorators.

PR #432, opened 2015-05-23
add setattr builtin and tests

bnmnetp:
Added the test_builtin.py for tests.

The test_builtin.py file in the Python source is a great source, and we ought to work on adding bunches of the tests for the builtins from that file to this one.


albertjan:
@bnmnetp you probably should have done a rebase on master instead of a merge with master.

PR #433, opened 2015-05-23
Unittest fixes

bnmnetp:
assertRaises did not take the correct parameters.

### DIFF
--- a/src/lib/unittest/__init__.py
+++ b/src/lib/unittest/__init__.py
@@ -131,21 +131,30 @@ class TestCase:
             msg = 'Pass'
             self.assertPassed += 1
         else:
-            msg = 'Fail: expected %s  %s ' % (str(actual),str(expected)) + feedback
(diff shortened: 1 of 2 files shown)


albertjan:
Looks good to me. @bnmnetp you feel this can be merged with one review?

bnmnetp:
Yes, I think this is pretty safe.  Other branches can now either use assertRaises, or if they worked around the old bad one they will have to update any assertRaises unit tests to do it the right way.

rixner:
Does ```assertRaises``` work correctly?  I'm not sure ```else``` clauses on ```try/except``` blocks work correctly in Skulpt.  I am seeing ```No error``` on code that clearly threw an exception.

bnmnetp:
@rixner, do you have an example for me? The no error part appeared to be working fine in my tests.  In other words, failing on a non error when there should have been one.  But I have not done extensive unittests for the unittest package.

rixner:
I screwed something up while I was trying to get the tests to run. It 
seems to work fine.  Sorry about that.

PR #436, opened 2015-05-25
Discovered how to import and use python functions from python modules…

bnmnetp:
… inside javascript modules.

### DIFF
--- a/HACKING.rst
+++ b/HACKING.rst
@@ -554,6 +554,14 @@ to remap the parameter from a javascript string to a Python string object.  Henc
 in the parameter.  Since ``iskeyword`` will return a Python bool object we need to remap that back
 to javscript for our if statement.
 
+You can use a similar strategy for creating an instance of a class:
+
+.. code-block:: javascript
+
+   var io = Sk.importModule("io", false, false);
+   var stdin = Sk.misceval.callsim(io.$d["TextIOWrapper"]);
+
+
 Seems like a lot of work to check for a keyword in an array.  But knowing how to do this for much more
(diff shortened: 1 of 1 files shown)


bnmnetp:
I question whether some of these documentation changes need multiple reviewers, but I'll follow the rules for now.

waywaaard:
I could have answered that question. Anyways, I think we should also add an example that shows how to create a new instance of a class. 

```python
var io = Sk.importModule("io", false, false);
var stdin = Sk.misceval.callsim(io.$d["TextIOWrapper"]);
```

Or at least mention it somewhere that this is also possible. You do not need to call the ```__init__``` on the object, just the class name.

bnmnetp:
Good example, I'll add it to the PR.

@waywaaard do you folllow the gitter discussion at all?   I think one problem is that most of the communication between developers only happens on the github issues/PRs.   Which is great for that, but I don't feel like other developers really monitor the google group (it very low traffic) so there is not an obvious place to ask such a question.

waywaaard:
@bnmnetp Yes I do sometimes, however I have been busy last days -> business trip.

eah13:
:+1: looks good to me.  @waywaaard's example is a helpful addition.

Going to go ahead and merge.  Personally I think docs changes like this do well to be PR'd, for helpful suggestions if nothing else.  A single reviewer could be fine too.  

If we don't want to have them go thru review then the wiki is the right answer.  Makes sense to me to have them in the repo as opposed to a wiki page for now so they're in the repo when cloned.

PR #437, opened 2015-05-25
Add namedtuples

bnmnetp:
This adds the basics of ``namedtuples`` and changes namedtuples from a package to a module by simply putting the code in collections.js rather than collections/__init__.js


bnmnetp:
Hmm, the dist tests are now giving me:
```
test/unit/test_namedtuple.py
undefined:74: ExternalError: ReferenceError: goog is not defined on line 5
c.length>0) { $err = err; $blk=$exc.pop(); continue; } else { throw err; }} }
```

Why is goog not defined??   goog.assert works for @rixner  in OrderedDict.

rixner:
I agree with @waywaaard, I do not think ```Skinherits``` belongs buried here.  If we are going to do this, it should be ```Sk.inherit``` somewhere in the core that can be called everywhere.  Then, as we improve how we inherit from builtins, this can be improved and all will benefit.  All of the classes in this module should probably standardize, at least, and use the same method, too.  I have no philosophical attachment to how I did it (it's clearly a hack, but I just wanted it done because I needed something working), so I am open to trying to standardize it in some more reasonable way.

bnmnetp:
So, I went in and looked at trying to implement kwargs (since the unittests for Cpython do test that) To make kwargs work ``cons`` must be promoted to a python function, which then breaks the inheritance hack of Skinherit including @waywaaard 's latest experiment that makes the builtins python functions along the way to making the builtin types inheritable.

I think this is worth further discussion about how we proceed so we can all work in the same direction instead of each inventing our own hacks.

rixner:
I don't use named tuples much, but isn't the use of keyword arguments to create them one of their major advantages?

bnmnetp:
I don't know, I haven't seen that many examples other than the basic

```
Point = namedtuple('Point', 'x y')
p = Point(12,15)
```

I was interested in messing with them as an alternative pre-introduction to structured data before diving into classes.  Kind of like learning ``structs`` before ``classes``

Anyway, the lack of kwarg support on all of the collections and dict is a drawback.

rixner:
I think this is a pretty normal use of them:

```
import collections

Person = collections.namedtuple('Person', 'name age gender')

print 'Type of Person:', type(Person)

bob = Person(name='Bob', age=30, gender='male')
print '\nRepresentation:', bob

jane = Person(name='Jane', age=29, gender='female')
print '\nField by name:', jane.name
```

Anyway, the lack of kwarg support is not a drawback of this particular pull request, but I do think they are used pretty frequently with named tuples.  My only issue with this pull request is that inheritance is now handled differently among the four different classes.  I don't care which way we do it (as long as it works reasonably well), but I feel like it should be consistent within the same file, at least...  Obviously, we should just come up with a more viable approach, but I don't think this pull request should be held up on that.

bnmnetp:
I think implementing keyword support for the namedtuple might be easier if I used composition rather than inheritance.

bnmnetp:
I think namedtuple is going to look different anyway, since the exposed interface, namedtuple, is a factory rather than a subclass of tuple.

rixner:
That's fair, but the factory uses a different method to make the new type a subclass of tuple than the other classes do to make themselves a subclass of dict.  But, your point makes sense.  My only question, then, is whether ```Skinherits``` belongs here or should be moved into the core.  (Or if you have decided you want to switch to composition.)

bnmnetp:
Skinherits is a pure hack inspired by choosing the path of least resistance to get some tests to pass.   But this exercise was not without merit.  It has helped me clarify a few things.

1.  Inheriting from builtin types needs to be solved in a sensible way, and to me the best path is to make the ``Sk.misceval.buildClass`` mechanism work from Javascript and for Python classes to be able to inherit from builtin types as well.  Or if that does not seem feasible then some other more general mechanism that can be widely used.
2.  Getting kwargs working in builtins and their descendants should be a priority

So, while there is no great rush to have a collections module, it would be nice to clear the current PR and then revisit the implementations of all of collections after we have solved 1 and 2.

rixner:
Okay, I'm fine with this, then, and I will accept it.

PR #441, opened 2015-05-27
Check return code on rununits subprocesses

bnmnetp:
OK, this is pretty simple, and should be accepted quickly as this plugs a hole in the new unittests.

By not checking the returncode of the subprocess we miss out on the fact that sometimes we get an error just loading up the code for a particular unittest which goes uncounted as a failure.  

If the test file exists then it better work or should cause the tests to fail.  I caught this in my own code a few days ago, and then I saw the problem again looking at @waywaaard's complex branch:

```
test/unit/test_complex.py
src/float.js:128: TypeError: a float is required at <unknown>
        throw new Sk.builtin.TypeError("a float is required");
```

This did not count as a fail and wrongly made it look like all unittests had passed.

### DIFF
--- a/skulpt.py
+++ b/skulpt.py
@@ -724,6 +724,8 @@ Sk.importMain("%s", false);
                 getFileList(FILE_TYPE_TEST))), shell=True, stdout=PIPE, stderr=PIPE)
 
         outs, errs = p.communicate()
+        if p.returncode != 0:
+            failTot += 1
         print outs
         if errs:
             print errs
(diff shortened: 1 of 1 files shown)


LeszekSwirski:
Looks sensible to me. Do you want to add an extra print statement, something along the lines of "Exited with error code %d", just in case there's nothing on stderr?

bnmnetp:
Good idea.  Just added, and checked this.

waywaaard:
I am going to merge that and see if there are any unresolved problems.

bnmnetp:
If @LeszekSwirski could accept it then you could just do another merge with the master.

LeszekSwirski:
I'm happy with it, go ahead.

PR #442, opened 2015-05-27
Hacking rst

albertjan:
Add a bit about building on windows. I have to google for this everytime I work on my windows machine . :smile: 

I also removed the old `mergesort.js`.


eah13:
Not sure how I feel about the hard line wraps.  Was that intentional? I see they happened in the 'make more readable' commit.  It will make reading in some text editors easier, but may make maintaining this document's formatting as more changes happen very difficult.  Plus, most text editors have a soft wrap mode.

I don't have a windows system to test your instructions so unfortunately that's my only comment.

albertjan:
That was intentional :smile: rst is whitespace agnostic, it doesn't care and this way you can also read it without it being rendered to html. Thats why I did it. But I'm not religious about it.

LeszekSwirski:
I would personally rather you reverted the "readability" commit and stuck with editor soft wraps. @eah13 is very correct that this makes maintenance harder, plus it nukes the git blame history for all those paragraphs.

LeszekSwirski:
As for the instructions in this document, perhaps we'd be better served by marking the expected test outputs as binary (or at least `eol=lf`) with a `.gitattributes` file.

albertjan:
@LeszekSwirski yes you are right I tried that once but a lot of other people had difficulties with their repos after it. There aren't that many people developing on windows anyway. I haven't once seen a question about it.

albertjan:
Reverted the offending commit.

albertjan:
See #122 and #117

eah13:
All looks good to me. Merging.

Issue #439, opened 2015-05-26
Operator Lookup is broken


waywaaard:
I was fiddling around with making built-ins subclassable and setting up the numeric tower. I've discovered that following simple example fails.

```python
class Foo:
    def __init__(self):
        self.a = "a"
    def __sub__(self, other):
        return "hurray"

class Bar(Foo):
    def __init__(self):
        self.b = "b"

print Bar() -1
```

This should actually output: ```hurray``` but throws an type error. The ```Sk.abstr.boNameToSlotFuncLhs_``` should also call the ```tp$getattr``` routine. However, it seems that this is not that easy. I do not really understand all of the internal mechanisms of lookup and when the type object it used and when not.

If somebody can get this working then we are one step closer to subclassable built-ins.

eah13:
Killer.  Chipping away at it.  Added a reference from #400 under built-in class extension.

bnmnetp:
Actually I think this is a simpler bug.    ``__sub__`` should be used for the RHS and ``__rsub__`` should be used for the LHS.

waywaaard:
No, the operator lookup does not use the MRO. I know all methods. Another example why this is broken.

```python
class Foo:
    def __init__(self):
        self.a = "a"
    def __sub__(self, other):
        return "hurray"

class Bar(Foo):
    def __init__(self):
        self.b = "b"

class Bah:
    def __init__(self):
        self.c = "aaa"
    
    def __sub__(self, other):
        return "bah"

print Foo() - Bah() # works __sub__ can be found on object itself
print Bar() - Foo()  # not working, MRO is not respected while searching for __sub__  
print Bar() - Bah() # same
```

bnmnetp:
Duh, you are right.  Nevermind.

bnmnetp:
Just messing around with it I came up with this strategy:

1.  Use obj.tp$getattr("__sub__") instead of bracket lookup.
2.  Because [] and tp$getattr return different objects this This required a modification to binary_op_ to detect the difference.  Maybe it would be better to return tp$getattr().im_func

Just putting this out there for discussion.

```
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -43,7 +43,7 @@ Sk.abstr.boNameToSlotFuncLhs_ = function (obj, name) {
     case "Add":
         return obj.nb$add ? obj.nb$add : obj["__add__"];
     case "Sub":
-        return obj.nb$subtract ? obj.nb$subtract : obj["__sub__"];
[...]

PR #443, opened 2015-05-27
Complex Type and Built-ins rework

waywaaard:
This is a clean version of the old complex type PR - once this is merged we can delete the old branch.

- [x] adds better ```Sk.python3``` support and ```print()``` as a function (this gets only called when the flag is set to ```true```
- [x] support for negative zeros in ```float```
- [x] complex builtin and type
- [x] all complex operations and methods (see ToDo)
- [x] required changes in ```ast.js``` and ```compile.js```
- [x] proper parsing of complex strings (regex FTW)
- [x] many passing unit tests
- [x] improved builtins: float/int call special methods now (se issue #369)
- [x] fixed builtin abs for complex
- [x] made appropriate adjustments in promoting number types for complex
- [x] working attributes on complex ```c.real```
- [x]  added unit tests for int
- [x]  added unit tests for dict
- [x] fixes a few issues and methods for dict
- [x] fixes builtins and calls magic methods if required
- [x] calling ```__float__``` and replacing the checkNumber call as it does not check for classes that just implement the float dunder method.


src/object.js
```diff
@@ -10,6 +10,34 @@ Sk.builtin.object = function () {
 };
 
 /**
+ * Special method look up. First try getting the method via
+ * internal dict and getattr. If getattr is not present (builtins)
+ * try if method is defined on the object itself
+ *
+ * Return null if not found or the function
+ */
+Sk.builtin.object._PyObject_LookupSpecial = function(op, str) {
```
waywaaard: Maybe change this to latest suggestions in #439.

src/parser.js
```diff
@@ -17,9 +17,23 @@
 function Parser (filename, grammar) {
     this.filename = filename;
     this.grammar = grammar;
+    this.p_flags = 0;
     return this;
 }
 
+// all possible parser flags
```
waywaaard: Those flags are required when we add parsing of future statements and passing flags to compile function.

src/parser.js
```diff
@@ -151,6 +165,12 @@ Parser.prototype.classify = function (type, value, context) {
     if (type === Sk.Tokenizer.Tokens.T_NAME) {
         this.used_names[value] = true;
         ilabel = this.grammar.keywords.hasOwnProperty(value) && this.grammar.keywords[value];
+
+        /* Check for handling print as an builtin function */
+        if(value === "print" && (this.p_flags & Parser.CO_FUTURE_PRINT_FUNCTION || Sk.python3 === true)) {
```
waywaaard: This line allows the use of print as a function.

albertjan:
Do we want to deprecate `Sk.python3`? This code base is never going to be python3 atleast not without a lot of work, which will most likely be destructive to the 2.7.* version. I think persuing the `__future__` stuff would be better.

waywaaard:
Well, I thought @bnmnetp and I do use this flag heavily. I am teaching and introductory course where the differences are not that big. Besides, before the future parsing stuff is not fully working we should still support this flag.

bnmnetp:
-1 on this for now.  I do use the Sk.python3 to allow me to fake the print as a function as well as for python3 style division in my CS1 courses.

bnmnetp:
@waywaaard, I see all of the boxes are ticked.  Are we ready to give this a good test and accept it?

src/import.js
```diff
@@ -396,6 +396,11 @@ Sk.importModuleInternal_ = function (name, dumpJS, modname, suppliedPyBody, canS
             }
 
             module["$d"] = modlocs;
+            
+            // empty doc string if not present
+            if (!modlocs["__doc__"]) {
+                modlocs["__doc__"] = new Sk.builtin.str("");
```
Jacco: I think in Python the default doc is None and not empty string.

bnmnetp:
Sorry @wayward, I accepted the files to dir for @rixner and they conflict with yours. Can you and @jacco have a look at the conflict.

Jacco:
I think you can accept @waywaaard s change since its much bigger than my bugfix. We can disable the unit test I added. I'll redo the work once this PR is merged. OK?

src/import.js
albertjan: Correct

albertjan:
You can easily enough merge this by hand @bnmnetp. I'm up for it once we say it's ready.

src/set.js
```diff
@@ -46,11 +45,15 @@ Sk.builtin.set.prototype["$r"] = function () {
...
+    if(Sk.python3) {
+        return new Sk.builtin.str("{" + ret.join(", ") + "}");
+    } else {
+        return new Sk.builtin.str("set([" + ret.join(", ") + "])");
+    }
 };
 Sk.builtin.set.prototype.tp$getattr = Sk.builtin.object.prototype.GenericGetAttr;
 // todo; you can't hash a set() -- what should this be?
-Sk.builtin.set.prototype.tp$hash = Sk.builtin.object.prototype.HashNotImplemented;
+Sk.builtin.set.prototype.tp$hash = Sk.builtin.none.none$;
```
albertjan:
```py
>>> a = { 1,2,3 }
>>> a
set([1, 2, 3])
>>> hash(a)
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'set'
```

src/object.js
albertjan: Might be a good idea it's now showing up on the python side of things. After I merge with master `dir()` reports it. It can be added to the internal list but it'll still be there.

albertjan:
When I start the interactive console (`./m repl`). and type `1j` or `complex("1j")` I get a `NotImplemented` exception

albertjan:
Other than these comments I think this is solid work @waywaaard ! If someone else vets this too I'll merge.

bnmnetp:
I am away from my computer for a couple of days, so I would be happy to have someone else go ahead and resolve the conflict and if all looks well accept the PR.

Then I think if we can get @rixner's change in, then we can make a new beta release on bower.  Any idea why skulpt is not showing up there anymore?  It hasn't been that long since I made an official release.

src/object.js
waywaaard: Yes I've seen this while merging, trying to fix.

src/import.js
waywaaard: Okay, I am going to change this.

src/set.js
waywaaard: That is correct, I've added a new case in the ```hash``` builtin, but it doesn't work anymore. I investigate. Setting it to ```none``` and handling this in the builtin is the cpython behavior.

waywaaard:
@albertjan I can reproduce the bug, but I still do not know what causes it. If you type 
```python
a = 1j
```
then the error does not occur. It only occurs when you try to print the number, but ```str(1j)``` and ```repr(1j)``` are working.

albertjan:
Could it be the comparison to None?

waywaaard:
Yes it is the comparison. It should return NotImplemented and not throw the NotImplementedError. I am going to fix this.

bnmnetp:
@albertjan,   This has been baking a long time.  I think it is very solid and can be merged now.  We have a lot to learn from this work, I hope that collectively we can follow through with some of the good examples here and make the rest of skulpt better.

albertjan:
@bnmnetp I agree the only thing is the set hash but we can easily enough add a new bug for that and merge this.

albertjan:
nevermind it works :smile: merged!

bnmnetp:
@albertjan, in my test, hash of set also failed in real python.  So I guess I must be missing something.

albertjan:
The hash of a set is supposed to fail. Earlier it returned None or something like it.

PR #447, opened 2015-05-27
Continue in while loop if return is pressed

eah13:
Still behaviors in the repl that don't match Python's but this small fix will avoid a nasty error for a common behavior: pressing return on an empty prompt.

### DIFF
--- a/doc/static/repl.js
+++ b/doc/static/repl.js
@@ -83,7 +83,7 @@ $(function () {
 		try {
             //Evaluate
             if (!lines || /^\s*$/.test(lines)) {
-                continue
+                return;
             }
             else {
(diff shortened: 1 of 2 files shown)


albertjan:
LGTM. Do you want to change it in `/doc/static/repl.js` too? I know I know it isn't very dry :smile:

eah13:
Ooh cool I didn't know that was there.  Well gross, kinda, but now I know where all that HTML lives

I think I'll win the award for tiniest pull requests.  Start small, they say...

albertjan:
Too quick :smile: the site repl is in a callback it's more of a cepl :tongue: So there it should be `return;`.

eah13:
Bonus: I actually tested the web REPL this time.  Works great now.  thx for catching that.

PR #448, opened 2015-05-27
Made skulpt.py host port customizable (default left at 20710)

eah13:
This is needed for some cloud IDEs and other dev setups with limited port access.  Does not alter behavior of `./m host` - but if there are arguments will take the first argument and pass it to the server to override the default port.  Catches errors from non-integer inputs.  Updated the usageString too.

Another contender for tiniest pull request evar.

### DIFF
--- a/skulpt.py
+++ b/skulpt.py
@@ -851,10 +851,9 @@ class HttpHandler(SimpleHTTPServer.SimpleHTTPRequestHandler):
         else:
             SimpleHTTPServer.SimpleHTTPRequestHandler.do_GET(self)
 
-def host():
+def host(PORT = 20710):
     """simple http host from root of dir for testing"""
     import SocketServer
-    PORT = 20710
     httpd = SocketServer.TCPServer(("", PORT), HttpHandler)
     print "serving at port", PORT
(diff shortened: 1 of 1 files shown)


bnmnetp:
I didn't know that *anyone* used the host command.

eah13:
Haha well that's either because no one has developed the Skulpt website in awhile or because there's a better workflow I don't know about. Either way, host worked great!

In other news, think a re-skinning of the site with a responsive framework like bootstrap or foundation might be worthwhile. Think its worth a few hours of hacking? It'd make it better for mobile and let visitors know the project is loved :)

bnmnetp:
Indeed, I just exchanged emails with Scott about the domain.  I had been thinking about moving the site to github.io.  I don't see any reason to continue to host the site as a google app.  It would simplify many things if we just hosted it as a github.io page.

I would love it if someone could take the lead on a nice home page.

eah13:
your wish is my #454

albertjan:
I usually just type `open /doc/index.html` :).

albertjan:
Merging btw.

PR #450, opened 2015-05-27
Test_support experiments to support test_random

eah13:
I ran into a lot of test_support calls when looking into the official `random` tests and saw test_support.py started in #310.  I need backup seeing if this is worthwhile though so thought I'd open a pull request.

The modified `test_random.py` passess all tests in Python but still doesn't seem to be importing test_support.  I added `test_support.py` to the lib/test folder expecting that to be enough.  Pardon my ignorance as I'm getting up to speed on things :)

If this sounds promising, my goal would be enough coverage of test_support within Skulpt to allow the modified test_random.py to pass, assuming that's a decent starting point for tests.  @bnmnetp is this distinct from the stuff you're doing on #310? This PR seems smaller and mergeable sooner if I understand what you're up to over there correctly.  the test_*.py files seem like the right unit of mergability rather than all of test_support or something.


bnmnetp:
@eah13  I think you are correct on a couple of points.

1.  #310 is too large, the original idea had been that we could do that as a sprint.  But I see that happening now in bits and pieces, and eventually we will reach a tipping point where it will be easy to copy in a unit test from CPython.

2.  test_support.py is a good intermediate goal.  The more of that we can get running the better.

3.  In order to do ``from test import test_support`` you need another part of the commits in #310, which I have cherry picked and resolved merge conflicts onto a new branch.  It involves a  change to import to support packages (I think @albertjan rediscovered this bug yesterday)  I can push that branch for you to look at.

eah13:
Thanks for the cherry pick.  i merged it in.

One major difference I see in this module is that [Python's random.py](https://github.com/python-git/python/blob/master/Lib/random.py) stuffs everything into a Random() object as methods and then [pulls them out at the end](https://github.com/python-git/python/blob/master/Lib/random.py#L873) to be the functions we know and love.  The CPython tests call them as methods, not functions.

Looking at the random.py source, I wonder if there's a way to get away with way more Python than we use now.  From what I understand we need the Mersienne Twister generator to be in Javascript.  But Python uses a C implementation of that algorithm.  If we could make Random.random() call that JS then a large amount of the pure Python would Just Work (tm).  Like random.uniform:

```python
    def uniform(self, a, b):
        """Get a random number in the range [a, b)."""
        return a + (b-a) * self.random()
```

I don't know a whole lot about how Skulpt could do this and honestly I'm not sure how Python hooks random.py into that random() method.  They `import _random` [here](https://github.com/python-git/python/blob/master/Lib/random.py#L69) - assuming it's there.  Could a similar strategy work where the JS becomes `_random.js` and stubbs out Random which then gets extended in Python?

My education continues....

albertjan:
Closing this because #578 fixed this.

PR #451, opened 2015-05-28
Add .gitattributes for test outputs

LeszekSwirski:
Marks test files as non-text in .gitattributes, to prevent line-ending
normalisation on them when checked out on Windows with autocrlf enabled.

Also, mark all .py files as lf-only, as this is what the parser expects (in
particular, for triple quoted strings).

- [x] windows
- [x] mac
- [x] linux

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,9 @@
+# Mark python files as text ending with lf
+*.py             text eol=lf
+
+# Mark test outputs as non-text, to prevent line-ending normalization
+/test/**/*.expect         -text
+/test/**/*.trans          -text
+/test/**/*.py.symtab      -text
+/test/**/*.py.real        -text
+/test/**/*.py.real.alt    -text


bnmnetp:
I think a couple of people should look at this and make sure it doesn't break things.

albertjan:
@LeszekSwirski for mac and linux you wouldn't have to clean up your working dir would you?

albertjan:
@LeszekSwirski I did a clone on windows and ran `.\skulpt.cmd dist` all tests failed. And after I did a `git rm --cached .` and `git reset --hard`. The tests did succeed. I don't understand why it wouldn't work immediately after a clone. Oh wait I used github for windows, that might have checked out the master branch first and then switched to the pr branch. But other than that it worked for windows.

LeszekSwirski:
@albertjan Could you try directly cloning the pr branch then, just to make sure it'll work from a clone? Also, this should not affect Mac or Linux at all.

albertjan:
I know but the last time I tried putting something like this into the repository both @rixner and @david-geo-holmes had issues with their repositories.

david-geo-holmes:
I'm not swimming in the Python sea right now. No issues from me.
Good to see so much activity on Skulpt. Following with interest...

LeszekSwirski:
As I recall from those discussions, this is because it affected d8? That .gitattributes file (and it's fixes) was slightly broken, forcing _every_ file to be both a text file and have LF endings -- this included d8, which is why it broke.

bnmnetp:
On my mac, I just did a direct clone of your (@LeszekSwirski) test-gitattributes branch.  I had no problem running the tests.

However, my memory (faulty as it may be) is that the problems started as soon as we started trying to push and do a PR, then we had all manner of merge conflicts do to line endings.

LeszekSwirski:
Well, hopefully that shouldn't be affected either -- anyone hacking on Windows will have already set `core.autocrlf` to false (otherwise their tests wouldn't be passing), which would make this .gitattributes be a no-op for them in the same way as it is for Mac and Linux.

bnmnetp:
Well, its funny how we can answer our own questions on github...  the contents of .gitattributes last time around were:

```
+# Auto detect text files and perform LF normalization
+* text=lf 
```

LeszekSwirski:
See, this doesn't auto-detect at all, it forces *all* files to be text files with LF endings. Including d8.

albertjan:
That explains things :).

bnmnetp:
Just did a check on linux on my Digital Ocean machine.  everything seems clean.

LGTM

LeszekSwirski:
Ok, I'll wait for @albertjan's approval too, since he had the cloning issue.

albertjan:
It works when I clone the branch directly. I'm merging. I completely miss understood `* text eol=lf` I thought that meant that all text files should use lf as line endings.

Issue #449, opened 2015-05-27
dir doesn't handle class hierarchies correctly


rixner:
In trying to resolve #440, I am trying to run the CPython index tests.  They use subclassing in the tests.  In our unittest module, we use ```dir``` to find all of the methods, and it is missing a bunch of methods.  The problem is illustrated here with a shorter program:

```
class Base:
    def __init__(self):
        print "init base"
        methods = []
        for name in dir(self):
            methods.append(name)

        print "methods:", methods

    def method1(self):
        pass

class Sub(Base):
    def method2(self):
        pass

sub = Sub()
```

In CPython this prints:

```
init base
methods: ['__doc__', '__init__', '__module__', 'method1', 'method2']
```

In Skulpt, this prints:

```
init base
methods: ['__module__', 'method2']
```

Obviously, ```dir``` doesn't see the base class methods. I thought this was handled by the code that walks through ```tp$mro```.  Is anyone more familiar with how subclassing currently works able to see a quick/easy fix to this?

Jacco:
I think there is a bug in Sk.builtin.dir.

It checks for the MRO on the instance instead of its type.

If an instance is passed ob$type should be used to access the MRO

PR #453, opened 2015-05-28
Bugfix dir (also removes __name__ from output)

Jacco:
#452

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -499,7 +499,7 @@ Sk.builtin.dir = function dir (x) {
 
     getName = function (k) {
         var s = null;
-        var internal = ["__bases__", "__mro__", "__class__"];
+        var internal = ["__bases__", "__mro__", "__class__", "__name__"];
         if (internal.indexOf(k) !== -1) {
             return null;
         }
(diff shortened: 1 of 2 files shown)


albertjan:
Could you also add a test for this?

bnmnetp:
test/run/t407.py incompletely tests the operation of dir, in that the tests do not check for inherited methods.

bnmnetp:
You can just add this to the current test_builtin.py  and see what happens.  Some things may need to be commented out and filed as further fixes required.

Jacco:
A *LOT* of fixes are required...

- dir can be called without parameters then is gets local definitions
- one test subclasses a Module replaces __dir__ and tests (modules cannot be subclassed yet and __dir__ is not handled yet)
- exit is not in sys
- strip is not in string

And more

albertjan:
Ah :smile: maybe it'd be better if  you just added a test that shows the inherited functions.

albertjan:
LGTM!

eah13:
I added string methods and sys.exit to #400.  Subclassing built-ins is already on there.

Aside: Wonder if we need a test task force to knock out little things like these preventing use of Python's tests as they come up. That way it's not an extra burden on specific PRs but progress gets made towards a better testing future. :notes: :notes:

Jacco:
This PR fixed #449 so it can be closed :-)

PR #456, opened 2015-05-29
Fix indexing

rixner:
This fixes the ```Sk.misceval.isIndex``` and ```Sk.misceval.asIndex``` functions and utilizes them properly in several places.  In trying to get the CPython indexing tests to run, other various things were updated, including the ```slice``` implementation and adding ```operator.__index__```.

This is not ready to be accepted.  There are a few issues:

- [x] I need #453 to be accepted to run some of the tests. (Or maybe now #443)
- [x] Some tests fail, but don't get detected by our ```unittest``` module.  The failures are because of the broken ```dir``` function, so I can't decide if ```unittest``` should have handled this differently or not.  (I don't yet know if these tests will pass or fail once ```dir``` is fixed.)
- [x] Because we don't have ```test_support.py```, I hacked the ```unittest``` module to take a list of test classes.  I'm not sure if this is the best course of action.
- [x] Need to settle on the "right" way to lookup dunder methods for ```__index__```, (#439)

I believe the actual code for indexing and slices is a strict improvement over what's already in Skulpt, so I don't think that's an issue.  The problems are all in running the tests.  I could just create simpler tests instead of trying to use CPython's tests, but that seems like a cop out.

### DIFF
--- a/src/lib/operator.js
+++ b/src/lib/operator.js
@@ -80,7 +80,6 @@ var $builtinmodule = function (name) {
     });
     mod.__floordiv__ = mod.floordiv;
 
-    // Doesn't look like anything has the __index__ magic function anyway
     mod.index = new Sk.builtin.func(function (a) {
         return Sk.builtin.assk$(Sk.misceval.asIndex(a));
     });
--- a/src/lib/operator.js
+++ b/src/lib/operator.js
@@ -82,7 +82,7 @@ var $builtinmodule = function (name) {
 
     // Doesn't look like anything has the __index__ magic function anyway
     mod.index = new Sk.builtin.func(function (a) {
-        throw new Sk.builtin.NotImplementedError("operator.index() is not yet implemented in Skulpt");
+        return Sk.builtin.assk$(Sk.misceval.asIndex(a));
     });
     mod.__index__ = mod.index;
 
--- a/src/lib/unittest/__init__.py
+++ b/src/lib/unittest/__init__.py
@@ -171,9 +171,11 @@ class TestCase:
 
 
 
-def main(verbose=False):
-    glob = globals()  # globals() still needs work
-    for name in glob:
+def main(verbose=False, names=None):
+    glob = globals() # globals() still needs work
+    if names == None:
+        names = glob
+    for name in names:
         if issubclass(glob[name],TestCase):
             try:
                 tc = glob[name]()
--- a/src/list.js
+++ b/src/list.js
@@ -256,11 +256,11 @@ Sk.builtin.list.prototype.sq$repeat = function (n) {
     var j;
     var i;
     var ret;
-    if (!Sk.builtin.checkInt(n)) {
+    if (!Sk.misceval.isIndex(n)) {
         throw new Sk.builtin.TypeError("can't multiply sequence by non-int of type '" + Sk.abstr.typeName(n) + "'");
     }
 
-    n = Sk.builtin.asnum$(n);
+    n = Sk.misceval.asIndex(n);
     ret = [];
     for (i = 0; i < n; ++i) {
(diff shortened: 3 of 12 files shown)


bnmnetp:
OK, I accepted #453 as more than one of us had given it the OK.

I think unittest and test_support will be a work in progress for some time to come.  It was relatively easy to get the basics working, but of course the CPython unittests are far from simple in many cases.

rixner:
Agreed.  That leaves the question of whether I should hack ```main``` in ```unittest.py``` the way I did or if there a better solution for the short term?

rixner:
This is basically waiting for someone to review it and offer opinions.  One specific point: @waywaaard and @bnmnetp do we have a consistent way of looking up dunder methods that I should be using to find ```__index__```, or should I leave it as I have done.

bnmnetp:
I will take a look at it today.

My hope is to get both yours and @waywaaard's approved.  These continue to move us forward.

As of now, I think we have an idea for a consistent way of handling the dunder methods, but unless @waywaaard has been implementing it on his branch, I don't think we have an implementation yet.

waywaaard:
I've changed all my lookups to ```f = Sk.builtin.type.typeLookup(obj.ob$type, "__dunder__");``` and call it. Though we need to handle the ```obj == null``` case before passing obj to the lookup.

bnmnetp:
Do you mean "call it" using Sk.misceval.callsim or does typeLookup return a javascript callable function already?

bnmnetp:
@waywaaard:   It looks like you have wrapped Sk.builtin.type.typeLookup inside PyObject_specialLookup_  everywhere except two places (complex constructor and the PyFloat_AsDouble methods.  

It seems to me that we ought to declare that PyObject_lookupSpecial_ becomes the standard way of dealing with dunder lookups?  It does already check to see if  obj is null.

bnmnetp:
@waywaaard, @albertjan:

I have run all the tests, and used the build in my data structures and intro books.  All of the code in those books continues to work

As I look over the code I'm amazed at how much stuff has been cleaned up.  And yet, how far we have to go.

@waywaaard you made a lot of changes to how the dunder methods in the dict class are exposed.  Should that be the way that list, and set and tuple are implemented going forward?  My understanding is that this is a step towards inheritance of builtin types.

I am sure I am missing some things.  If you could maybe summarize your suggestions for us that would be nice.

bnmnetp:
@albertjan, this one has been baking a long time.  I see no reason not to accept it.

bnmnetp:
@rixner, while clearly not wrong, I feel like we ought to standardize the lookups of special methods on:

```
Sk.builtin.object.PyObject_LookupSpecial_ 
```

The only thing is the name.  We have avoided naming things with the PyObject or PyXXX prefix so far. This is a change from that convention.  But I have no objections to using it until we decide on something better.  -- See my comments on the API that I posted earlier today.

rixner:
```object.PyObject``` is incredibly redundant.  Just like we have cleaned things up with consistent style checking, it seems like we should shoot for consistent naming.  I think it makes more sense for this to be something like ```Sk.builtin.object.lookupSpecial```.  Given that it's already been merged in, though, I will use the existing name.  But, I do think it needs to be changed.

rixner:
I think this is done now and ready to be accepted after it has been reviewed.

waywaaard:
@bnmnetp Yes, that is a step towards inheritance. The problem is that I still do not know how to automatically map the javascript functions to python functions with working parameters and et cetera.

I've chosen the ```Sk.builtin.object.PyObject_LookupSpecial_ ``` because this is what is used in cpython and allows for me a better mapping of cpython code to my impl. However, we can still rename it without any problems.

And yes I mean by calling the ```Sk.misceval.callsim``` as all dunder methods should be bound ones.

PR #461, opened 2015-05-31
unpack reverse before comparison

### LINKED ISSUE #459: sort(reverse=False) acts as if it is sort(reverse=True)
Run the following code in Skulpt:

x = ['a', 'x', 'm']
x.sort()
print x
x.sort(reverse=True)
print x
x.sort(reverse=False)
print x

The output will be:

['a', 'm', 'x']
['x', 'm', 'a']
['x', 'm', 'a']

The first two are correct - the last one is incorrect.  I am guessing this is an easy fix.  :)

albertjan:
Fixes #459

### DIFF
--- a/src/list.js
+++ b/src/list.js
@@ -409,10 +409,12 @@ Sk.builtin.list.prototype.list_sort_ = function (self, cmp, key, reverse) {
     var has_cmp = cmp !== undefined && cmp !== null;
     var rev;
 
-    if (reverse === undefined || reverse === Sk.builtin.none.$none) {
+    if (reverse === undefined) {
         rev = false;
+    } else if (reverse === Sk.builtin.none.none$) {
+        throw new Sk.builtin.TypeError("an integer is required");
     } else {
-        rev = Sk.ffi.remapToJs(reverse);
+        rev = Sk.ffi.remapToJs(Sk.builtin.bool(reverse));
     }
 
     timsort = new Sk.builtin.timSort(self);
--- a/src/list.js
+++ b/src/list.js
@@ -440,13 +440,13 @@ Sk.builtin.list.prototype.list_sort_ = function (self, cmp, key, reverse) {
         };
     }
 
-    if (reverse) {
+    if (Sk.ffi.remapToJs(reverse)) {
         timsort.list.list_reverse_(timsort.list);
     }
 
     timsort.sort();
 
-    if (reverse) {
+    if (Sk.ffi.remapToJs(reverse)) {
         timsort.list.list_reverse_(timsort.list);
     }
 
--- a/src/list.js
+++ b/src/list.js
@@ -414,7 +414,7 @@ Sk.builtin.list.prototype.list_sort_ = function (self, cmp, key, reverse) {
     } else if (reverse === Sk.builtin.none.none$) {
         throw new Sk.builtin.TypeError("an integer is required");
     } else {
-        rev = Sk.ffi.remapToJs(Sk.builtin.bool(reverse));
+        rev = Sk.misceval.isTrue(reverse);
     }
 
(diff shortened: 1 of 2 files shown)


waywaaard:
The current use of the ```reverse``` variable is a little bit confusing. It is used a variable for python values and for javascript values. You set it to ```false``` when it undefined. The impl. is working, but we could improve the style.

albertjan:
You feel the `list.prototype.list_sort_` should not use `reverse` as a argument?

waywaaard:
I think that the use of the variable ```reverse``` is inconsistent. I would expect that ```reverse``` stores either and Python Object or is undefined/null. But we set it to ```false``` and still unpack the value. Maybe a few comments would make the function more readable.

It is just a suggestion. Do we have style guideline for the use of variable? So that variables can be used for javascript objects and python ones (ofc they are also javascript objects)?

albertjan:
Ah yes you're right. Might be good to formulate some guides around this. I will change it to make it clearer.

rixner:
I agree that calling ```remapToJS``` with something that is already a JS value is bad style.  I would be okay with something like this, though:

```
if (reverse === undefined || reverse === Sk.builtin.none.none$) {
       reverse = false;
} else {
       reverse = Sk.ffi.remapToJs(reverse);
}
```

Then it is a Python value before this and a Javascript value after this.  This kind of thing occurs throughout the code in many other places, where the same variable is used to hold the remapped Javascript value.

src/list.js
```diff
@@ -407,8 +407,12 @@ Sk.builtin.list.prototype.list_sort_ = function (self, cmp, key, reverse) {
     var timsort;
     var has_key = key !== undefined && key !== null;
     var has_cmp = cmp !== undefined && cmp !== null;
-    if (reverse === undefined) {
-        reverse = false;
+    var rev;
+
+    if (reverse === undefined || reverse === Sk.builtin.none.$none) {
```
rixner: Python ```None``` is ```Sk.builtin.none.none$```.

albertjan: Yeah I've found out some other things while I looked into this :smile: fix coming soon.

test/unit/test_list_sort.py
```diff
@@ -0,0 +1,34 @@
...
+        x = [1,2,3]
+        x.sort(reverse=False)
+        self.assertEqual(x, [1,2,3])
+
+    def test_revserNoneShouldThrowError(self):
+        x = [1,2,3]
+        try:
+            x.sort(reverse=None)
+        except TypeError as e:
+            self.assertEqual(str(e), "TypeError: an integer is required on line 19")
```
albertjan: `e.message` would be better here but it doesn't exist see #464

albertjan:
@rixner I fixed the `None` stuff and made the reverse param work more like cpython.

src/list.js
```diff
@@ -407,8 +407,14 @@ Sk.builtin.list.prototype.list_sort_ = function (self, cmp, key, reverse) {
...
     var has_cmp = cmp !== undefined && cmp !== null;
+    var rev;
+
     if (reverse === undefined) {
-        reverse = false;
+        rev = false;
+    } else if (reverse === Sk.builtin.none.none$) {
+        throw new Sk.builtin.TypeError("an integer is required");
+    } else {
+        rev = Sk.ffi.remapToJs(Sk.builtin.bool(reverse));
```
rixner: This is correct, but would it be better to just use ```Sk.misceval.isTrue``` directly here?  That saves a round trip mapping to Python and back to Javascript.  Sorry for not pointing that out earlier...

rixner:
I think this looks good now and I can accept this, unless you agree that using ```isTrue``` is the right way to go.

albertjan:
I agree :smile:

PR #462, opened 2015-05-31
don't use string.trimLeft use goog.string.trimLeft

### LINKED ISSUE #460: Need workaround for missing trimLeft() and trimRight() in IE
Skulpt uses trimLeft() and trimRight() in its runtime.  In particular using the python split() function uses trimLeft() and fails on IE as a result.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/TrimRight
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/TrimLeft

This StackOverflow (sort of) covers the issue:

http://stackoverflow.com/questions/2308134/trim-in-javascript-not-working-in-ie

The code I added to my own scripts that workaround this without changing Skulpt are:

```js
if(typeof String.prototype.trimLeft !== 'function') {
    String.prototype.trimLeft = function() {
        return this.replace(/^\s+/,"");
    }
}

if(typeof String.prototype.trimRight !== 'function') {
    String.prototype.trimRight = function() {
        return this.replace(/\s+$/,"");
    }
}
```

I think it would be best if Skulpt came up with a simplar workaround or something else so split() works on IE.

albertjan:
Fixes #460

### DIFF
--- a/src/str.js
+++ b/src/str.js
@@ -331,7 +331,7 @@ Sk.builtin.str.prototype["split"] = new Sk.builtin.func(function (self, on, howm
     regex = /[\s]+/g;
     str = self.v;
     if (on === null) {
-        str = str.trimLeft();
+        str = goog.string.trimLeft(str);
     } else {
         // Escape special characters in "on" so we can use a regexp
         s = on.v.replace(/([.*+?=|\\\/()\[\]\{\}^$])/g, "\\$1");


albertjan:
@skulpt/committers can one of you have a look at this and at #461

waywaaard:
I am not on the committers list, but I think this one is ready to merge. ```goog.string.trimleft``` just executes a regular replace expression ```return str.replace(/^[\s\xa0]+/, "");```. This should also handle some of the non-standard whitespaces.

rixner:
This looks fine.

PR #470, opened 2015-06-02
KeyboardInterrupt Exception and Compiler Handling

waywaaard:
With the addition of suspensions, long-running python programs or infinite loops cannot be terminated. I am using skulpt to program some virtual Arduino and need to terminate the execution of skulpt as most of the programs run in a loop.

Python has a KeyboardInterrupt Exception that is thrown when you press Ctrl+C or Del. This PR tries to mimic this behavior by adding the exception and a runtime check if the ```Sk.keyboardInterrupt``` flag has been set. When set it throws the KeyboardInterrupt Exception and, therefore, terminates the execution of the program.

I am currently using this code and it works pretty well with suspensions and promises. However, I am open for other approaches or solutions.

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -287,6 +287,20 @@ Compiler.prototype.outputInterruptTest = function () { // Added by RNL
     return output;
 };
 
+/**
+ * Function to test if an interrupt should occur if external Sk.keyboardInterrupt flag has been set
+ * This function is executed at every test/branch operation.
+             if("keyboardInterrupt" in Sk && Sk.keyboardInterrupt === true) {
+                throw new Sk.builtin.KeyboardInterrupt("aborted execution");
+            }
+ */
+Compiler.prototype.keyboardInterrupt = function () {
+    var output = "";
+    output += "if (Sk.hasOwnProperty('keyboardInterrupt') && Sk.keyboardInterrupt === true) {throw new Sk.builtin.KeyboardInterrupt('aborted execution');}";
+
+    return output;
+};
+
 Compiler.prototype._jumpfalse = function (test, block) {
     var cond = this._gr("jfalse", "(", test, "===false||!Sk.misceval.isTrue(", test, "))");
     out("if(", cond, "){/*test failed */$blk=", block, ";continue;}");
@@ -1600,6 +1614,7 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
     // New switch code to catch exceptions
     this.u.switchCode = "while(true){try{"
     this.u.switchCode += this.outputInterruptTest();
+    this.u.switchCode += this.keyboardInterrupt();
     this.u.switchCode += "switch($blk){";
     this.u.suffixCode = "} }catch(err){ if (!(err instanceof Sk.builtin.BaseException)) { err = new Sk.builtin.ExternalError(err); } err.traceback.push({lineno: currLineNo, colno: currColNo, filename: '"+this.filename+"'}); if ($exc.length>0) { $err = err; $blk=$exc.pop(); continue; } else { throw err; }} }});";
 
@@ -1841,6 +1856,7 @@ Compiler.prototype.cclass = function (s) {
     }
     this.u.switchCode += "while(true){";
     this.u.switchCode += this.outputInterruptTest();
+    this.u.switchCode += this.keyboardInterrupt();
     this.u.switchCode += "switch($blk){";
     this.u.suffixCode = "}break;}}).apply(null,$rest);});";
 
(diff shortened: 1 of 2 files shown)


LeszekSwirski:
I think this would be possible with Suspensions alone, by throwing an exception in the handler of the Suspension. I'm always hesitant to add special behaviour like this to the compiler when really it belongs in the running interpreter.

waywaaard:
Yeah, I think that would be a better way of implementing. As long as it behaves the way as this. As I've stated this is open for discussion. But that would allow people to bind to Ctrl+C on a webpage or in a specific context and mimic the same behavior.

meredydd:
@LeszekSwirski, how would you do this with suspensions alone? If you mean "set the debugging flag, and use the debug-break predicate function to throw an exception if Ctrl+C is pressed", then yes, that would work, but it would run on every statement rather than at every (non-fallthrough) basic block.

Might we consider a more generalised mechanism for "call this check function once in a while, but it doesn't have to be every statement", that could be used for both this and the timeout mechanism?

LeszekSwirski:
I meant with yield suspensions rather than debug suspensions, although I see your point that this would be a less common check than every loop (then again, in some ways that's good?).

Perhaps a general interrupt mechanism would be a nicer solution though, where the timeout checks would be integrated into that.

meredydd:
Oh, of course. Yes, yield suspensions would seem to be the thing to hang
off here :)

And yes, a generalised "check for interrupts" mechanism of which yield
suspensions are a special case is something I'd be happily in favour of.

waywaaard:
I would like to see such a behavior and how would it be possible to add the keyboard interruption in there? I guess I am not the best candidate of implementing this.

LeszekSwirski:
The way I would do it would be to add a universal suspension handler, using '*' to specify that the handler will be called for any suspension, as part of your `Sk.misceval.asyncToPromise` call. This handler would check the value of some 'keyboard interrupt' flag; if true, it would return `Promise.reject(new Sk.builtin.KeyboardInterrupt)`, otherwise it would return null to fallthrough to the default behaviour for that suspension.

waywaaard:
I finally had some time and figured it out by myself.

The interrupt can work with no modification of the compiler and just requires a custom suspension handler:

```javascript
        var interruptHandler = function (susp) {
            if (Sk.hardInterrupt === true) {
                throw new Sk.builtin.KeyboardInterrupt('aborted execution');
            } else {
                return null; // should perform default action
            }
        };
```

and use this here:

```javascript
        // async execution of skulpt code
        this.myPromise = Sk.misceval.asyncToPromise(function() {
            that.isRunning = true;
            return Sk.importMainWithBody("<stdin>",false,code,true);
        }, {"*": interruptHandler});
```

This requires that the ```KeyboardInterrupt``` is available:

```javascript
/**
* Is thrown when the execution has been stopped by Ctrl+C or Del. The compiler
* checks for an Sk.keyboardinterrupt flag, if set the exception is thrown.
* @constructor
* @extends Sk.builtin.BaseException
* @param {...*} args
*
*/
Sk.builtin.KeyboardInterrupt = function (args) {
    var o;
    if (!(this instanceof Sk.builtin.KeyboardInterrupt)) {
        o = Object.create(Sk.builtin.KeyboardInterrupt.prototype);
        o.constructor.apply(o, arguments);
        return o;
    }
    Sk.builtin.BaseException.apply(this, arguments);
};
Sk.abstr.setUpInheritance("KeyboardInterrupt", Sk.builtin.KeyboardInterrupt, Sk.builtin.BaseException);
```

alexwenbj:
> I finally had some time and figured it out by myself.
> 
> The interrupt can work with no modification of the compiler and just requires a custom suspension handler:
> 
> ```js
>         var interruptHandler = function (susp) {
>             if (Sk.hardInterrupt === true) {
>                 throw new Sk.builtin.KeyboardInterrupt('aborted execution');
>             } else {
>                 return null; // should perform default action
>             }
>         };
> ```
> 
> and use this here:
> 
> ```js
>         // async execution of skulpt code
>         this.myPromise = Sk.misceval.asyncToPromise(function() {
>             that.isRunning = true;
>             return Sk.importMainWithBody("<stdin>",false,code,true);
>         }, {"*": interruptHandler});
> ```
> 
> This requires that the `KeyboardInterrupt` is available:
> 
> ```js
> /**
> * Is thrown when the execution has been stopped by Ctrl+C or Del. The compiler
> * checks for an Sk.keyboardinterrupt flag, if set the exception is thrown.
> * @constructor
> * @extends Sk.builtin.BaseException
> * @param {...*} args
> *
> */
> Sk.builtin.KeyboardInterrupt = function (args) {
>     var o;
>     if (!(this instanceof Sk.builtin.KeyboardInterrupt)) {
>         o = Object.create(Sk.builtin.KeyboardInterrupt.prototype);
>         o.constructor.apply(o, arguments);
>         return o;
>     }
>     Sk.builtin.BaseException.apply(this, arguments);
> };
> Sk.abstr.setUpInheritance("KeyboardInterrupt", Sk.builtin.KeyboardInterrupt, Sk.builtin.BaseException);
> ```

@ebertmi Hi,I want to stop the execution manually,too.I try to use your workaround to accomplish it.What I did is:
1. Add the `KeyboardInterrupt` in errors.js;
2. Add `interruptHandler` function and use it in myPromise as you show in my page;
My question is,If I want to add a button ,when clicking to stop the execution, what can I do?I have a try to change the `Sk.hardInterrupt`:
```
function stop(){
    Sk.hardInterrupt = true;
}
```
But it can't stop the execution, it only prevent to restart the second execution.How can I do for that?

alexwenbj:
It works. @ebertmi Thank you.

alexwenbj:
@ebertmi Excuse me, I wonder why this PR did not merge to master? It is useful to stop the application.Thank you.

PR #472, opened 2015-06-02
Add optional uncaughtException handler to Sk.configure

### LINKED ISSUE #455: some Turtle errors not surfacing
Minimal example (also [here](https://trinket.io/library/trinkets/257774ef0e)):

```python
import turtle

tina = turtle.Turtle()
tina.shape("turtle")

def bad_function():
  # no x and y args - should throw error
  print "fails silently"

tina.onclick(bad_function)
```

More code from a trinket user showing this in action is [here](https://trinket.io/library/trinkets/8d1dfab54a)

The errors end up in the Javascript console.  I think I remember @bzwheeler saying this has something to do with how turtle uses suspensions?  But very hazy on that.

bzwheeler:
Any event driven skulpt callbacks can result in uncaught exceptions.
This PR adds the ability to specify an uncaughtException handler when setting up Skulpt
which will allow the implementer to surface those errors to the end user.

This method must be called explicitly in whatever js library is generating the suspensions.
See the changes in turtle.js of this PR for an example of how it can be used.

```python
import turtle

def click():
  pass

s = turtle.Screen()
t = turtle.Turtle()
t.shape('turtle')
s.onclick(click)
```

If the page implementer specifies an Sk.uncaughtException handler,
the error thrown in the above example will be sent to the handler.

Fixes #455

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -84,7 +84,8 @@ goog.exportSymbol("Sk.configure", Sk.configure);
  */
 Sk.uncaughtException = function(err) {
     throw err;
-}
+};
+goog.exportSymbol("Sk.uncaughtException", Sk.uncaughtException);
 
 /*
  *	Replaceable message for message timeouts
--- a/src/env.js
+++ b/src/env.js
@@ -27,6 +27,9 @@ Sk.configure = function (options) {
     Sk.debugout = options["debugout"] || Sk.debugout;
     goog.asserts.assert(typeof Sk.debugout === "function");
 
+    Sk.uncaughtException = options["uncaughtException"] || Sk.uncaughtException;
+    goog.asserts.assert(typeof Sk.uncaughtException === "function");
+
     Sk.read = options["read"] || Sk.read;
     goog.asserts.assert(typeof Sk.read === "function");
 
@@ -75,6 +78,13 @@ Sk.configure = function (options) {
     Sk.misceval.softspace_ = false;
 };
 goog.exportSymbol("Sk.configure", Sk.configure);
+
+ /*
+ * Replaceable handler for uncaught exceptions
+ */
+Sk.uncaughtException = function(err) {
+    throw err;
+}
 
 /*
  *	Replaceable message for message timeouts
--- a/src/env.js
+++ b/src/env.js
@@ -79,7 +79,7 @@ Sk.configure = function (options) {
 };
 goog.exportSymbol("Sk.configure", Sk.configure);
 
- /*
+/*
  * Replaceable handler for uncaught exceptions
  */
 Sk.uncaughtException = function(err) {
--- a/src/lib/processing.js
+++ b/src/lib/processing.js
@@ -1551,8 +1551,13 @@ var $builtinmodule = function (name) {
 
                 mod.frameCount = processing.frameCount;
                 if (Sk.globals["draw"]) {
-                    Sk.misceval.callsim(Sk.globals["draw"]);
-		}
+                	try {
+                   	    Sk.misceval.callsim(Sk.globals["draw"]);
+                    }
+                    catch(e) {
+                        Sk.uncaughtException(e);
+                    }
+				}
             };
 
             var callBacks = ["setup", "mouseMoved", "mouseClicked", "mouseDragged", "mouseMoved", "mouseOut",
@@ -1560,7 +1565,7 @@ var $builtinmodule = function (name) {
             ];
             for (var cb in callBacks) {
                 if (Sk.globals[callBacks[cb]]) {
-                    processing[callBacks[cb]] = new Function("Sk.misceval.callsim(Sk.globals['" + callBacks[cb] + "']);");
+                    processing[callBacks[cb]] = new Function("try {Sk.misceval.callsim(Sk.globals['" + callBacks[cb] + "']);} catch(e) {Sk.uncaughtException(e);}");
                 }
             }
         }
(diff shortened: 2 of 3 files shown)


meredydd:
This seems turtle-specific (and liable to mislead people who think that setting this option will catch all uncaught exceptions, as opposed to "uncaught exceptions in turtle javascript callbacks"). Is there any turtle-specific config we could use instead of making this Skulpt-global?

bzwheeler:
I don't think it is turtle specific, this should be able to benefit any skulpt library that reacts to asynchronous interaction like keyboard/mouse events.  For example, if you introduce a runtime error into the render loop of processing, it silently fails.  With this PR (and an update in processing to call the handler :wink:), we can actually surface this error.

I started with a use case in turtle because we had a direct customer request for it at trinke however I think it will prove valuable to processing and possibly other Skulpt addons over time.

LeszekSwirski:
I, like @meredydd, don't feel comfortable with this sort of special casing. Why does the exception get swallowed in the first place? Surely the event loop (or whatever is calling the handler) should catch it, and report it?

bzwheeler:
@LeszekSwirski, @meredydd: Thanks for your concern with this PR, I'll try to explain why I don't think this is special casing.

Essentially, as @LeszekSwirski suggested, this PR is catching the error where the handler is being called and is then reporting the error by calling the uncaughtException handler.  The same concept can (and should when I have a bit more time) be applied to the render loop as well as the key/mouse handlers in the processing module.

These errors are currently swallowed because the top level promise chain that would usually allow you to catch exceptions when evaluating a Skulpt program is independent from the promise chain that is generated via a user triggered action like mouse/key events.  In turtle, the promise that is generated by the skulpt call to 'applyAsync' when handling a user event now has a catch statement and uses the Sk.uncaughtException as a hook for reporting the error.

This isn't an issue that is unique to Skulpt, it is the nature of Promises.  Any promise that is rejected and is not explicity handled will silently fail.  There is not, AFAIK, a generic place to always catch a promise because a promise is temporal and you can't generically predict when it is a safe time to assume that the a rejection is going to be unhandled (see [here](https://github.com/cujojs/when/blob/master/docs/api.md#rejections-handled-later) for a better description if that is unclear). This means that we have to handle the promise chain at its inception which can only be known by the code that handles the event (turtle,processing,etc).

I don't consider this feature to be special casing, it is simply a way for modules to report errors that would otherwise be uncaught. I'd be happy to consider an alterative implementation if one can be proposed.

meredydd:
If this patch had also applied to (to take your suggestions) the render loop, key handlers, mouse handlers in `processing`, etc, then it probably would look a little less special-cased - as it is, it's advertising itself as a global option and in fact affects only one particular place an exception could go unhandled! The existence of other places this could be useful makes me more positively inclined to this PR, although we had better be careful to catch 'em all (that is, at all the call sites included in modules included with Skulpt) if that's what we're advertising we'll do.

More generally, as you point out, "catch every unhandled exception" isn't something Skulpt can do in the general case, because you can't tell what's "unhandled" in the Promise world (a major omission failing, but so it goes). This is therefore something that needs to be considered at every point where Javascript calls into Skulpt (which all of our users' code does at least once!), and if we provide a global flag it feels like we're advertising something we can't offer.

How are the turtle and processing modules initialised? Presumably they require some degree of Javascript setup in order to bind them to actual HTML and DOM elements on which they can then handle events. Would it make sense for this per-module Javascript setup operation to provide an unhandled-exception callback, instead of making it Skulpt-global? That way it's just a contract between that module (eg turtle) and the user, which the module can be confident it will fulfil.

LeszekSwirski:
The expand on that slightly, there is already some config in place for the turtle module, `Sk.TurtleGraphics`, which the initialisation uses to call back to the parent context (in particular, finding the DOM element which will be the container for the canvas). This could be extended to include an error handler function, which by default could show and alert or log to console or something like that. Then, it is the system's responsibility to wire that handler up appropriately, as part of the setup of the turtle environment.

bzwheeler:
@meredydd:  Good points, I should have been more thorough and added it to the other places, which I have now done! I just pushed up a commit that added the Sk.uncaughtException for the processing module as well.  As far as I know, that is the only other module that uses a render loop or responds to user events at this time.  I think it is reasonable to expect that as maintainers of this repo, we can expect all async javascript calls to be properly wrapped with exception handlers.  This doesn't mean that we WILL always properly wrap them all, everyone makes mistakes, but when we (or a user) finds an omission, it should be reported as a bug and fixed.  In my opinion, just because we can't guarantee a perfect implementation doesn't mean we shouldn't offer a feature.

@LeszekSwirski: I would much rather this be a globally available feature and not module specific.  Right now there are only 2 modules that benefit but there is no reason why any number of future modules couldn't have similar needs.  We could ask that each module offer its own config option for error handling  but that could become confusing and inconsistent over time.

bnmnetp:
I really think a general mechanism for surfacing errors that are swallowed up inside a Suspension is a good addition.  As a programmer, it is very surprising behavior that you just lose obvious exceptions with no good explanation as to where they went or why.

eah13:
In support of this being a global feature, I think any project that uses the killer visual interactivity that Skulpt offers is going to need something like this. A clone of Tkinter or wx in Skulpt, for instance, would need this to surface errors- wouldn't that be an awesome third party module to see developed?! These are exactly the kinds of projects that Emscripten-based solutions will struggle to offer, and key to Skulpt's unique value.

albertjan:
In lieu of a more full fledged solution to handling uncaught exceptions, this looks to be a solution that solves the problem at hand. I'm for merging this for now and if some thing better comes we can revert this easily enough.

bnmnetp:
@meredydd @LeszekSwirski are you OK with this being accepted now or do you still have objections?  If there are no further objections then I am of the mind to accept it.

LeszekSwirski:
I'm mildly against, but not enough to block a merge.

Issue #469, opened 2015-06-02
Instance with __getitem__ should be iterable


Jacco:
```python
class iterable:

    def __getitem__(self, i):
        if i < 13:
            return i
        else:
            raise IndexError("index out of range")

x = iterable()
    
for v in x:
    print v
```

Skulpt gives error: class iterable is not iterable

rixner:
I don't think you even need ```__len__```.  It is sufficient to just have ```__getitem__``` to be iterable.

Jacco:
Hm, indeed, didn't test that. Updated the issue to make it simpler :-)

bnmnetp:
Sadly, the error is thrown here where we are looking for an internal definition of ``tp$iter``

```javascript
Sk.abstr.iter = function (obj) {
    if (obj.tp$iter) {
        return obj.tp$iter();
    }
    else {
        throw new Sk.builtin.TypeError("'" + Sk.abstr.typeName(obj) + "' object is not iterable");
    }
};
```

bnmnetp:
Interestingly, I always thought that for an object to be iterable that it had to conform to the iterable protocol and provide an ``__iter__`` and ``next`` method.

Going a little deeper, I have a fix, but its a hack.  It seems like maybe the compiler out to generate different code...

```
        klass.prototype.tp$iter = function () {
            var ret;
            var iterf = this.tp$getattr("__iter__");
            var tname = Sk.abstr.typeName(this);
            if (iterf) {
                ret = Sk.misceval.callsim(iterf);
                return ret;
            }
            var gitem = this.tp$getattr("__getitem__");
            if (gitem) {
                this.iteridx = 0
                return this;
            }
            throw new Sk.builtin.TypeError("'" + tname + "' object is not iterable");
        };
        klass.prototype.tp$iternext = function () {
            var iternextf = this.tp$getattr("next");
            if (iternextf) {
                return Sk.misceval.callsim(iternextf);
            }
            var gitem = this.tp$getattr("__getitem__");
            if (gitem) {
                print(arguments[0])
                ret = Sk.misceval.callsim(gitem,this.iteridx);
                this.iteridx++;
                return ret;
            }
        };
```

waywaaard:
CPython impl. uses a check a check first:

```c
PyObject *
PyObject_GetIter(PyObject *o)
{
    PyTypeObject *t = o->ob_type;
    getiterfunc f = NULL;
    f = t->tp_iter;
    if (f == NULL) {
[...]

PR #473, opened 2015-06-03
Unpack sequences

### LINKED ISSUE #430: Unpacking doesn't check lengths
Currently, things are just blindly unpacked.  For example, the following statements do not match the behavior of Python:

```
lst = [1, 2, 3]
a, b = lst
a, b, c, d = lst
a, b = 3
```

I think the issue is in ```Compiler.prototype.ctuplelistorset```.  On a store, I think this should probably inject code that checks the length of the ```data``` argument to see if it matches the number of items on the left hand side.

rixner:
Use iteration at run time to unpack sequences to match CPython's behavior.

This does not currently work with user-defined classes that support iteration (#469).  However, if the solution to that preserves the internal iteration interface, then they should just work once that is resolved.  Those tests from CPython are in the ```test_unpack.py``` file, but are commented out.

Resolves #430

### DIFF
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -571,6 +571,37 @@ Sk.abstr.sequenceSetSlice = function (seq, i1, i2, x) {
     }
 };
 
+// seq - Python object to unpack
+// n   - JavaScript number of items to unpack
+Sk.abstr.sequenceUnpack = function (seq, n) {
+    var res = [];
+    var it, i;
+
+    if (!Sk.builtin.checkIterable(seq)) {
+        throw new Sk.builtin.TypeError("'" + Sk.abstr.typeName(seq) + "' object is not iterable");
+    }
+
+    for (it = seq.tp$iter(), i = it.tp$iternext(); 
+         (i !== undefined) && (res.length < n); 
+         i = it.tp$iternext()) {
+        res.push(i);
+    }
+
+    if (res.length < n) {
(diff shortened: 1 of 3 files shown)


albertjan:
LGTM! I say merge.

bnmnetp:
I merged our two PR's together and uncommented the tests you had commented out.  After making the following change:

```
it = Sk.builtin.object.getIter_(seq)
```

All the tests pass.  ``The Sk.builtin.object.getIter_`` is patterned after the same function that @waywaaard pointed out.  This should be the standard "iter finder" rather than tp$iter.

bnmnetp:
If everyone is in agreement, I believe I can simply push this merged/uncommented/changed in one push.

rixner:
Does this mean we need to universally replace calls to ```tp$iter``` with ```Sk.builtin.object.getIter_```?

bnmnetp:
No, definitely not.  There are many places where tp$iter is used to refer to the iterator for one of the builtins.     There may be other places, like your example dictionary example where getIter_ should be used, but its not global.

I think there is a problem wherein EVERY object has tp$iter defined for it.  So whereas CPython can check to see if it is defined for an object we cannot.

You probably saw that the compiler does insert getIter_ now where tp$iter used to be.

rixner:
Won't your ```getIter_``` just handle that gracefully, though, and return ```obj.tp$iter()``` when appropriate?  Then, if/when builtins are ever subclassed, things will just "work" (hopefully?).

PR #475, opened 2015-06-03
Betteriter  -- Fixes 469

bnmnetp:
All tests except for the dict update unit test test pass.  Not sure why but there is a comment about Skulpt previously not supporting something with iteration.  So that seems likely.  @rixner any ideas?

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -28,11 +28,14 @@ Sk.builtin.object.PyObject_LookupSpecial_ = function(op, str) {
 };
 
 
+
 Sk.builtin.object.getIter_ = function(obj) {
+    var iter;
+    var getit;
+    debugger;
     if (obj.tp$iter) {   // builtin  or check for tp$nextiter
         return obj.tp$iter();
     }
-
     if (obj.tp$getattr) {
         iter = obj.tp$getattr("__iter__");
         if (iter) {
@@ -50,8 +53,8 @@ Sk.builtin.object.getIter_ = function(obj) {
                 var ret = Sk.misceval.callsim(this.myobj["__getitem__"], this.myobj, this.idx);
                 this.idx++;
                 return ret;
-            }
-        }
+            };
+        }();
     }
     throw new Sk.builtin.TypeError("'" + Sk.abstr.typeName(obj) + "' object is not iterable");
 };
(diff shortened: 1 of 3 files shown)


bnmnetp:
Well, there is more broken than I thought.  I'll try to sort this out in the morning.

waywaaard:
It seems that the exception, that is maybe thrown in ```__getitem__``` is swallowed.

You can test it with:

```python
class badseq(object):
    def __iter__(self):
        return self
    def __next__(self):
        raise Exception()

{}.update( badseq())

class FailingUserDict:
    def keys(self):
        class BogonIter:
            def __init__(self):
                self.i = ord('a')
            def __iter__(self):
                return self
            def __next__(self):
                if self.i <= ord('z'):
                    rtn = chr(self.i)
                    self.i += 1
                    return rtn
                raise StopIteration
        return BogonIter()
    def __getitem__(self, key):
        raise Exception

{}.update(FailingUserDict())
```

rixner:
@bnmnetp, I think I dealt with the exceptions correctly now.

rixner:
I also think that ```iteridx``` needs to be turned into a Python object before it is passed to ```__getitem__```.

bnmnetp:
OK, I have a cleaned up and fully working version now, that is much closer to the CPython version.

bnmnetp:
If I can get an LGTM I would like to accept both 473 and 475 simulataneously.  I uncommented @rixner's tests and made one small change in his code (See the PR) to make these two play nicely together.

albertjan:
Other then my comment it looks good to me :smile:

albertjan:
Euhm @bnmnetp did you see my comment?

bnmnetp:
Yes, good catch.  It was fixed in the push I just did.

albertjan:
Ah ok :smile:

bnmnetp:
It was kind of cool that since the two PR's were co-dependent, I could merge and accept both of them together.

rixner:
Wait, shouldn't you have used ```PyObject_LookupSpecial_``` to find the dunder methods instead of ```tp$getattr```? Didn't we just decide that was the right way of doing things?

bnmnetp:
You are probably right.  I was just following the lead of the existing code and not thinking hard enough.

rixner:
Also, is ```Sk.builtin.checkIterable``` now "broken"?

(This is great, by the way!  These are small things that can easily be fixed.)

bnmnetp:
Well, looking at the code, and based on what I know, I would say that it was always broken.  But now I might know enough to fix it.

PR #477, opened 2015-06-04
Cleanups:

bnmnetp:
* name change of PyObject_LookupSpecial to just lookupSpecial.
* use lookupSpecial instead of tp$getattr in iteration code

### DIFF
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -823,29 +823,6 @@ goog.exportSymbol("Sk.abstr.iternext", Sk.abstr.iternext);
 
 
 /**
- * @constructor
- */
-function seqIter(obj) {
-    var ret;
-    this.idx = 0;
-    this.myobj = obj;
-    this.getitem = Sk.abstr.lookupSpecial(obj, "__getitem__");
-    this.tp$iternext = function () {
-        try {
-            ret = Sk.misceval.callsim(this.getitem, this.myobj, Sk.ffi.remapToPy(this.idx));
-        } catch (e) {
-            if (e instanceof Sk.builtin.IndexError) {
-                return undefined;
-            } else {
-                throw e;
-            }
-        }
-        this.idx++;
-        return ret;
-    };
-}
-
-/**
  * Get the iterator for a Python object  This iterator could be one of the following.
  * This is the preferred mechanism for consistently getting the correct iterator.  You should
  * not just use tp$iter because that could lead to incorrect behavior of a user created class.
@@ -864,6 +841,32 @@ Sk.abstr.iter = function(obj) {
     var iter;
     var getit;
     var ret;
+
+    /**
+     * Builds an iterator around classes that have a __getitem__ method.
+     * 
+     * @constructor
+     */
+    var seqIter = function (obj) {
+        this.idx = 0;
+        this.myobj = obj;
+        this.getitem = Sk.abstr.lookupSpecial(obj, "__getitem__");
+        this.tp$iternext = function () {
+            var ret;
+            try {
+                ret = Sk.misceval.callsim(this.getitem, this.myobj, Sk.ffi.remapToPy(this.idx));
+            } catch (e) {
+                if (e instanceof Sk.builtin.IndexError) {
+                    return undefined;
+                } else {
+                    throw e;
+                }
+            }
+            this.idx++;
+            return ret;
+        };
+    };
+
     if (obj.tp$getattr) {
         iter =  Sk.abstr.lookupSpecial(obj,"__iter__");
         if (iter) {
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -822,6 +822,21 @@ function seqIter(obj) {
     };
 }
 
+/**
+ * Get the iterator for a Python object  This iterator could be one of the following.
+ * This is the preferred mechanism for consistently getting the correct iterator.  You should
+ * not just use tp$iter because that could lead to incorrect behavior of a user created class.
+ *
+ * - tp$iter
+ * - A user defined `__iter__` method
+ * - A user defined `__getitem__` method
+ *
+ * @param obj
+ *
+ * @throws {Sk.builtin.TypeError}
+ * @returns {Object}
+ */
+
 Sk.abstr.iter = function(obj) {
     var iter;
     var getit;
@@ -854,7 +869,7 @@ goog.exportSymbol("Sk.abstr.iter", Sk.abstr.iter);
  * internal dict and getattr. If getattr is not present (builtins)
  * try if method is defined on the object itself
  *
- * Return null if not found or the function
+ * @returns {null|Object} Return null if not found or the function
  */
 Sk.abstr.lookupSpecial = function(op, str) {
     var res;
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -807,9 +807,10 @@ function seqIter(obj) {
     var ret;
     this.idx = 0;
     this.myobj = obj;
+    this.getitem = Sk.abstr.lookupSpecial(obj, "__getitem__");
     this.tp$iternext = function () {
         try {
-            ret = Sk.misceval.callsim(this.myobj["__getitem__"], this.myobj, Sk.ffi.remapToPy(this.idx));
+            ret = Sk.misceval.callsim(this.getitem, this.myobj, Sk.ffi.remapToPy(this.idx));
         } catch (e) {
             if (e instanceof Sk.builtin.IndexError) {
                 return undefined;
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -438,7 +438,7 @@ Sk.abstr.sequenceContains = function (seq, ob) {
     special = Sk.abstr.lookupSpecial(seq, "__contains__");
     if (special != null) {
         // method on builtin, provide this arg
-        return Sk.misceval.callsim(special, seq, ob);
+        return Sk.misceval.isTrue(Sk.misceval.callsim(special, seq, ob));
     }
 
     if (!Sk.builtin.checkIterable(seq)) {
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -465,21 +465,43 @@ Sk.abstr.sequenceConcat = function (seq1, seq2) {
 
 Sk.abstr.sequenceGetIndexOf = function (seq, ob) {
     var seqtypename;
+    var i, it;
+    var index;
     if (seq.index) {
         return Sk.misceval.callsim(seq.index, seq, ob);
     }
+    if (Sk.builtin.checkIterable(seq)) {
+        index = 0;
+        for (it = Sk.abstr.iter(seq), i = it.tp$iternext();
+             i !== undefined; i = it.tp$iternext()) {
+            if (Sk.misceval.richCompareBool(ob, i, "Eq")) {
+                return Sk.builtin.assk$(index, Sk.builtin.nmber.int$);
+            }
+            index += 1;
+        }
+        throw new Sk.builtin.ValueError("sequence.index(x): x not in sequence");
+    }
 
     seqtypename = Sk.abstr.typeName(seq);
-    if (seqtypename === "dict") {
-        throw new Sk.builtin.NotImplementedError("looking up dict key from value is not yet implemented (supported in Python 2.6)");
-    }
     throw new Sk.builtin.TypeError("argument of type '" + seqtypename + "' is not iterable");
 };
 
 Sk.abstr.sequenceGetCountOf = function (seq, ob) {
     var seqtypename;
+    var i, it;
+    var count;
     if (seq.count) {
         return Sk.misceval.callsim(seq.count, seq, ob);
+    }
+    if (Sk.builtin.checkIterable(seq)) {
+        count = 0;
+        for (it = Sk.abstr.iter(seq), i = it.tp$iternext();
+             i !== undefined; i = it.tp$iternext()) {
+            if (Sk.misceval.richCompareBool(ob, i, "Eq")) {
+                count += 1;
+            }
+        }
+        return Sk.builtin.assk$(count, Sk.builtin.nmber.int$);
     }
 
     seqtypename = Sk.abstr.typeName(seq);
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -435,7 +435,7 @@ Sk.abstr.sequenceContains = function (seq, ob) {
      *  Look for special method and call it, we have to distinguish between built-ins and 
      *  python objects
      */
-    special = Sk.builtin.object.PyObject_LookupSpecial_(seq.ob$type, "__contains__");
+    special = Sk.builtin.object.lookupSpecial_(seq, "__contains__");
     if (special != null) {
         // method on builtin, provide this arg
         return Sk.misceval.callsim(special, seq, ob);
@@ -612,7 +612,7 @@ Sk.abstr.objectFormat = function (obj, format_spec) {
     }
 
     // Find the (unbound!) __format__ method (a borrowed reference)
-    meth = Sk.builtin.object.PyObject_LookupSpecial_(obj.ob$type, "__format__");
+    meth = Sk.builtin.object.lookupSpecial_(obj, "__format__");
     if (meth == null) {
         throw new Sk.builtin.TypeError("Type " + Sk.abstr.typeName(obj) + "doesn't define __format__");
     }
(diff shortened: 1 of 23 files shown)


bnmnetp:
@rixner There are 188 occurrences of tp$iter  just for fun, I replaced all instances of   foo.tp$iter() in dict.js with Sk.builtin.object.getIter_(foo)

Ran the tests, and all heck broke loose.  only 518/559 of the orginial tests passed Failures were for a variety of reasons.  I haven't really poked at it any further.

rixner:
Okay, clearly that's not a trivial change, then.

bnmnetp:
Today I've mostly been thinking about the API and structuring some kind of project to document skulpt.  Which has made me question the name.

I don't think that ``lookupSpecial`` really belongs in ``Sk.builtin.object``  It seems like it should be part of just ``Sk.builtin``  or ``Sk.builtin.abstr``  The same with ``Sk.builtin.object.getIter_``

rixner:
I agree.  I think these both belong in ```Sk.builtin.abstr``` based on the current structure of the code.   Currently ```Sk.builtin.abstr``` contains the utility functions that work for different groups of objects.

bnmnetp:
Well, @rixner As the commit message says, ``getIter_`` has now come full circle and replaced the Sk.abstr.iter that has always been there.  The good thing is that Sk.abstr.iter is used in a number of places by the compiler, so we should be getting the benefits of the improved function in more places.

I think this is ready, for review and merge now.

src/function.js
```diff
@@ -60,7 +60,22 @@ Sk.builtin.checkSequence = function (arg) {
...
+    var ret = false;
+    if (arg !== null ) {
+        try {
+            ret = Sk.abstr.iter(arg);
+            if (ret) {
+                return true;
+            }
+        } catch (e) {
+            if (e instanceof Sk.builtin.TypeError) {
+                return null;
```
rixner: Should this return ```false```?

bnmnetp:
I believe the old version returned null, so I followed suit.

bnmnetp:
Ooops no, you are correct it returned:

```
(arg !== null && arg.tp$iter !== undefined);
```

rixner:
Looking at the code, it all looks good to me other than the ```null```/```false``` issue.  I would like to spend some time running things, though, but I don't have time right now...

rixner:
I think this is a definite move in the right direction.  

However, it seems like the builtins that take iterables (like ```any```, ```all```, and ```sum``` for instance) should now use ```Sk.abstr.iter``` instead of ```tp$iter``` to get an iterator.  If I modify those three functions to do that, then those functions seem like they work better with no further changes.  I feel like it is easy enough to go through the builtins that take iterables that this should just be done here.

I think the list in ```builtin.js``` is just ```any```, ```all```, ```sum```, ```zip```, ```map```, ```reduce```, and ```filter```.  I guess the constructors of the builtin types, as well.

If others disagree that this should be done here, then I obviously have no objections to the code inside this pull request, as it facilitates doing that, so this can be merged.  I have no serious objection to arguing that updating the builtins belongs in a different pull request.

bnmnetp:
@rixner, I have no objection to using this PR to further clean up iteration.  iterfixes seems an apt name for the branch.

One question for you that I missed earlier:  On line 184 of type.js you added:  e instanceof Sk.builtin.StopIteration    But I find no definition for StopIteration anywhere.

bnmnetp:
Its one thing to make the changes and feel confident that we are making things better, but I suppose in all fairness, we really ought to construct test cases that actually demonstrate that this works with no other side effects.

Construct lists, dictionaries, sets and use your builtins with:

- A user defined class with __getitem__
- A user defined class with __iter__ and next

I think it might be find to start unit tests for the basic builtin types, and one could certainly add test cases to test_builtin.  There are probably cases that you can copy and paste in their as I deleted huge chunks of test_builtin when I was trying to get some stuff to test.

bnmnetp:
Nevermind, I see that the compiler does StopIteration as a special case by intercepting the raise StopIteration and returning undefined

rixner:
The fact that ```undefined``` is used instead of ```StopIteration``` to end the iteration predates my exposure to Skulpt.  That has been in there forever, and I think there are warts on top of warts because of it.

src/abstract.js
```diff
@@ -793,16 +793,93 @@ Sk.abstr.sattr = function (obj, nameJS, data, canSuspend) {
...
+Sk.abstr.iternext = function (it, canSuspend) {
+    return it.tp$iternext(canSuspend);
+};
+goog.exportSymbol("Sk.abstr.iternext", Sk.abstr.iternext);
+
+
+/**
+ * @constructor
+ */
+function seqIter(obj) {
```
albertjan: This is a bit weird a global function just sitting around here couldn't this be added to Sk.abstr also?

albertjan: Oh wait this an iter object. But then is should be called something like `Sk.seqIter` or something. Or will it only ever be used from Sk.abstr.iter?

rixner:
@bnmnetp would you like me to commit directly to the branch or make pull requests against it?

bnmnetp:
Everything I've done is on the branch. So go ahead and commit directly. But it might be good to coordinate files. Let me. Know what you are working on   I have list in progress and dict is done.

rixner:
I updated the builtins that take iterables and added the CPython tests for those functions.  Unfortunately, the CPython tests are not as comprehensive as one might hope.  They don't test the different kinds of iterables very well.  Perhaps that's somewhere else that I haven't found yet or I'll have to write my own.

bnmnetp:
OK, list, tuple, dict, and set are all done.  

I did not change the constructor to set as it already takes its argument and makes a list out of it before doing any further processing.  That seems kind of strange to me, but I would want to deal with that separately rather than doing a major rewrite of the set constructor.

rixner:
I found ```test_iter.py``` and am looking at it.  Many of the tests seem relevant.  After commenting out the tests for things we don't support, it passes 9 tests and fails 34.  A lot of the failures are because ```iter``` is not yet implemented.  I'm not sure how easy/hard it will be to implement ```iter``` at this point.

rixner:
```Sk.builtin.complex.prototype.tp$richcompare``` returns a python wrapped object.  It is supposed to return Javascript ```true``` or ```false```.  This breaks one of the CPython iteration tests.  Fixing it, however, breaks some of the complex unit tests.  Not sure what to do...

**Comment by waywaaard** [comment]
@rixner the problem is that I need to return either ```NotImplemented``` (not the exception) or true and false. Maybe we have to use ```Sk.misceval.isTrue``` until we have fixed the other places?

**Comment by rixner** [comment]
@waywaaard, you can't unilaterally change the behavior of that function, as it is supposed to follow a contract that callers expect.  Furthermore, it is an internal function whose return value is never exposed to Python, so it makes little sense to wrap it as a Python object, forcing callers to immediately unwrap it.  Therefore, I think it is the complex implementation that needs to be fixed to return Javascript objects.

If we need to change the contract of that function to possibly return a third Javascript value in addition to ```true```/```false```, so be it.  But, that is orthogonal to returning Javascript vs. Python objects.  If you need to convert to Python objects internally, I think that should happen in the complex code after calling the function, as that is unique to that situation.

Furthermore, changing the contract requires fixing all call sites to check for the third possible return value.

The following fails:

```print 3j in [1, 2, 3]```

**Comment by waywaaard** [comment]
@rixner Yes, you are right - I've missed that. I try to figure out how we can use the best from both sides. CPython does use the third return type to indicate that some comparisons are not implemented and they return this in their ```richcompare``` impl.

Should I maintain 2 versions to be compatible with the obvious incorrect skulpt implementation and the other one that can be called using ```__eq__``` etc? Or should I see if we can fix the other one?

**Comment by rixner** [comment]
What is the meaning of the "not implemented" return value?  If ```richCompareBool``` gets this from ```tp$richcompare``` does it then convert it into an exception?  If it also has to return this value, then that is going to have a pretty big ripple effect throughout the code, I fear.

Also, ```richCompareBool``` is a performance critical function, so I am concerned about slowing down the common case to deal with an uncommon case.

**Comment by waywaaard** [comment]
```NotImplemented```: Special value which can be returned by the “rich comparison” special methods (```__eq__()```, ```__lt__()```, and friends), to indicate that the comparison is not implemented with respect to the other type.


see https://docs.python.org/2/library/constants.html#NotImplemented

**Comment by bnmnetp** [comment]
And here's a nice example of why this seemingly useless constant is actually useful:  http://shahriar.svbtle.com/python-notimplemented-type

**Comment by rixner** [comment]
I see.  The ultimate problem are these lines of code in ```richCompareBool```:

```
    op2method = {
        "Eq"   : "__eq__",
        "NotEq": "__ne__",
        "Gt"   : "__gt__",
        "GtE"  : "__ge__",
        "Lt"   : "__lt__",
        "LtE"  : "__le__"
    };

    method = op2method[op];
    swapped_method = op2method[Sk.misceval.swappedOp_[op]];

    if (v[method]) {
        res = Sk.misceval.isTrue(Sk.misceval.callsim(v[method], v, w));
        if (res != Sk.builtin.NotImplemented.NotImplemented$) {
            return res;
        }
    } else if (w[swapped_method]) {
        res = Sk.misceval.isTrue(Sk.misceval.callsim(w[swapped_method], w, v));
        if (res != Sk.builtin.NotImplemented.NotImplemented$) {
            return res;
        }
    }
```

For the builtin types, this ends up being a call to ```tp$richcompare```, which all return Javascript types (except for complex).  The methods need to be found using ```lookupSpecial``` and they could be Python code, so in that case, they would return Python types.  I think this block of code needs to just change to the following and all other ```tp$richcompare``` functions need to return Python values:

```
    method = Sk.abstract.loookupSpecial(v, op2method[op]);
    swapped_method = Sk.abstract.lookupSpecial(w, op2method[Sk.misceval.swappedOp_[op]]);

    if (method) {
        res = Sk.misceval.callsim(method, v, w);
        if (res != Sk.builtin.NotImplemented.NotImplemented$) {
            return Sk,misceval.isTrue(res);
        }
    }
    if (swapped_method) {
        res = Sk.misceval.callsim(swapped_method, w, v);
        if (res != Sk.builtin.NotImplemented.NotImplemented$) {
            return Sk.misceval.isTrue(res);
        }
    }
```

Correct?

**Comment by rixner** [comment]
If we make those changes, the subsequent lookups of ```__cmp__``` should be similarly changed to use ```lookupSpecial```.

**Comment by rixner** [comment]
I just pushed what I had so you can see all of the changes.  We can revert this, move it elsewhere, whatever, if need be.  I did not change all of the ```tp$richcompare``` functions to return Python values and just cheated by calling ```isTrue``` as @waywaaard suggested. These should just be hooked into the dunder methods anyway.  I think @mchat is doing that in her #481 work.

**Comment by bnmnetp** [comment]
I think using `isTrue` is a decent compromise for now.  When @mchat is all done then we can revisit this issue.

**Comment by rixner** [comment]
Can ```__cmp__``` return ```NotImplemented```?

**Comment by waywaaard** [comment]
I am not sure the documentation says that ```__cmp__``` may call ```__eq__``` and this can return ```NotImplemented```

>Called by comparison operations if rich comparison (see above) is not defined. Should return a negative integer if self < other, zero if self == other, a positive integer if self > other. If no __cmp__(), __eq__() or __ne__() operation is defined, class instances are compared by object identity (“address”). See also the description of __hash__() for some important notes on creating hashable objects which support custom comparison operations and are usable as dictionary keys. (Note: the restriction that exceptions are not propagated by __cmp__() has been removed since Python 1.5.)

**Comment by bnmnetp** [comment]
In my quick test, if `__cmp__` returns NotImplemented I get the error  `TypeError: comparison did not return an int`

**Comment by bnmnetp** [comment]
Except that my test was too quick, as I was using an "old style" class instead of a new style class.  

This test works:

```
class A(object):
    def __init__(self,a):
        self.a = a
        
    def __cmp__(self, other):
        if type(other) == A:
            print 'compare A with A'
            return self.a - other.a
        else:
            print "cannot compare"
            return NotImplemented

class B(object):
    def __init__(self,a):
        self.a = a
        
    def __cmp__(self, other):
        if type(other) == A or type(other) == B:
            print 'compare B with A'
            return self.a - other.a

a = A(10)
b = B(5)
print type(a)
print "a < a"
print a < a
print " b < a"
print b < a
print "a < b"
print a < b

```

**Comment by bnmnetp** [comment]
Which inevitably leads to the question.  Does skulpt really distinguish between old style and new style classes?  It seems to me that all classes inherit from object in skulpt making them new style.???

**Comment by rixner** [comment]
Currently passing 17 and failing 24 of the iter tests from CPython.  The failures are all due to ```iter``` not being implemented (although there obviously may be other lurking bugs, as well).

**Comment by bnmnetp** [comment]
Scott,  I will leave this in your hands, as I am not going to have good internet for the next 11 days.

Brad

**Comment by rixner** [comment]
Really the only thing I see that we could do at this point is to implement ```iter```, and I don't really have time for that right now.  So, the question is whether this should sit until someone (me, you, or someone else) implements ```iter``` or if I should comment out those test cases, add the working tests and have people review the pull request...  Anyone else have any thoughts?

**Comment by bnmnetp** [comment]
I think that `iter` would make sense as its own PR.  We get into these internal things and we can keep on pulling threads for months!

**Comment by rixner** [comment]
I believe this pull request is now ready for review.  I would appreciate it if someone would review it for Brad and I so that we can get it accepted.

**Comment by rixner** [comment]
I should add that an implementation of ```iter``` still remains missing.  If/when someone takes that on, there are many tests in ```test_iter.py``` that should be uncommented.

File: src/abstract.js (hunk 2 above)
Comment (rixner) [inline src/abstract.js]: @albertjan ```seqIter``` is a constructor for an iteration object.  Such objects are only ever created and returned by ```Sk.abstr.iter```.  I think you are right that this shouldn't be global, but I'm not sure if we have an established convention for where to place such a constructor.

**Comment by albertjan** [comment]
I can review it tomorrow.

File: src/abstract.js (hunk 2 above)
Comment (albertjan) [inline src/abstract.js]: couldn't we just move it inside `Sk.abstr.iter`? I'll check when I review this tonight. We haven't establishes such a convention, are there many of these "classes"? I don't remember any. :smile:.

**Comment by albertjan** [comment]
Loving this! :smile:

```py
>>> class Test():       
... 	def __init__(self):
... 		self.list = [ 1,2,3 ]
...     def __iter__(self):
... 		yield 5
... 		for i in self.list:
... 			yield i
... 
>>> a = Test()
>>> for b in a:
... 	print b
... 
5
1
2
3
```
  
About the weird `seqIter` class shouldn't we monkey patch the `tp$iter` function into the classes instead of wrapping them in a weird mini class? 

That would make the `Sk.abst.iter` class look like this:

```js
Sk.abstr.iter = function(obj) {
    var iter;
    var getit;
    var ret;

    var seqIter = function seqIter(obj) {
        obj.idx = 0;
        obj.myobj = obj;
        obj.getitem = Sk.abstr.lookupSpecial(obj, "__getitem__");
        obj.tp$iternext = function () {
            var ret;
            try {
                ret = Sk.misceval.callsim(obj.getitem, obj.myobj, Sk.ffi.remapToPy(obj.idx));
            } catch (e) {
                if (e instanceof Sk.builtin.IndexError) {
                    return undefined;
                } else {
                    throw e;
                }
            }
            obj.idx++;
            return ret;
        };
        return obj;
    };

    if (obj.tp$getattr) {
        iter =  Sk.abstr.lookupSpecial(obj,"__iter__");
        if (iter) {
            return Sk.misceval.callsim(iter,obj);
        }
    }
    if (obj.tp$iter) {
        try {  // catch and ignore not iterable error here.
            ret = obj.tp$iter();
            if (ret.tp$iternext) {
                return ret;
            }
        } catch (e) { }
    }
    getit = Sk.abstr.lookupSpecial(obj, "__getitem__");
    if (getit) {
        // create internal iterobject if __getitem__
        return seqIter(obj);
    }
    throw new Sk.builtin.TypeError("'" + Sk.abstr.typeName(obj) + "' object is not iterable");
};
goog.exportSymbol("Sk.abstr.iter", Sk.abstr.iter);
```

albertjan:
Other that that bit this looks solid and can be pulled if you ask me. @skulpt/committers who's up next :smile:

rixner:
I don't think you can monkey patch it that way.  You will then not be able to have nested loops that iterate over the same object.  I will move the constructor into the local scope of ```Sk.abstr.iter```, though.

PR #478, opened 2015-06-05
A Proposal for a New  Skulpt API

bnmnetp:
Far from complete, I wanted to elevate this for discussion and improvement.


API.rst
```diff
@@ -0,0 +1,161 @@
...
+-------------------
+
+* skulpt.exec(code) -- handles promises, exceptions, calls skulpt.error when an error occurs...
+* skulpt.eval(code)  -- returns a javascript object resulting from evaling a Python expression
+* skulpt.input  --  function -- called when ``input`` is called a default with a dialog provided
+* skulpt.import -- function called for importing files the current builtinRead
+* skulpt.output = function   -- A default would be provided that would append text to a configured html element
+* skulpt.error = function  -- A default would be provided that would append text to a configured html element
+* skulpt.config()  to configure things such as
+* skulpt.global(name) -- returns a javascript object mapping names to javscript values based on the last exec or if name the value for that name
```
LeszekSwirski: > returns a javascript object mapping names to javscript values based on the last exec or if name the value for that name

Tricky, does it deep-copy values in global scope to map them to javascript values?

API.rst
```diff
@@ -0,0 +1,161 @@
...
+----------------------------
+
+* A smallish but more nicely and consistently named set of functions could wrap the current stuff and make life easier for people trying to get started.
+* The old functions would remain so as to provide backward compatibility, in fact I anticipate that this new api would largely be a thin wrapper over the old.  So keeping the old is important for backward compatibiliy and for advanced users who may want more fine grained control.
+
+Objects and Methods
+-------------------
+
+* skulpt.exec(code) -- handles promises, exceptions, calls skulpt.error when an error occurs...
+* skulpt.eval(code)  -- returns a javascript object resulting from evaling a Python expression
```
LeszekSwirski: Is this expression in a language sense, or expression in the sense of "a string of code?". If the former, it might be a bit restrictive (e.g. no blocks, no function definitions), plus we'd need to modify Compiler to also support compiling expressions. If the latter, then I'm not sure what value we'd return.

API.rst
```diff
@@ -0,0 +1,161 @@
...
+* skulpt.input  --  function -- called when ``input`` is called a default with a dialog provided
+* skulpt.import -- function called for importing files the current builtinRead
+* skulpt.output = function   -- A default would be provided that would append text to a configured html element
+* skulpt.error = function  -- A default would be provided that would append text to a configured html element
```
LeszekSwirski: The four above sounds like they either belong in `config()`, or should be styled more like event handlers (e.g. 'onInput')

API.rst
bnmnetp: I guess I had envisioned a true expression evaluator (knowing that we cannot do it right now) returning a remaped result.

API.rst
bnmnetp: Excellent point.  I think I like the onInput idea, but I could be convinced either way.

eah13:
A tiny plug for [semantic versioning](http://semver.org/) and actual numbered releases, esp in connection with API changes.  This wouldn't need to change the flow of development for the most part- we're already working on feature branches for the big stuff (which would equate to .X increments, roughly).

I'm all for the changes (they look very reasonable)- we'll just need a way of communicating them and allowing devs to use a specific version of the project in the case of inevitable (if unintended) breaking changes.

On that note, it seems that turning the old API into a wrapper around the new would make more sense.  Unless the new API is just intended as sugar, why not make the name indicate where the code can be found?

bnmnetp:
@eah13 I did not understand your last comment.  You are suggesting exactly the opposite of what the new API is supposed to be.  `skulpt.exec` wraps the current `Sk.importMainWithBody` and other stuff.

albertjan:
@eah13 since we're pre 1.0 we can do what ever we want to the api :smile:

bnmnetp:
So, @eah13 I thought we had adopted SemVer when we started posting releases on bower.  But I guess I haven't followed the guidelines about 0.X.X like I should.  We would be at 0.29.x by this time.

Are you suggesting that we  should do a release with every PR accepted?  I have typically just waited until it felt like a new release was in order.  Not a very scientific approach I will grant.  The other downside to doing a release with every PR is that every now and then something bad slips through.  But maybe that will be reduced now with an expanded group of reviewers and approvers :smile:

eah13:
> Are you suggesting that we should do a release with every PR accepted?

No, definitely not.  Small PRs would become a bullet in the release notes whenever a 0.0.X is created.  I'd like to say that I think you've been doing a great job so far- none of this is intended as criticism, just more of the good stuff.

Seems like 0.0.X remains a whenever seems right kind of thing, 0.X is, roughly, a feature branch, and X. would be any breaking changes to the API.  So the API changes you're proposing wouldn't neccesarily be a 1.0 but I think they're a candidate for a long term API, so once they're mature maybe they do become 1.0 and directly accessing the internals is depricated/not supported.

@albertjan I think the power has gone to your head :crown: :)

> You are suggesting exactly the opposite of what the new API is supposed to be. skulpt.exec wraps the current Sk.importMainWithBody and other stuff.

`Sk.importMainWithBody` has never made much sense to me as a name. I suppose what I meant is that if there's an opportunity to make the internals match the new API more closely let's take it. Otherwise seems like onboarding develoepers into the internals becomes that much more confusing as time goes on because of the layers.

API.rst
```diff
@@ -0,0 +1,103 @@
...
+
+The new API could look like this
+
+.. code-block:: javascript
+
+   <script type="text/javascript">
+   function runit() {
+      var prog = document.getElementById("yourcode").value;
+      var mypre = document.getElementById("output");
+      skulpt.configure({outelem: mypre, turtleCanvas: 'mycanvas'})
```
albertjan: I don't know if I like the `outelem`. It binds skulpt to the dom rather strictly which isn't necessary at all. I really like having an `outf`. We could easily add both offcourse :smile:

rixner: I agree.  I do not think that the Skulpt API should be tied to the DOM.  For instance, I could not use this proposed API in CodeSkulptor, as it doesn't work this way.

I like the idea of having a function as the interface for output (so one can post-process/format/whatever the output in any way they want).  For the program, I think Skulpt should either get it directly as a string or it should take a function that returns a code string when Skulpt calls it so that code can come from arbitrary elements (and be pre-processed if necessary).  We could provide default functions that tie directly to particularly named basic DOM elements as an easy start.  

I don't use turtle, so I don't have a similarly strong opinion, but I do feel somewhat similarly about decoupling.  I guess it's inherently tied to a canvas, though, so it might not be worth it to try to isolate it from the DOM.  In that case, though, it seems like you should pass it an actual canvas to use, not a string.

bnmnetp: So, my intention was to have ``skulpt.output`` be a function same as today, which could do whatever.  BUT, we would provide a default function that used the ``outelem``.  The function specified by ``skulpt.output`` could, of course, ignore the ``outelem`` property entirely. 

I was trying to provide a really simple entry point for people, while preserving the current ability to have an arbitrary function.

More complex than it needs to be??

PR #479, opened 2015-06-05
An outline and some other random stuff for a reference manaul

bnmnetp:
The start of an outline for a real documentation project / reference manual for skulpt.  It is interesting how thinking about documenting and explaining makes you question a number of things or gives me ideas for naming conventions that might make the code MUCH more understandable.

For example as I look at all of the ``Sk.abstr.foo`` functions I wonder how would a dunder method ever get called when using some of these?  Are they all broken in that sense?

Anyway I started this outline, and then started putting a few things in places as they were top of mind.


bnmnetp:
It would be interesting to have a section that talked about some of the important conceptual ideas that you have to learn and get over.  What were those hurdles for you?

bnmnetp:
Despite the outline I created, it finally occurs to me that there must be some kind of documentation tool we could use to generate documentation for some or all of the skulpt implementation.  At a minimum things like Sk.misceval, Sk.abstr would be very useful, and some documentation already exists.

After a couple minutes of google searching I find jsdoc  which understands the format we are already using for the closure compiler.

``npm install jsdoc``

So, it seems that our documentation efforts could be helped a lot by documenting the source as we go, and then turning the hacking.rst file and some other stuff into more of a cookbook/conceptual document.


What do you think?  @albertjan Have you used jsdoc?

bnmnetp:
You will need to add ``npm install -g jsdoc`` to the list of utilities we now use.

1.  The HACKING.md file forms the conceptual, overview documentation
2.  jsdoc generates docs for   abstract.js, misceval.js, function.js, and ffi.js at the moment.   These are still very incomplete, but you get the idea of what everything will look like.
3.  skulpt.py doc  will now run ``jsdoc -c jsdoc.json HACKING.md`` this will generate a bunch of HTML files in ``doc/ProgMan``

With a bit of distributed effort on the comments in the key js files I think we will have something really nice.

bnmnetp:
I think this could be merged any time.  Further work on adding documentation comments to existing functions can be done on their own PR's.  Also additional work on restructuring and polishing the HACKING.md document can also be done independently.

albertjan:
This looks good to me, jsdoc is a great choice if you ask me, esp since it uses the closure comments.

bnmnetp:
Yes, and it has a side benefit of enforcing additional type annotations on functions that are not yet, but should be documented.  I think we could have a very nice doc set, in a short time if we focused on import.js, abstr.js function.js, misceval.js and import.js.

Next priority would be the builtin functions in builtin.js and then the standard types:  list, tuple, dict, set  I'm not certain exactly how much or little should be exposed as part of the API documentation, but it seems like at least the constructors.  

A cookbook section on writing a generator would be pretty useful too.

Of course, the hidden side to this is that it exposes how inconsistently implemented Skulpt is.  There are all kinds of different ways used to do this and that (iteration for example) depending on who added the feature.

PR #481, opened 2015-06-09
Inheritance and Subclassing

mchat:
This PR is a work in progress towards the ultimate goal of having subclassable builtins. Here's a first draft of the work that has been done so far, and what's ahead.

- [x] Use Javascript inheritance to enforce Python relationships between builtins
  - [x] All builtins inherit from Object
  - [x] Create abstract numtype and seqtype classes from which numeric and sequence types inherit, respectively
  - [x] Make bool a subclass of int
  - [x] Separate number.js into int_.js and float_.js for clarity and accuracy when subclassing

- [ ] Use inherited dunder methods to implement builtin functions
  - [x] As in CPython, enforce coercion among number types by returning NotImplemented from the dunder method of the "lesser" type
  - [x] Use dunder methods in Sk.misceval.richCompareBool
  - [x] Use dunder methods in Sk.abstr.binary_op and Sk.abstr.binary_iop
  - [x] abs()
  - [ ] divmod()
  - [ ] hash()
  - [ ] len()
  - [x] pow()
  - [ ] repr()
  - [ ] str()
  - [ ] unicode()

- [ ] Allow user-defined classes to subclass builtins

There's still quite a lot to be done, clearly! Unfortunately the last item on this list is a bit vague, as I'm still working out the details of my approach.

As an example of what the existing changes already provide, try the following code:

```
i = 25
print i.__str__()       # Inherited from Sk.builtin.object

f = 2.5

print i / f             # Implemented by Sk.abstr.binary_op_, which calls first i.__div__(f) then f.__rdiv__(i)

print i.__div__(f)      # NotImplemented, as ints do not know how to handle floats
print f.__rdiv__(i)     # Reflected div is used instead, as float can handle int
```

There's quite a lot of changes in this PR, and I expect there will only be more as I continue to add to it. However, I anticipate that enforcing inheritance in this method will help contribute to a cleaner codebase, with clearer encapsulation and a (slightly) more consistent architecture.


bnmnetp:
Hi @mchat I haven't looked at your code yet but two questions:

1.  This looks to be a massive PR, is there a way to break it up into smaller chunks?
2.  I had to comment out two tests from CPython today for `__delitem__` and `__setitem__` on lists, that may be candidates for fixing in your PR.

Thanks for your work, this will be amazing when its done.

LeszekSwirski:
Does this allow both old- and new-style classes?

waywaaard:
The last one is actually not that difficult, we just need to discuss how we do it. We cannot subclass builtins as they do not have the ```__mro__```, ```__dict__```, ```__bases__``` and our internal ```$d```. ```object``` is subclassable because those properties are initialized in ```import.js``` in ```doOneTimeInit...```. We have to figure out how we set up the mro (do we just set the bases and use one of the type methods for copying?) or do it manually. However, we need to map the internal methods ```tp$```, ```mp$``` and so on to real slot names.

albertjan:
@mchat wow! awesome work!

mchat:
@bnmnetp -- I can certainly split this PR into several smaller ones. Maybe one for enforcing Javascript inheritance/splitting number.js and another for using dunder methods in builtin functions?

@LeszekSwirski  -- Right now, I believe this only supports new style classes. I've been largely following Python 3 implementation details where it differs from Python 2, and I believe Python 3 removed the old style classes.

@waywaaard  -- That's good to know! I'll be sure to look into that later today.

@albertjan  -- Thanks!

PR #482, opened 2015-06-10
add brun command to run a file in the browser

bnmnetp:
I suspect this is a workflow that many of us follow but is usually a pain.  So often I put up with the crappy command line debugger and waste time because its too much of a pain to make yet another test webpage.  So I fixed that.   Introducing `./skulpt.py brun myfile.py`

Did a wee bit of refactoring in skulpt.py to make the `brun` command.  It works just like `run` except it generates a simple webpage so you can run in the browser nearly as fast as running on the command line.  But without all the hassle and manual copying of files.

1. build an uncompressed skulpt.js  (no waiting for dist to complete)
2. build skulpt-stdlib.js
3. grab the file from the command line and throw it into a webpage template
4. open said page in browser (I don't know how to do this in Windows @albertjan)

Suggestions for improvements welcome.

### DIFF
--- a/skulpt.py
+++ b/skulpt.py
@@ -543,6 +543,7 @@ def run_in_browser(fn, options):
     make_skulpt_js(options,RUN_DIR)
     docbi(options,RUN_DIR)
     shutil.copy("support/jsbeautify/beautify.js","{0}/beautify.js".format(RUN_DIR))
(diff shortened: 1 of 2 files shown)


albertjan:
On windows it's `start` instead of `open`. :smile: I think it's nicer to use the files from the `src` directory instead of the concatenated file. If you run into an error then you get a filename and a line number that corresponds to your editor :smile:

albertjan:
Oh thank you for this though!

PR #484, opened 2015-06-10
Split Sk.builtin.nmber into Sk.builtin.int_ and Sk.builtin.float_

mchat:
As per @bnmnetp's request, here is the first of several smaller pull requests.

This PR splits `Sk.builtin.nmber` into `Sk.builtin.int_` and `Sk.builtin.float_`. It builds on my initial work from #481 but works with the current state of Skulpt (aka no Javascript inheritance in the builtins). I also did some extra work to track down all the references to `Sk.builtin.nmber` and replace them with the appropriate class.

```javascript
new Sk.builtin.nmber(a, Sk.builtin.nmber.int$)      -->     new Sk.builtin.int_(a)
new Sk.builtin.nmber(a, Sk.builtin.nmber.float$)    -->     new Sk.builtin.float_(a)
new Sk.builtin.nmber(a, undefined)                  -->     Sk.builtin.assk$(a)
```

I modified `Sk.builtin.assk$` to return a new instance of the appropriate class based on a given Javascript value. This can be helpful If the type of number is not known at compile time, as previously passing `undefined` as the type to `Sk.builtin.nmber` accomplished the same thing.

Currently, `Sk.builtin.nmber` still exists largely unmodified, except that the constructor returns an instance of either `Sk.builtin.float_` or `Sk.builtin.int_`. If a third party module calls the constructor, they shouldn't have any problems. However, if they attempt to use any of the methods directly (i.e. `Sk.builtin.nmber.prototype.nb$add(x)`) they may experience problems. 

If this PR is accepted, we'll need to address those inconsistencies. To my mind, there are two real options:
- Make `Sk.builtin.nmber` methods call the corresponding methods in `Sk.builtin.int_` or `Sk.builtin.float_`
- Throw a deprecated error from all `Sk.builtin.nmber` methods indicating that maintainers should switch to the new number types

Personally, the former seems a bit close to silent failure and could also introduce new bugs. However, it might be a gentler transition. What do y'all think? Am I missing another option?


bnmnetp:
Thanks @mchat !

I like the idea of throwing deprecated from Sk.builtin.number.  There are some, but not a huge amount of third party modules out in the wild.  There are a couple that come to mind.  @waywaaard's work on numpy and @bzwheeler's work with graphing in pygal.  I would love to hear from both of them on this, and whether they see any particular challenges.

Secondly.  In another thread we are moving towards getting the Skulpt API documented using jsdoc strings, and I have stated that new PR's must provide jsdoc strings for the parts of the API that other programmers will use.  So, please do add the appropriate documentation for `int_` and `float_`

Thanks again for your work.

Brad

waywaaard:
I see no problem here, I've counted 6 places which would need to change.

mchat:
@bnmnetp I've got most of src/int.js documented now. I've been having a bit of trouble with Sk.builtin.func declarations.

```
Sk.builtin.int_.prototype.__complex__ = new Sk.builtin.func(function(self) {
    return Sk.builtin.NotImplemented.NotImplemented$;
});
```

Closure compiler seems to be confused, as it considers the new Sk.builtin.func instance a type, but the inner Javascript function as a function to document. I can separate the two and document each separately:

```
/**
 * Implements \_\_complex\_\_ dunder method.
 * @param  {Sk.builtin.int_} self This instance
 * @return {Sk.builtin.NotImplemented} This method is currently not implemented.
 */
Sk.builtin.int_.prototype.tp$complex = function(self) {
    return Sk.builtin.NotImplemented.NotImplemented$;
};

/**
 * Python function which returns this instance as a Python complex. 
 * Implemented by Sk.builtin.int_.tp$complex.
 * @type {Sk.builtin.func}
 */
Sk.builtin.int_.prototype.__complex__ = new Sk.builtin.func(Sk.builtin.int_.prototype.tp$complex);
```

This compiles correctly, but does clutter the codebase and the documentation. I haven't been able to find an example like this one that is already documented. Do you have any recommendations for how to document functions such as this?

bnmnetp:
Hmmmmm, 

It seems jsdoc is seeing it as an instance of func.  Maybe look at the @instance tag or the @global tag as a better workaround??

mchat:
int_ and float_ are both documented, as is Sk.builtin.assk$ as I changed that function significantly. 

Further, all Sk.builtin.nmber methods now throw an ExternalError with a deprecation message.

rixner:
This looks solid to me.  I do have a question, though:  what is the point of ```Sk.builtin.assk$``` at this point?  Are there still places where it is not clear whether an int or a float should be created?

mchat:
`Sk.builtin.assk$

mchat:
I am ready to begin work on the next PR in the sequence related to #481, which would implement builtin functions (such as `abs` and `pow`) with the appropriate dunder methods. However, all that work would be dependent upon this PR and #485, neither of which has been accepted yet. I would prefer not to begin the next section of work until those two PRs are under review, or preferably accepted.

If we're not comfortable merging this PR or #485 into master, we could start a feature branch and I can submit my PRs against that.

As a side note, I am working under a bit of a time crunch. I graduated in May and will begin my full time job in July, so I am only able to contribute to Skulpt full time throughout the rest of June.

albertjan:
I think we should work towards getting this into master asap. So everyone can break their stuff and file bugs. I'm going to review #477 now. So when we get that in I'll do this one after that one. Maybe @rixner can do a final once over after that. So you can get started on the second PR.

albertjan:
Congratulations on the job! :smile: where are you going to work?

mchat:
Thank you @albertjan! I'm going to work for Square in San Francisco.

src/builtin.js
```diff
@@ -118,7 +133,10 @@ Sk.builtin.asnum$nofloat = function (a) {
     if (typeof a === "number") {
         a = a.toString();
     }
-    if (a.constructor === Sk.builtin.nmber) {
+    if (a.constructor === Sk.builtin.int_) {
+        a = a.v.toString();
+    }
+    if (a.constructor === Sk.builtin.float_) {
```
albertjan: nitpcik: can be written as `a.constructor === Sk.builtin.int_ || a.constructor === Sk.builtin.float_`

src/complex.js
```diff
@@ -1060,3 +1060,118 @@ Sk.builtin.complex.prototype.__nonzero__ = new Sk.builtin.func(function (self){
...
+ *
+ * flags can be zero or more of the values Py_DTSF_SIGN, Py_DTSF_ADD_DOT_0, or Py_DTSF_ALT, or-ed together:
+ *
+ * Py_DTSF_SIGN means to always precede the returned string with a sign character, even if val is non-negative.
+ * Py_DTSF_ADD_DOT_0 means to ensure that the returned string will not look like an integer.
+ * Py_DTSF_ALT means to apply “alternate” formatting rules. See the documentation for the PyOS_snprintf() '#' specifier for details.
+ * If ptype is non-NULL, then the value it points to will be set to one of Py_DTST_FINITE, Py_DTST_INFINITE, or Py_DTST_NAN, signifying that val is a finite number, an
+ * infinite number, or not a number, respectively.
+ */
+Sk.builtin.complex.PyOS_double_to_string = function(val, format_code, precision, flags, type) {
```
albertjan: This function is called `str$` in int and float isn't it? It looks like this is a port from cpython, atleast the name suggests it. shouldn't we rename this function to str$ too and add it to the `prototype`? Or am I mistake, if so ignore this comment :smile:.

waywaaard: I've added this for my complex type implementation. ```str$``` may call this but cannot be renamed as it requires a few params to work.

albertjan: Allright clear :smile:

albertjan:
This makes numbers in skulpt look a million times saner! No more weird constructions with the types. I fiddled with it and everything seems fine. I added a few comments, nothing ground breaking. I think this is ready to be merged! 

I fiddled with the numbers a while ago (in an ill fated attempt to make the max int 32bit) and there are so many tests for the numbers I think most of this code is pretty well covered by them.

@skulpt/committers who's gonna pull the trigger?

albertjan:
@mchat sorry this hasn't been pulled in yet, I would do it myself, but the scale of the PR suggests it's a probably a good idea to have another set of eyes on it before we do.

bnmnetp:
Sounds like it is up to you @albertjan. From Santorini. :smile:

albertjan:
@bnmnetp ohhh man Greece on its best! Be sure to drink the wine there! :smile: I did review this one and I'm ok with this being merged but we said two committers should look at a large PR like this before merging. :smile: And I don't want to deviate from that plan already. It's still early in the US they may get around to it.

rixner:
@albertjan, I already looked at it. See my comment above that it looks solid.  I can go over it again tonight if you want, but I think it's safe for you to accept it.

albertjan:
@rixner ahh ok I missed that one :smile: merging

mchat:
Awesome! I will work on rebasing #485 right away.

mchat:
I did not rebase this branch after #477 was merged, and it looks like some of git's automatic merges are causing the build to fail. Let me see if I can track down the bugs and fix that before I work on #485. Sorry about that!

albertjan:
Don't worry about it, git usually gets things right but it isn't fail safe. Thanks for looking into it!

PR #485, opened 2015-06-12
Introduce Javascript inheritance for builtins

mchat:
Here is the second PR related to #481.

This pull request makes every builtin object inherit from `Sk.builtin.object`. All methods attached to `Sk.builtin.object.prototype` (and their documentation) are thus inherited by every instance of every builtin object.

I also added two abstract classes: `Sk.builtin.numtype` and `Sk.builtin.seqtype`. All numeric types (`Sk.builtin.int_`, `Sk.builtin.float_`, `Sk.builtin.lng`, `Sk.builtin.complex`, and `Sk.builtin.bool`) inherit from the former. All sequence types (`Sk.builtin.str`, `Sk.builtin.list`, and `Sk.builtin.tuple`) inherit from the latter. Note that `Sk.builtin.set` and `Sk.builtin.dict` are considered container types, but not sequence types, and so they do not inherit from `Sk.builtin.seqtype`.

`Sk.builtin.bool` now is a proper sublcass of `Sk.builtin.int_`.

NOTE: This PR depends on #484, as setting up the inheritance tree only works if ints and floats are in separate classes. Currently, all of the commits for that PR are also included in this one (the branch for this PR is based off the branch for that PR in my fork). However, if #484 is accepted, I will rebase this one. For clarity, the commits starting on June 12 (beginning with 5d2c2fa) belong to this PR exclusively; those on June 10-11 belong to #484.


albertjan:
This makes numbers in skulpt look a million times saner! No more weird constructions with the types. I fiddled with it and everything seems fine. I added a few comments, nothing ground breaking. I think this is ready to be merged! 

I fiddled with the numbers a while ago (in an ill fated attempt to make the max int 32bit) and there are so many tests for the numbers I think most of this code is pretty well covered by them.

@skulpt/committers who's gonna pull the trigger?

mchat:
This is a bit confusing because all the commits from #484 are also in this pull request (#485) because this PR's branch in my fork is based off the branch for the other. So #484 should probably be accepted first.

I've been running some performance tests on this pull request (#485) and it looks like my work on introducing Javascript inheritance has a significantly poor impact on performance. I'm working to determine the exact cause of the slowdown and  modify the implementation. In the meantime, this PR is not ready to be merged. #484, as far as my tests indicate, does not impact performance.

albertjan:
Oh darn the comment should have gone with the other PR. I got tangled up in the PR's sorry.

albertjan:
Due to the merge of #477 this needs to be rebased or merged with the master. I keep saying sorry @mchat :smile:

mchat:
This PR is now rebased off master. I still have concerns about the performance impact of introducing Javascript inheritance. I'm going to continue looking into that today.

mchat:
I finally determined the remaining performance bottleneck for this PR. I was attempting to set up the internal `$d` object for every builtin instance as it was instantiated -- about three lines of code. Whether as an external call or inlined, that caused about a 30% increase in the runtime of the test suite. I believe part of this is due to the number of builtins that are created as part of internal Skulpt functions -- for example, both `Sk.builtin.GenericGetAttr` and `Sk.builtin.type.typeLookup` create at least one new `Sk.builtin.str` every call, and these are frequently called functions.

In order to prevent the decrease in performance, I decided to wait to instantiate the `$d` object until it was needed, when either `Sk.builtin.GenericGetAttr` or `Sk.builtin.GenericSetAttr` are called. This seems very hacky to me, but it does work. I hope that when I begin working on subclassing builtins, I will be able to design a more elegant solution but that seems like work for another PR.

LeszekSwirski:
Would calling the object constructor be a more appropriate way of initialising things like `$d`?

mchat:
That was my initial plan. The original version had every constructor calling its super constructor, which slowed down performance by 50%. I then tried calling object's constructor directly from every other builtin's constructor, which slowed down performance by 30% -- less, but still significant.

LeszekSwirski:
Strange that a single field assignment should be so costly, considering how much other crap those constructors are doing. I wonder if it's triggering a deoptimisation.

albertjan:
@mchat is this PR ready for review? Or do you still want to tinker with it for a while :smile:

mchat:
Yes, I think it's ready for review. I've been working on subclassing builtins, and part of that work is to initialize the `$d` object for builtins in `Sk.doOneTimeInitialization`, as @waywaaard suggested. So that will resolve the weirdness with the `$d` object in this PR. 

I'll make the PR for subclassing builtins within the next day or so (I want to test multiple inheritance first, but single inheritance is currently working!). That PR depends on this one being accepted.

src/int.js
```diff
@@ -121,18 +128,22 @@ Sk.builtin.int_ = function (x, base) {
     return this;
 };
 
+Sk.abstr.setUpInheritance("int", Sk.builtin.int_, Sk.builtin.numtype);
+
 Sk.builtin.int_.co_varnames = [ "base" ];
 Sk.builtin.int_.co_numargs = 2;
-Sk.builtin.int_.$defaults = [ new Sk.builtin.int_(10) ];
```
albertjan: If you've removed this, how do you setup the default value of the kwarg?

mchat: I moved this value to [constants.js](https://github.com/mchat/skulpt/blob/subclass/src/constants.js#L21). All the builtin constants are initialized after all the builtin classes have been defined.

albertjan: Ah ok, you don't want to keep it close to the method? There are some more here: https://github.com/skulpt/skulpt/blob/master/src/sorted.js#L56

albertjan:
Looks great! The only thing I'm missing is some tests. So we don't accidentally break any of the javascript inheritance prototype magic :smile: in the future.

src/int.js
mchat: Good to know! I'll move those too. It's a bit easier to initialize all the constants in one place in case there are any internal dependencies in their constructors.

mchat:
I'll work on creating a new unit test right away!

src/int.js
albertjan: Ah yes ofcourse you're right

mchat:
I just rebased this PR off master, so it should be ready to merge when it's done being reviewed.

albertjan:
Alright everyone, I think this is ready to be pulled in. @bnmnetp did a test against his books today. If none of you disagree, or if one of you wants to merge push the button. Otherwise I'll do it tomorrow.

src/object.js
```diff
@@ -102,27 +132,302 @@ Sk.builtin.object.prototype.tp$setattr = Sk.builtin.object.prototype.GenericSetA
...
+ * @name  __ge__
+ * @memberOf Sk.builtin.object.prototype
+ * @instance
+ */
+Sk.builtin.object.prototype["__ge__"] = function (self, other) {
+    Sk.builtin.pyCheckArgs("__ge__", arguments, 1, 1, false, true);
+
+    return self.ob$ge(other);
+};
+
```
bnmnetp: Hi @mchat 

So I was looking at another problem today and started experimenting a bit.  And I think this is not quite right.  Here's the example:

```
class Foo1(object):
    def __add__(self, other):
        if isinstance(other, Foo2):
            print "OK"

class Foo2(object):
    def __init__(self):
        pass

x = Foo1()
y = Foo2()

# this goes OK
z = x + y

print x.__ge__(y)
```
Now this fails with a wrong argument count.

I tracked it through to line 225 in abstract.js (which has been the same since 2010 I would add) where it says
``vop.call(v,w)``  The problem of course is that call treats ``v`` as ``this`` and then w is undefined, hence the wrong argument count.

But I think the bigger issue is that in iboNameToSlotFunc_ there was an assumption that ``obj['__xxx__')`` would be a user created python function not an internal javascript function.

What do you think the right way to go is?

bnmnetp: So, after watching some soccer, It came to me that one solution is to modify the dunder functions in object and remove the self parameter, and use ``this`` as intended.

mchat: Those dunder methods in object are actually wrapped as `Sk.builtin.func` in [`Sk.doOneTimeInitialization`](https://github.com/skulpt/skulpt/blob/master/src/import.js#L220), as object is declared before function. I did add a comment in there mentioning this but it's hard to find and should be made more visible/clear -- see https://github.com/skulpt/skulpt/blob/master/src/object.js#L376.

mchat: When I run your code example, I get
```python
OK
NotImplemented
```
which is the same output as python 3 (python 2 doesn't support the `__ge__` dunder on objects).

PR #487, opened 2015-06-17
Fix bugs introduced by merging #484

mchat:
Automatic merge of #484 broke the build as it was not properly rebased off master. This PR fixes those bugs.


albertjan:
LGTM thanks for fixing it :smile:

Issue #425, opened 2015-05-22
Bitwise operators


eah13:
I have a spidey sense that this may be tough to implement but no real knowledge to back that up.

Reports from a user the bit shifting isn't working. Actually, now that she mentions it I recall a student in my UNC class discovering this as well.  https://trinket.io/python/60df47d267 This example is in the context of a fancy fizz buzz solution.  

I imagine most if not all other bitwise operators will need work as well.

I could see wider applicability for trying to bridge intermediate student into an understanding of binary data.

indrora:
woohoo! My bug!

Further minimal example:

```
print True << 1, False << 1, False -1 
```

Expected results: 2 0 -1

Skulpt/Trinket results: 2 1 1

waywaaard:
I had a small look at ```False - 1```. Bool does not inherit from number and therefore, does not have any of the internal operator methods. In ```Sk.abstr.binary_op_``` the LHS-check fails for above mentioned reasons. Then it tries the RHS and switches the arguments and this produces the wrong result as it calculates for this example ```1 - 0```

bnmnetp:
This is fixed by PR #490 currently under review!

PR #490, opened 2015-06-25
Make builtins subclassable

mchat:
This is the last PR related to #481, and makes builtin types subclassable.

Here's a brief overview of what this PR makes possible. See [test_subclass.py](https://github.com/mchat/skulpt/blob/subclass/test/unit/test_subclass.py) for more comprehensive testing.

```python

# Subclass a builtin with no changes
class SimpleInt(int): pass

x = SimpleInt(10)
print x == 10           # True  (int.__eq__)
print x + 10            # 20    (int.__add__)
print x - 10            # 0     (int.__sub__)


# Subclass a builtin and override a magic method
class MyInt(int):
    def __add__(self, other):
        return 5

x = MyInt(10)
print x == 10           # True  (int.__eq__)
print x + 10            # 5     (MyInt.__add__)
print x - 10            # 0     (int.__sub__)


# Subclass a subclassed builtin and override a different magic method
class AnotherInt(MyInt):
    def __sub__(self, other):
        return -5

x = AnotherInt(10)
print x == 10           # True  (int.__eq__)
print x + 10            # 5     (MyInt.__add__)
print x - 10            # -5    (AnotherInt.__sub__)


# Subclass multiple classes which inherit from the same builtin
class MultipleInheritance(MyInt, SimpleInt): pass

x = MultipleInheritance(10)
print x == 10           # True  (int.__eq__)
print x + 10            # 5     (MyInt.__add__)
print x - 10            # 0     (int.__sub__)

# Cannot subclass multiple classes which inherit from different builtins
try:
    class BadMultipleInheritance(int, float): pass
    print "should not see this"
except TypeError:
    print "Caught TypeError"
```

### DIFF
--- a/src/float.js
+++ b/src/float.js
@@ -228,6 +228,11 @@ Sk.builtin.float_.prototype.nb$add = function (other) {
 };
 
 /** @override */
+Sk.builtin.float_.prototype.nb$reflected_add = function (other) {
+    return Sk.builtin.float_.prototype.nb$add.call(this, other);
+};
+
+/** @override */
 Sk.builtin.float_.prototype.nb$subtract = function (other) {
     if (other instanceof Sk.builtin.int_ || other instanceof Sk.builtin.float_) {
         return new Sk.builtin.float_(this.v - other.v);
@@ -239,6 +244,12 @@ Sk.builtin.float_.prototype.nb$subtract = function (other) {
 };
 
 /** @override */
+Sk.builtin.float_.prototype.nb$reflected_subtract = function (other) {
+    var negative_this = this.nb$negative();
+    return Sk.builtin.float_.prototype.nb$add.call(negative_this, other);
+};
+
+/** @override */
 Sk.builtin.float_.prototype.nb$multiply = function (other) {
     if (other instanceof Sk.builtin.int_ || other instanceof Sk.builtin.float_) {
         return new Sk.builtin.float_(this.v * other.v);
@@ -247,6 +258,11 @@ Sk.builtin.float_.prototype.nb$multiply = function (other) {
     }
 
     return Sk.builtin.NotImplemented.NotImplemented$;
+};
+
+/** @override */
+Sk.builtin.float_.prototype.nb$reflected_multiply = function (other) {
+    return Sk.builtin.float_.prototype.nb$multiply.call(this, other);
 };
 
 /** @override */
@@ -584,7 +600,11 @@ Sk.builtin.float_.prototype.nb$inplace_floor_divide = Sk.builtin.float_.prototyp
 /** @override */
 Sk.builtin.float_.prototype.nb$inplace_power = Sk.builtin.float_.prototype.nb$power;
 
-/** @override */
+/**
+ * @override
+ *
+ * @return {Sk.builtin.float_} A copy of this instance with the value negated.
+ */
 Sk.builtin.float_.prototype.nb$negative = function () {
     return new Sk.builtin.float_(-this.v);
 };
--- a/src/float.js
+++ b/src/float.js
@@ -229,6 +229,8 @@ Sk.builtin.float_.prototype.nb$add = function (other) {
 
 /** @override */
 Sk.builtin.float_.prototype.nb$reflected_add = function (other) {
+    // Should not automatically call this.nb$add, as nb$add may have
+    // been overridden by a subclass
     return Sk.builtin.float_.prototype.nb$add.call(this, other);
 };
 
@@ -245,6 +247,8 @@ Sk.builtin.float_.prototype.nb$subtract = function (other) {
 
 /** @override */
 Sk.builtin.float_.prototype.nb$reflected_subtract = function (other) {
+    // Should not automatically call this.nb$add, as nb$add may have
+    // been overridden by a subclass
     var negative_this = this.nb$negative();
     return Sk.builtin.float_.prototype.nb$add.call(negative_this, other);
 };
@@ -262,6 +266,8 @@ Sk.builtin.float_.prototype.nb$multiply = function (other) {
 
 /** @override */
 Sk.builtin.float_.prototype.nb$reflected_multiply = function (other) {
+    // Should not automatically call this.nb$multiply, as nb$multiply may have
+    // been overridden by a subclass
     return Sk.builtin.float_.prototype.nb$multiply.call(this, other);
 };
 
--- a/src/float.js
+++ b/src/float.js
@@ -42,7 +42,7 @@ Sk.builtin.float_ = function (x) {
     }
 
     // Floats are just numbers
-    if (typeof x === "number" || x instanceof Sk.builtin.int_ || x instanceof Sk.builtin.lng) {
+    if (typeof x === "number" || x instanceof Sk.builtin.int_ || x instanceof Sk.builtin.lng || x instanceof Sk.builtin.float_) {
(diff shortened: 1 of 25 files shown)


bnmnetp:
Marie,

Thanks for all of this amazing work! 

I tried the following:

```
class TestFailed(Exception):
    """Test failed."""

raise TestFailed()
```

This fails with the usual:

```
UNCAUGHT EXCEPTION: ExternalError: TypeError: Cannot read property 'mp$subscript' of undefined on line 15
```

How hard would it be to add Exceptions to the list of built-in types that we can subclass?  This would help immensely with the unittest infrastructure.

mchat:
Done! You can now subclass Exceptions and override their magic methods. Also:

```python
# Verify levels of inheritance 
class MyTypeError(TypeError): pass

try:
    raise MyTypeError()
except TypeError:
    print "Recognized MyTypeError as a TypeError"
except:
    print "Did not recognized MyTypeError as a subclass of TypeError"

try:
    raise MyTypeError()
except Exception:
    print "Recognized MyTypeError as an Exception"
except:
    print "Did not recognized MyTypeError as a subclass of Exception"
```

For future reference, all you need to do to make another class subclassable is to call
```python
Sk.abstr.setUpInheritance("childName", child, parent);
```
after the constructor is defined. This also sets up the `tp$name` and `ob$type` of the child, so you shouldn't have to do those separately.

bnmnetp:
Brilliant.   We will definitely want to document that function for future skulpt hackers.

mchat:
`Sk.abstr.setUpInheritance` is now documented, along with a few other functions that are used in inheritance and subclassing.

bnmnetp:
@skulpt/committers  @waywaaard  Would be great to have one more person take a look at this and give it your OK.  Or provide some comments/feedback.

Thanks,

Brad

albertjan:
I've followed it and it looks awesome. And can look at it Monday.

waywaaard:
I have not seen any strange things and could follow your changes. But you have not changed the behavior for ```__rdiv__```, do you? And is #425 resolved with your pull request?

bnmnetp:
I haven't run the whole example but playing around with the repl it appears that this does address #425

waywaaard:
Cool, then I am really happy :)

mchat:
@waywaaard -- Thank you for pointing out that `__rdiv__` wasn't addressed yet! These commits fix overriding ancestor operations, as described in the [Python docs](https://docs.python.org/2/reference/datamodel.html#index-92):

> If the right operand’s type is a subclass of the left operand’s type and that subclass provides the reflected method for the operation, this method will be called before the left operand’s non-reflected method. This behavior allows subclasses to override their ancestors’ operations.

bnmnetp:
OK, I tried the whole fizzbuzz program from #425 and it works just fine.  So with these latest additions to address the reflected methods, this is looking really good to me.

src/float.js
```diff
@@ -250,6 +261,11 @@ Sk.builtin.float_.prototype.nb$multiply = function (other) {
 };
 
 /** @override */
+Sk.builtin.float_.prototype.nb$reflected_multiply = function (other) {
+    return Sk.builtin.float_.prototype.nb$multiply.call(this, other);
```
albertjan: any reason why you don't write `return this.nb$multipy(other)` here?

albertjan:
@mchat great works looks solid! I'm ok for merging, but I'd like you're answer on the `call(this)` stuff :smile:

mchat:
It has to do with subclassing. In Python, reflected magic methods must be manually overridden -- they don't simply use their non-reflected counterparts. If the `nb$reflected...` methods returned `this.nb$...` as you suggest (and as I originally implemented), the following code example would return `"MyFloat.__mul__"` for every call:

```python
class MyFloat(float):
    def __mul__(self, other):
        return "MyFloat.__mul__"

x = MyFloat(2.5)
y = 2.0

print x * y             # MyFloat.__mul__
print y * x             # 5.0

print x.__mul__(y)      # MyFloat.__mul__
print x.__rmul__(y)     # 5.0
```

I can comment as such in each of those reflected methods so that future developers don't try to optimize and break the tests in [test_subclass.py](https://github.com/mchat/skulpt/blob/subclass/test/unit/test_subclass.py).

albertjan:
Really!?!? :smile: wow. What kind of unholy javascript feature makes:

```js
Sk.builtin.float_.prototype.nb$reflected_multiply = function (other) {
    return Sk.builtin.float_.prototype.nb$multiply.call(this, other);
}
```
not equivalent to:

```js
Sk.builtin.float_.prototype.nb$reflected_multiply = function (other) {
    return this.nb$multiply.call(other);
}
```
?

mchat:
Haha, it's not a problem when `this` is an instance of `Sk.builtin.float_` directly. 

In the above example, `MyFloat` is a `klass` instance, and it has a special `nb$multiply` function which calls the user-defined `__mul__` Python function. When `x.nb$reflected_multiply` is called, the `klass` doesn't have `nb$reflected_multiply` in its own prototype. Going up the prototype chain, Javascript finds `Sk.builtin.float_.prototype.nb$reflected_multiply` and calls it with `x` as `this`. Unfortunately, `x.nb$multiply` now calls `MyFloat.__mul__` instead of the original float multiply function.

(See [this funky bit of code](https://github.com/mchat/skulpt/blob/subclass/src/type.js#L5), which helps improve performance tremendously by overriding Skulpt "shortcut" functions in `klass` instances so that you don't have to check for a Python function before calling any "shortcut" functions).

mchat:
Which, thinking about it, none of the Skulpt functions which implement a dunder method should implicitly rely on any other Skulpt function. For example, [`Sk.builtin.int_.nb$divide`](https://github.com/mchat/skulpt/blob/subclass/src/int.js#L301) simply calls `this.nb$floor_divide` and  [`Sk.builtin.int_.nb$reflected_divide`](https://github.com/mchat/skulpt/blob/subclass/src/int.js#L306) simply calls `this.nb$reflected_floor_divide`, which causes the following code to break (it prints `"MyInt.__floordiv__"` for every line):

```python
class MyInt(int):
    def __floordiv__(self, other):
        return "MyInt.__floordiv__"

x = MyInt(5)
y = 3

# x.nb$divide(y)
print x / y                     # 1

# x.nb$divide(y)
print x.__div__(y)              # 1

# x.nb$floor_divide(y)
print x.__floordiv__(y)         # MyInt.__floordiv__

# y.nb$reflected_divide(x) --> y.nb$reflected_floor_divide(x) --> x.nb$divide(y)
print y.__rdiv__(x)             # 1

# y.nb$reflected_floor_divide(x) --> x.nb$divide(y)
print y.__rfloordiv__(x)        # 1

```

This wasn't a problem before because builtins could not be subclassed. I don't have the time to track down, fix, and properly test all the cases like this but I did want to let y'all know this bug will exist once you can subclass builtins.

albertjan:
@bnmnetp I think it's a good idea we add the explanations that @mchat gave to the hacking file. @mchat don't worry about not being able to sift through all the files. @rixner and @trinketapp and @csev and I will find those fast enough and they'll be fixed as we run into them. Or when someone else has the time to check those. 

@bnmnetp I say merge and lets do a new release. With all the new features. I'm reasonably sure we won't have regressions. This is such a big step forward.

bnmnetp:
I agree, this is really great work.  I will accept and make a new release right now.

Issue #465, opened 2015-06-01
Inplace operator does not fall back to normal one


Jacco:
Test case:

``` python
class Foo1:
    def __add__(self, other):
        if isinstance(other, Foo2):
            print "OK"

class Foo2:
    def __init__(self):
        pass

x = Foo1()
y = Foo2()

# this goes OK
z = x + y     

# but this gives error
x += y
```

Maybe this is resolved in @waywaaard s complex PR?

waywaaard:
No, I have not fixed this. But it seems that we are going to rework the dunder method handling.

bnmnetp:
Still not working in 0.10.0 but with @mchat's changes this now seems like it could be implemented by adding nb$add and nb$inplace_add implementations to object.

waywaaard:
Any progress here?

PR #492, opened 2015-07-01
Add default behavior for __iadd__ to call regular __add__

bnmnetp:
This demonstrates an easy approach to fixing #465  -- causes t407.py (which tests `dir` of a classes and subclasses) to fail, but that test is pretty bogus if you ask me as it really does not match real python in any case.

So if we think this is good I can add defaults for other inplace operators.

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -249,6 +249,18 @@ Sk.builtin.object.prototype["__ge__"] = function (self, other) {
     return self.ob$ge(other);
 };
 
+/**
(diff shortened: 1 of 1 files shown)


albertjan:
@meredydd is this fixed by your PR #657

meredydd:
Yes, it is. That PR fixes all ops (not just addition)  in a much more general way. This I think was more of an idea / demonstration from @bnmnetp

meredydd:
Superseded by PR #657. Closing.

PR #494, opened 2015-07-16
Turtle clone works, still has some issues

### LINKED ISSUE #409: turtle.clone() method
Hi, 

it is convenient in game programming -- to make one object, and then clone it... 
Maybe, wouldn't be hard to implement?
https://docs.python.org/3.4/library/turtle.html#turtle.clone

petroleyum:
Looking to resolve #409 I have a working turtle.clone() method, but with one problem that I can see. It loads the turtle module using Sk.misceval.loadname using the Sk.globals object, so if the user writes 'import turtle' in a function then this method will not be able to find it.

Also, I'm not totally sure if the new clone turtle instance has the write Turtle type according to skulpt.

But this is a start

### DIFF
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -1072,6 +1072,70 @@ function generateTurtleModule(_target) {
         };
         proto.$getscreen.isSk = true;
 
+        proto.$clone = function() {
+            console.log(this.skInstance);
+            var turtleMod = Sk.misceval.loadname('turtle',Sk.globals);
+            console.log(turtleMod);
+            var turtleKlass = Sk.abstr.gattr(turtleMod,'Turtle', true);
+            console.log(turtleKlass);
+            var newTurtleInstance = Sk.misceval.callsimOrSuspend(turtleKlass);
+            console.log(newTurtleInstance);
+
+            newTurtleInstance._x = this._x;
+            newTurtleInstance._y = this._y;
+            newTurtleInstance._angle = this._angle;
+            newTurtleInstance._radians = this._radians;
+            newTurtleInstance._shape = this._shape;
+            newTurtleInstance._color = this._color;
+            newTurtleInstance._fill = this._fill;
+            newTurtleInstance._filling = this._filling;
+            newTurtleInstance._size = this._size;
+            newTurtleInstance._computed_speed = this._computed_speed;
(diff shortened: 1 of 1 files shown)


bnmnetp:
I'm not sure why you are using Sk.misceval.loadname.  You should not have to load the turtle module from inside the turtle module!   You can get access to the module object using ``_module``

petroleyum:
Yep, sorry about that. Misceval.loadname did seem a little roundabout, but I still haven't figured out the layout of the turtle module since it was refactored recently so that's the first way I could get it to work.

now the new instance in clone is made with 
`var newTurtleInstance = Sk.misceval.callsimOrSuspend(_module.Turtle);` 
which makes more sense.

bnmnetp:
@bzwheeler do you have a few minutes to take a look at this one?

bzwheeler:
Nice job navigating the new module, it turned out a bit more convoluted than I had initially hoped!

A couple more attributes probably need to get cloned like `_isRadians` and `_fullCircle`. I'll be honest, I don't know if python clones things like the undo buffer size?  If it does, that setting is stored in `_bufferSize`.

Other than that it seems to work nicely.

petroleyum:
I finally got around to this. I added the _isRadians and _fullCircle. I looked around for more properties I was missing but didn't see any.

For `_undoBuffer` and `_bufferSize`, I did some investigating in the python module. A clone will copy the buffer size and undo buffer as it is just a `deepcopy()` operation. The goofy thing is how a cloned turtle will inherit a undo buffer from a parent turtle. In python, a child (cloned) turtle can remove lines that the parent turtle drew. Currently in this skulpt build, the child turtle will follow a parents undo buffer, but it won't remove it's lines. I'll toy with it and try to see if I can recreate that.

I've been doing something like this to test:

```
import turtle
wn = turtle.Screen()

babbage = turtle.Turtle()
babbage.shape("triangle")
babbage.setundobuffer(10)

n = 8
angle = 360/n

for i in range(n):
    charles = babbage.clone()
    charles.undo()
    charles.forward(50)
    charles.undo()
    charles.undo()
    babbage.right(angle)
    babbage.forward(90)
    babbage.stamp()

wn.exitonclick()
```

This will draw an octogon in skulpt, and leave all lines. In python, the undo statements remove all lines.

PR #495, opened 2015-07-21
check for python3 flag to get python3 division

bnmnetp:
In all of @mchat 's great work we lost the check for the python3 flag for the regular / operator.  I realize some may think that was a good thing, but it breaks a bunch of examples in my books.

### DIFF
--- a/src/int.js
+++ b/src/int.js
@@ -304,6 +304,11 @@ Sk.builtin.int_.prototype.nb$reflected_multiply = function (other) {
 
 /** @override */
 Sk.builtin.int_.prototype.nb$divide = function (other) {
+    var thisAsFloat;
+    if (Sk.python3) {
+        thisAsFloat = new Sk.builtin.float_(this.v);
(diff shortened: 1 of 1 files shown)


waywaaard:
The request seems okay for me.

rixner:
Is it possible to have some tests run with the python3 flag set to test these sorts of things so that we don't have regressions again in the future?

bnmnetp:
I'll look into it.  The trick is to be able to set the python3 flag for a particular test.  Maybe I could add a function to sys like we did to set the execution time limit.    I wouldn't really recommend that for casual use, but for testing something like that might work.

bnmnetp:
On second thought, this approach would be too late to catch things during the parse or compile phase...  Maybe a a command line option passed to the test runner similar to debugon.

rixner:
This is pretty straightforward, so I'm going to accept it, given that adding tests is going to take more thought.  I'll open an issue on that topic.

PR #502, opened 2015-08-20
Make $ret, currLineNo, and currColNo local variables.

rixner:
These were global variables.  At a minimum, this causes problems when there is an exception thrown by an imported module. The traceback will be corrupted, as the line and column numbers will all be the same, since all modules were sharing the same variables.


bnmnetp:
This looks pretty easy to accept for me.  Any objections?

waywaaard:
Seems good to me

PR #505, opened 2015-09-09
Support suspending in generator expressions

meredydd:
What it says on the tin. Turns out I missed a call to `Sk.abstr.iternext()` when I first suspension-ised generators.


meredydd:
This build is currently failing because of the problem addressed by PR #504. All tests pass for me with that change applied.

albertjan:
kicked the build. after the npm package merge.

LGTM @skulpt/committers

meredydd:
I've just added an almost-trivial commit that labels line numbers so we get a proper backtrace if we suspend or throw from a genexp.

@albertjan if you LGTM this too, am I good to merge, or do we need someone else?

PR #506, opened 2015-09-09
remapToJS() wasnt working for Dicts 

orionz:
Sk.remapToJs(Sk.remapToPy({hello:"world"})) was failing on properties with a $hash but not items.  I added an explicit check and now it works.  Happy to add a test to go with it if you show me where the javascript only tests go.


albertjan:
LGTM. It's becoming more and more obvious we need a test suite that tests the javascript api of skulpt.

PR #507, opened 2015-09-10
Support (and trace) exceptions in class definitions

meredydd:
`compile.js` wasn't emitting exception-handling code in class definitions. This meant that when students caused an exception in a class definition, the trace wasn't showing up on the correct line.

This change makes that happen correctly (as well as supporting try/catch within a class block, which is pretty perverse, but supported by mainline Python so we might as well).


albertjan:
LGTM merge at will :smile:

albertjan:
@bnmnetp will you merge this?

bnmnetp:
Yes

PR #515, opened 2015-10-03
Minor Turtle changes

ahuhn:
I'm running multiple instances of skulpt on a page, so I added code to let me stop a turtle animation when the next one starts.  (Feature addition)

I also added e.preventDefault in a few places to help out the onkey handling in turtle code.  (Bug fix)

This is my first pull request, so please feel free to give me feedback.

Thanks!

### DIFF
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -1420,6 +1420,7 @@ function generateTurtleModule(_target) {
                         // trigger the intial keydown handler
                         self._keyListeners[key]();
                         self._createKeyRepeater(key, code);
+                        e.preventDefault();
                         break;
                     }
                 }
@@ -1436,6 +1437,7 @@ function generateTurtleModule(_target) {
             this._keyUpListener = function(e) {
                 var interval = self._keyLogger[e.charCode || e.keyCode];
                 if (interval !== undefined) {
+                    e.preventDefault();
                     window.clearInterval(interval);
                     window.clearTimeout(interval);
                     delete(self._keyLogger[e.charCode || e.keyCode]);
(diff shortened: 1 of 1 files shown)


albertjan:
@ahuhn couldn't you call `e.preventDefault` in the implementation of the your eventhandler? Other implementors might want their click events to bubble up.

ahuhn:
I do have a file for my own modifications, and I could put the e.preventDefault there, but I think this change might be useful for others.  It only applies to keypresses, not clicks, and it only happens when the turtle canvas is in focus and there is a turtle screen keyhandler for that key.  You don't want the page moving around under you.  Here's some sample code, and the issue shows up if you paste it into the top box of http://www.skulpt.org/.  

That said, I see the overall place that you are coming from.  If this change is in place, it can't be worked around from the outside whereas e.preventDefault can be added from the outside.  Should I take it out?

```
import turtle

wn = turtle.Screen()                 # Get a reference to the window
wn.bgcolor("lightgreen")             # Set the background color
tess = turtle.Turtle()               # Create our favorite turtle

print "Click on the canvas then press the up arrow key."

# The next four functions are our "event handlers".
def h1():
    tess.forward(30)

def h2():
    tess.left(45)

def h3():
    tess.right(45)

def h4():
    wn.bye()                        # Close down the turtle window

# These lines "wire up" keypresses to the handlers we've defined.
wn.onkey(h1, "Up")
wn.onkey(h2, "Left")
wn.onkey(h3, "Right")
wn.onkey(h4, "q")

# Now we need to tell the window to start listening for events,
# If any of the keys that we're monitoring is pressed, its
# handler will be called.
wn.listen()
wn.mainloop()
```

albertjan:
Sounds reasonable. You're right that you don't want the page to move around when you use the keys. @bzwheeler do you concur?

bnmnetp:
I certainly concur if this fix prevents the page from scrolling when up and down arrow keys are pressed.

albertjan:
Well lets merge it's easily enough reverted.

Issue #513, opened 2015-10-01
Modulo doesn't cause an exception in the case of division by zero


acbart:
> 5 % 0
    NaN
    > 1 + (5 % 0)
    NaN

In regular python, modulo by zero raises `ZeroDivisionError: integer division or modulo by zero`. However, in Skulpt, it happily computes it to be NaN (an Integer value), which can then be treated like any other integer.

waywaaard:
Correct, there is no check for this case. We should use builtin division (addition) method for the remainder and not javascript modulo.

waywaaard:
This can be closed

PR #516, opened 2015-10-05
Fixes integer modulo for zero divisions

waywaaard:
This PR fixes the #513. Instead of using the javascript calculation I've used "python" methods for calculating. Those ensure the correct handling of the types and throw appropriate errors.

### DIFF
--- a/src/int.js
+++ b/src/int.js
@@ -356,21 +356,14 @@ Sk.builtin.int_.prototype.nb$reflected_floor_divide = function (other) {
 Sk.builtin.int_.prototype.nb$remainder = function (other) {
     var thisAsLong, thisAsFloat;
     var tmp;
+    var divResult;
 
     if (other instanceof Sk.builtin.int_) {
-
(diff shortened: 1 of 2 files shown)


albertjan:
LGTM. Want to add a test?

albertjan:
@skulpt/committers merging this, it's a small change

LeszekSwirski:
Did you measure for any performance regression?

waywaaard:
I've not made any performance tests. Though, the calculation adds more calls, but removes a couple of conditions. I think, it could be a bit smaller. On the other side would a pure JS implementation require some more additional checks.

albertjan:
The previous implementation had a bug, I feel that weighs more then performance.

LeszekSwirski:
Oh yeah, correctness is definitely more important than performance, I was just wondering if there was any point adding a todo to include a pure JS implementation in the future (for performance reasons) or if this one is fine.

PR #517, opened 2015-10-05
Timemodule

albertjan:
As a follow up for #420


albertjan:
@waywaaard you can do PR's to this branch. I'll merge them immediately and we'll do a full review when your done :smile:

meredydd:
Is this now in a state to review and merge? (On the basis that most of a `time` module is considerably better than nothing)

No problem if @waywaaard and @albertjan are actively working on this, but the last activity before mine  appears was in 2015...

meredydd:
Oh, scratch that. I see I have a test failure in my code to fix before we can bring `strftime()`/`strptime()` into this branch. Gimme a sec.

meredydd:
Right. Now I've (AFAIK) fixed the tests in #572, I'd be up for merging that into this branch, then this branch into master. Review welcome!

albertjan:
@meredydd I rebased and when all is well/green. I should say @skulpt/committers merge at will. @Jacco thanks for laying the ground work 👍

waywaaard:
Seems good to me. We can fix minor errors always later on. Having the time module allows use to tackle `datetime`.

albertjan:
I'm going to merge this as we won' t be breaking current behaviour. And @waywaaard wants to tackle `datetime` 😄

rixner:
I now get this failure:

test/unit/test_time.py
Fail: expected Dec 31 1969 19:01:01 to be equal to  got Jan 01 1970 01:01:01 
Tests failed in  TimeTestCase.test_strftime  
Ran 5 tests, passed: 4 failed: 1

In my subsequent pull request, though, things pass in travis.  So, I assume this is something about the setup on my machine?   Does anyone else see this (and/or know what is going on)?

albertjan:
I'll try now. Very strange.

albertjan:
could it be timezone based? let me tell my computer I'm texas 😄

rixner:
That makes sense.  I'm getting 6 hours earlier than the test expects in both CPython and Skulpt.  The test is therefore flawed if it depends on where you are, right?

albertjan:
@rixner #577 should fix this.

meredydd:
Hrm. I thought I'd tested that, given that I'm not on GMT (curse of the
British programmer - local time == UTC half the year, but it's summer atm).
Perhaps it's applying the summer-me correction but not the time-zone
correction?

albertjan:
@meredydd you call `localtime()` which should be the localtime right? I just added the time.timezone offset to the test. You think that the right solution?

rixner:
I'm at UTC-6 and I'm getting 5 hours off.  So, I think it only works in your current timezone.  Where does travis run, though?

#577 trivially fixes it and I think it is the correct solution, so I accepted it.

meredydd:
Ah, yes. On the 1st of January 1970, the UK was not on summer time, but
Texas was not on UTC. Nice catch, and sorry, guys!

PR #518, opened 2015-10-06
Add Sk.misceval.tryCatch()

meredydd:
There is currently no easy way, from Javascript, to catch an exception from a Python function that suspends. `Sk.misceval.tryCatch()` fixes this. It calls a function that may return a suspension, and catches any exceptions thrown directly by that function or when resuming any suspension it returns.

### DIFF
--- a/src/misceval.js
+++ b/src/misceval.js
@@ -997,6 +997,29 @@ Sk.misceval.chain = function (initialValue, chainedFns) {
 };
 goog.exportSymbol("Sk.misceval.chain", Sk.misceval.chain);
 
+
+/**
+ * Catch any exceptions thrown by a function, or by resuming any suspension it
+ * returns.
+ */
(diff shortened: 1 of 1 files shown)


waywaaard:
Seems good to me, maybe we should add an example in the doc string? I konw that we do not have any tests for our internal methods, but a good description should be easy todo.

bnmnetp:
Yes, please add an example to the commented docstring.  We do autogenerate documentation from misceval.js and this is one that I would like to make use of.

After that it looks good to go.

bnmnetp:
@meredydd Now that finals are over I have window of time to clear out this backlog of PR's.  could you add a little docstring with an example of how to use?  Then I'll accept.

bnmnetp:
Why did we merge this without documentation?  What's the point of adding things that nobody knows how to use?

albertjan:
@bnmnetp i added the documentation in #545

PR #520, opened 2015-10-07
Integer division

ahuhn:
This does not pass tests, but I'm opening a pull request to make discussion easier.

### DIFF
--- a/src/int.js
+++ b/src/int.js
@@ -332,12 +332,12 @@ Sk.builtin.int_.prototype.nb$floor_divide = function (other) {
 
     if (other instanceof Sk.builtin.lng) {
         thisAsLong = new Sk.builtin.lng(this.v);
-        return thisAsLong.nb$divide(other);
+        return thisAsLong.nb$floor_divide(other);
     }
 
     if (other instanceof Sk.builtin.float_) {
         thisAsFloat = new Sk.builtin.float_(this.v);
-        return thisAsFloat.nb$divide(other);
+        return thisAsFloat.nb$floor_divide(other);
     }
 
     return Sk.builtin.NotImplemented.NotImplemented$;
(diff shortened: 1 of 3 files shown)


ahuhn:
This bug is more extensive.  I'm looking into it more.

albertjan:
@ahuhn could you add a test for this case too?

ahuhn:
@albertjan Is that a good location for the tests?

waywaaard:
Yes, it is. Currently, we cannot port all of cpython's unit tests due to some of skulpts restrictions.

albertjan:
LGTM @skulpt/committers do you concur?

bnmnetp:
I would just want to make sure this still works correctly when the Sk.python3 flag is true.

albertjan:
Oh good catch never thought of that. @waywaaard what do you think?

ahuhn:
I wasn't sure where to put the tests, but I did test it by hand:

```
print(3/2 == 1.5, "\t3/2 should be 1.5")
print(3//2 == 1, "\t3//2 should be 1")
print(3/2.0 == 1.5, "\t3/2.0 should be 1.5")
print(3//2.0 == 1.0, "\t3//2.0 should be 1.0")
print(-3/2 == -1.5, "\t-3/2 should be -1.5")
print(-3//2 == -2, "\t-3//2 should be -2")
print(-3/2.0 == -1.5, "\t-3/2.0 should be -1.5")
print(-3//2.0 == -2.0, "\t-3//2.0 should be -2.0")
```

Is there a way to change which tests run based on the Sk.python3 flag?

albertjan:
@ahuhn the tests are in the correct place. But there's a flag in skulpt that changes somethings around so it acts a bit more like python3. You can just set it globally just add this to your javascript: `Sk.python3 = true`

ahuhn:
I tested the lines above by changing the Sk.python3 flag in the javascript to test it by hand, but I am not sure where to add the tests in the automated tests.  The file test_int.py should be different for python2 and python3.

albertjan:
There are no automated tests for `Sk. python3`

albertjan:
@bnmnetp I'm satisfied, this doesn't break the python3 int/int division.

PR #521, opened 2015-10-12
Fixes math.log and adds first math unittests

waywaaard:
Currently, the math module is pretty untested and fails for a lot of unit tests. I've implemented `log2` and tried to fix a couple of functions to get the tests running.

`math.pow` needs a lot more time in order to behave like the real one.

### DIFF
--- a/src/float.js
+++ b/src/float.js
@@ -617,6 +617,11 @@ Sk.builtin.float_.prototype.nb$negative = function () {
 
 /** @override */
 Sk.builtin.float_.prototype.nb$positive = function () {
+    // -0 -> 0 (but JavaScript behaves differently)
+    if (Sk.builtin.isNegativeZero(this.v)) {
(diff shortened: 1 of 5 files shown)


waywaaard:
It seems that the one check does fail on travis but not on my local branch. This is weird.

albertjan:
`Fail: expected 1023.0 to be equal to  got 1023.0` :confused:

albertjan:
@waywaaard didn't you want to use the assertAlmostEqual here?

waywaaard:
Yes, at the very first moment. But we should use `frexp` (which is also not implemented yet) for this special case.

albertjan:
@waywaaard can you rebase this PR so it can be merged in?

waywaaard:
I try to do this next week.

meredydd:
Hey, it would be great to get this merged. @waywaaard any chance of a rebase?

PR #522, opened 2015-10-14
Fixes from module import list

waywaaard:
This should fix the missing from list feature for `import  from`. Does not break any other import behavior.


waywaaard:
contains other commits, gonna open a new one

PR #523, opened 2015-10-14
Fixes from importlist without aliases

### LINKED ISSUE #299: package import bugs
Given the directory structure:

```
test/
   __init__.py
   test_support.py
```

The following import statement fails:  ``from test import test_support``

However this one works:  ``from test.test_support import run_unittests``

We need to fix this behavior otherwise we are going to have to modify all of the CPython unit tests to import the ``run_unittests`` function.

waywaaard:
This should fix the missing from list feature for `import  from`. Does not break any other import behavior.

Should fix #299


waywaaard:
Needs some more modification and thinking

PR #524, opened 2015-10-15
Fixes imports that are own files but show up in the from list

### LINKED ISSUE #299: package import bugs

waywaaard:
Should fix #299 .

Example:

```python
from unittest import gui as mystuff

print("nothing to see here")
print(dir(mystuff))
```

### DIFF
--- a/src/import.js
+++ b/src/import.js
@@ -364,6 +364,8 @@ Sk.importModuleInternal_ = function (name, dumpJS, modname, suppliedPyBody, canS
         external = Sk.loadExternalLibrary(name);
         if (external) {
             co = external;
+            filename = Sk.externalLibraries[name].path; // get path from config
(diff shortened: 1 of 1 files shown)


albertjan:
@meredydd what do you think about this. It looks good to me. Pretty important missing feature.

meredydd:
I'm afraid I just timed out on Skulpt stuff for now; I can't review this properly within the next week at least :-/

albertjan:
Alright :smile:

eah13:
Looks good to me

PR #525, opened 2015-10-18
Fixes abs builtin for User Defined Classes

waywaaard:
Fixes `abs` for calling `__abs__` on user defined classes.

```python
class TestAbs:

    def __init__(self):
        self.foo = -3

    def __abs__(self):
        return -self.foo

bar = TestAbs()
print abs(bar)
```


albertjan:
chipping away at it! very nice! LGTM

albertjan:
@bnmnetp I'm pulling this is small fix and @waywaaard know his stuff :smile:

PR #526, opened 2015-10-19
Update packeage.json to be compatible with npm.

Kaworru:
This package.json modification are to used to load the current source code over npm directly from github.
Because of the "private": "true" flag it will not be uploaded the to the npm repository.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,24 @@
 {
+  "name": "skulpt",
+  "version": "1.0.0",
+  "description": "Skulpt is a Javascript implementation of Python 2.x. Python that runs in your browser!",
+  "keywords": ["skulpt"],
+  "scripts": {
+    "lint": "jshint ."
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/skulpt/skulpt.git"
+  },
+  "bugs": {
+    "url": "https://github.com/skulpt/skulpt/issues"
+  },
+  "homepage": "http://www.skulpt.org/",
+  "author": {
+    "name": "Brad Miller"
+  },
+  "license": "MIT",
+  "private": true,
   "dependencies": {
     "jshint": "~2.5.2",
     "jscs": "~1.12",


albertjan:
I don't think you should publish it to npm. If it were to happen we should do it. And I don't believe we actually support node. It may work but It might not continue work in the future. This repository also doesn't include the compiled version of skulpt.

You should probably do this pull request to the skulp/skulpt-dist repository. And then install it using `npm install skulpt/skulpt-dist` installing it from github.

Kaworru:
Sorry if this was unclear, the idea is not to publish it to npm. But it would allow to add the source code as dependency so that it could be automatically fetched (and build) over npm like:
...
 "dependencies": {
    "skulpt": "https://github.com/Kaworru/skulpt/tarball/master",
...

Its also not the question of supporting nodejs by itself, it more that you could fetch the needed source code easily over npm. Which give you the possibility to compile it on your own.

I will create an additional file for the skulpt-dist repro, but it would be great this could be accepted as well.
As I already mention, because of the "private": "true" flag, it is not possible to submit this source code to the npm repro, which is also not the intention of this change.

albertjan:
Ah ok got it :+1: @skulpt/committers sounds reasonable no?

meredydd:
Can it be built by `npm`? With @Kaworru 's patch, what will adding Skulpt to an npm project's dependencies accomplish? It doesn't look like there's any way that `npm` will know how to generate `skulpt.js` etc.

Unless I'm misunderstanding something, looks like it would be better to either a) make this change to the `skulpt-dist` repo, as @albertjan suggests, or b) extend `package.json` further to teach `npm` how to fetch _and compile_ Skulpt from this repository.

albertjan:
@meredydd you can add script commands to npm to tell it how to build skulpt. @Kaworru it's a good idea to add those to the package.json file in this repo. But it might still be easier to add the package file to the dist repo.

Kaworru:
I created a additional pull request for skulpt-dist under:
https://github.com/skulpt/skulpt-dist/pull/4

Yes npm would be able to build the compiled version. But to build it only over npm alone I would need to add an additional "build.js" file, which should be done in an separate pull request.

In general as I mention this modifications allows you to fetch the source code easily over npm to use it as requirement in your project.

If you want an full npm integration it could be that some parts of the codes needs to be modified first.

meredydd:
My point is that, without the ability to build `skulpt.js`, fetching the source code over npm isn't much use. I think the skulpt-dist pull request is a good step forward, and if you had a full npm-driven fetch-and-build cycle for this repository I'd be happy to +1 it as a single pull request.

Kaworru:
First of all it seems there is an small miss understanding between npm and nodejs.
NPM is only a package manager, so npm alone is never be able to "build" something.
You could define the dependencies and some rules like for "npm run build" execute "python skulpt.py".

But your idea sound more like an integration from nodejs over npm to fetch and compile the code without the need to execute the python script or to have GitPython.

I checked some parts of the source code and it seems that there are a lot of changes needed to make something like this to work. See: https://github.com/skulpt/skulpt/issues/527

It would mean I need to adjust every .js file to compile the skulpt.js directly without the python script.

If this is something which is fully supported and wanted I could work on this.
But as already mention because of the many of files which needs to be adjusted I would like to split the changes in several pull request instead of one big pull request.

meredydd:
I am not an npm or node expert, but I am aware of the distinction. I'm just not sure what you want to accomplish here.

It looks like you're turning a workflow that is "git clone the Skulpt repository, `npm install`, `./m dist`" into "put the URL of the Skulpt repository into my npm project file, then `npm install` or whatever, then `cd npm-modules/skulpt && ./m dist`". I guess I'm just not sure what you're buying with this extra indirection, and was mistakenly assuming that "put this line into your npm deps and you have a `skulpt.js` ready to use in your project" was where you were heading (which is why the skulpt-dist PR sounds ace). What's the benefit here that I'm missing?

Kaworru:
Lets try to explain this from an high level.

I assume the following:
* All main files which are needed for skulpt are javascript files
* These files are compiled with the closure compiler to skulpt.js

If you have an bigger project, lets say an IDE or something like this with several modules which are all compiled with the closure compiler, you need to compile all of them together into an single file.

You can only have one compiled closure javascript file per run time, because several files would overwrite the name space, predefined settings or causing other problems.

With such package.json you could add the skulpt source as dependency and use your own closure compiler setup to compile all of your project files (including skulpt) for your project.

So the idea is not to fetch the code and build it as stand alone version.
Its more to fetch the code and build the complete project code to include skulpt.

You are right that for an webpage or simple use cases the dist file would be enough.
But as soon you need to use the closure compiler for your project you can't use such dist file without putting the code in an extra sandbox or so.

meredydd:
Aha! My missing information was "the repository in its raw format is directly useful (indeed, preferable) as an input to Closure-compiled projects".

Assuming that this is the case, then LGTM. (@Kaworru, can you confirm you've tried this workflow and it actually works without any other fiddling with the Closure command line? I have never used Skulpt as part of a larger Closure project.)

(In fact, we would do well to add a section to the README and/or skulpt.org explaining that there are two ways to get started: 1. include skulpt[.min].js in a `<script>` tag, or 2. if you already use Closure, just download this repository existing source in a Closure project, and here's a handy npm snippet that will downloaded/update it with the rest of your dependencies.)

Kaworru:
If you really want to have skulpt as simplest as possible for Closure, I need to get an comment on #527.

@meredydd 
Are you able to comment on #527 or is this something which could be better answered from other persons ?

albertjan:
I'm merging this in as it would sooner help someone then hurt anyone.

PR #529, opened 2015-10-22
Adds POST request to urllib.request.urlopen(..)

cskau:
The Python spec says urlopen issues a POST request in case the data parameter is set. (https://docs.python.org/3.0/library/urllib.request.html)

This PR adds a simple check for that and changes the request type to POST and sends the data as the request body.

### DIFF
--- a/src/lib/urllib/request/__init__.js
+++ b/src/lib/urllib/request/__init__.js
@@ -107,8 +107,16 @@ var $builtinmodule = function (name) {
      */
     request.urlopen = new Sk.builtin.func(function (url, data, timeout) {
         var xmlhttp = new XMLHttpRequest();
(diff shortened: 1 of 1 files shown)


cskau:
Polite ping.

albertjan:
@skulpt/committers I've merged this as it doesn't affect current behaviour. The reference is to the py3 docs but the functionality is the same for py2.

cskau:
Thanks for the quick response.

You are right.
For reference here is the Python 2 doc for the same method:
https://docs.python.org/2/library/urllib.html
The relevant part being:
> If the url uses the http: scheme identifier, the optional data argument may be given to specify a POST request (normally the request type is GET).

PR #535, opened 2015-12-10
Fix issue #536

jedevc:
str.count('.') now first escapes all the special regex characters.

### DIFF
--- a/src/str.js
+++ b/src/str.js
@@ -428,6 +428,7 @@ Sk.builtin.str.prototype["rpartition"] = new Sk.builtin.func(function (self, sep
 });
 
 Sk.builtin.str.prototype["count"] = new Sk.builtin.func(function (self, pat, start, end) {
+    var normaltext;
     var ctl;
     var slice;
     var m;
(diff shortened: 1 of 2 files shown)


albertjan:
Thank you very much for your contribution! Could you please add a test so it doesn't get broken again in the future?

jedevc:
Sorry, new here, which dir should the test go in?

albertjan:
In the tube atm. If you look at euther cotributing.md of hacking.rst. I
think it explains some stuff. Also look at the options for the skulpt.py
file and in the test/unit directories. Good luck!

waywaaard:
I think, your changes should already cause a couple of failing tests. Please run `python skulpt.py -dist -u -v` to get an overview about the tests. You can also pipe the results into a text file.

Currently, skulpt uses two types of tests. The old ones compare the output of skulpt with the output of real python. The 2nd ones are real ported unit tests from cpython.

albertjan:
All tests pass on travis. Or are there some with a forced output?

jedevc:
@waywaaard on my machine, all those tests seem to work. Although, it does say "Tests failed on uncompressed version".

jedevc:
And here are some tests so that it doesn't happen again.

albertjan:
Thanks! LGTM!

jedevc:
Awesome :)
I'd love to help more with this project - it's very clever.

PR #540, opened 2016-01-19
Some missing methods for generating random samples

aajanki:
Adds the following methods for generating random variables: random.triangular(), random.gauss(), random.lognormvariate() and random.expovariate()

### DIFF
--- a/src/lib/random.js
+++ b/src/lib/random.js
@@ -224,6 +224,7 @@ var $builtinmodule = function (name) {
     var mod = {};
 
     var myGenerator = new MersenneTwister();
+    var nextNormalSample = undefined;
 
     mod.seed = new Sk.builtin.func(function (x) {
         Sk.builtin.pyCheckArgs("seed", arguments, 0, 1);
@@ -368,6 +369,37 @@ var $builtinmodule = function (name) {
 
         return new Sk.builtin.float_(sample);
     });
+
+    mod.gauss = new Sk.builtin.func(function (mu, sigma) {
+        Sk.builtin.pyCheckArgs("gauss", arguments, 2, 2);
+
+        var r1, r2, u, v, s;
+
+        mu = Sk.builtin.asnum$(mu);
+        sigma = Sk.builtin.asnum$(sigma);
+
+        // Box-Muller transform
+        // (https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform)
+        // generates two independent samples from a Gaussian
+        // distribution. Return one of them and store the another one
+        // and return it next time.
+
+        if (nextNormalSample !== undefined) {
+            s = nextNormalSample;
+            nextNormalSample = undefined;
+        } else {
+            r1 = myGenerator.genrand_res53();
+            r2 = myGenerator.genrand_res53();
+            u = Math.sqrt(-2*Math.log(r1));
+            v = 2*Math.PI*r2;
+            s = u * Math.cos(v);
+            nextNormalSample = u * Math.sin(v);
+        }
+
+        return new Sk.builtin.float_(mu + sigma*s);
+    });
+
+    mod.normalvariate = mod.gauss;
 
     mod.choice = new Sk.builtin.func(function (seq) {
         Sk.builtin.pyCheckArgs("choice", arguments, 1, 1);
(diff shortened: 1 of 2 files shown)


bnmnetp:
Thanks for this.  They look like good additions to me.  @skulpt/committers One more and I'm happy to accept.

meredydd:
OK for style, can't speak to the maths.

albertjan:
Maybe a good idea to ask @waywaaard. He's quite the maths boffin. :smile:

src/lib/random.js
```diff
@@ -340,6 +341,88 @@ var $builtinmodule = function (name) {
...
+        var rnd, sample, swap;
+
+        low = Sk.builtin.asnum$(low);
+        high = Sk.builtin.asnum$(high);
+        if (low > high) {
+            swap = low;
+            low = high;
+            high = swap;
+        }
+        if (mode === undefined) {
```
waywaaard: What about passing `None` or any non numeric types here? We would need to check for those and throw an appropriate `TypeError` or `ValueError`.

src/lib/random.js
```diff
@@ -340,6 +341,88 @@ var $builtinmodule = function (name) {
         return new Sk.builtin.float_(c);
     });
 
+    mod.triangular = new Sk.builtin.func(function (low, high, mode) {
+        Sk.builtin.pyCheckArgs("triangular", arguments, 2, 3);
+
+        var rnd, sample, swap;
+
+        low = Sk.builtin.asnum$(low);
```
waywaaard: We need to check the types of the `low` and `high` arguments.

src/lib/random.js
```diff
@@ -340,6 +341,88 @@ var $builtinmodule = function (name) {
...
+            nextNormalSample = u * Math.sin(v);
+        }
+
+        return mu + sigma*s;
+    };
+    
+    mod.gauss = new Sk.builtin.func(function (mu, sigma) {
+        Sk.builtin.pyCheckArgs("gauss", arguments, 2, 2);
+
+        mu = Sk.builtin.asnum$(mu);
```
waywaaard: Argument checking required beforing passing any unwanted type to other functions which may result in an ExternalError.

src/lib/random.js
```diff
@@ -340,6 +341,88 @@ var $builtinmodule = function (name) {
...
+
+        return new Sk.builtin.float_(normalSample(mu, sigma));
+    });
+
+    mod.normalvariate = mod.gauss;
+
+    mod.lognormvariate = new Sk.builtin.func(function (mu, sigma) {
+        Sk.builtin.pyCheckArgs("lognormvariate", arguments, 2, 2);
+
+        mu = Sk.builtin.asnum$(mu);
```
waywaaard: argument checking

src/lib/random.js
```diff
@@ -340,6 +341,88 @@ var $builtinmodule = function (name) {
...
+        sigma = Sk.builtin.asnum$(sigma);
+
+        return new Sk.builtin.float_(Math.exp(normalSample(mu, sigma)));
+    });
+
+    mod.expovariate = new Sk.builtin.func(function (lambd) {
+        Sk.builtin.pyCheckArgs("expovariate", arguments, 1, 1);
+
+        lambd = Sk.builtin.asnum$(lambd);
```
waywaaard: same

src/lib/random.js
```diff
@@ -340,6 +341,88 @@ var $builtinmodule = function (name) {
...
+    mod.gauss = new Sk.builtin.func(function (mu, sigma) {
+        Sk.builtin.pyCheckArgs("gauss", arguments, 2, 2);
+
+        mu = Sk.builtin.asnum$(mu);
+        sigma = Sk.builtin.asnum$(sigma);
+
+        return new Sk.builtin.float_(normalSample(mu, sigma));
+    });
+
+    mod.normalvariate = mod.gauss;
```
waywaaard: In CPython those two methods are a little bit different. `normalvariate` uses a different algorithm `Kinderman and Monahan` and is thread safe. The `gauss` is not thread-safe and may be faster. I think those differences do not really affect your implementation and the usage of skulpt. However, I think it would be better to drop a comment in the source.

aajanki:
Updated the pull request with checks for argument types.

albertjan:
Well done @aajanki welcome to the club :+1:

PR #541, opened 2016-01-21
Unpack reverse before comparison for sorted()

### LINKED ISSUE #459: sort(reverse=False) acts as if it is sort(reverse=True)

fsladkey:
Fixes #459.

The issue is that currently, calling sorted([2, 3, 1], reverse=False) returns [3, 2, 1] (same as sorted([2, 3, 1], reverse=True). This is my first time contributing, please let me know anything I can do to make this more correct. Thanks!

### DIFF
--- a/src/sorted.js
+++ b/src/sorted.js
@@ -4,6 +4,16 @@ Sk.builtin.sorted = function sorted (iterable, cmp, key, reverse) {
     var iter;
     var compare_func;
     var list;
+    var rev;
+
+    if (reverse === undefined) {
+        rev = false;
+    } else if (reverse === Sk.builtin.none.none$) {
+        throw new Sk.builtin.TypeError("an integer is required");
+    } else {
+        rev = Sk.misceval.isTrue(reverse);
+    }
+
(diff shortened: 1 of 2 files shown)


src/sorted.js
```diff
@@ -4,6 +4,16 @@ Sk.builtin.sorted = function sorted (iterable, cmp, key, reverse) {
     var iter;
     var compare_func;
     var list;
+    var rev;
+
+    if (reverse === undefined) {
+        rev = false;
+    } else if (reverse === Sk.builtin.none.none$) {
```
waywaaard: This check is not sufficient as we can still pass a float etc. I think we need to check if the passed argument is an int/index. Bool is a subtype of int.

albertjan:
I merged this manually addressing @waywaaard comments.

PR #542, opened 2016-01-23
Implement random.sample()

aajanki:
This differs somewhat from CPython. The CPython implementation consists of two paths: a general algorithm and an optimized path for the case when k is much smaller than len(population). This pull request implements only the general case (runtime proportional to len(population) but no extra space is required). An optimized case could be implemented later, but I think it is more important to have a working version even if it is slow on large populations.

### DIFF
--- a/src/lib/random.js
+++ b/src/lib/random.js
@@ -471,5 +471,55 @@ var $builtinmodule = function (name) {
         return Sk.builtin.none.none$;
     });
 
+    mod.sample = new Sk.builtin.func(function (population, k) {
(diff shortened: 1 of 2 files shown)


albertjan:
Looks good to me! Could you rebase it please?

aajanki:
Rebased.

albertjan:
@skulpt/committers merge at will :smile:

albertjan:
Nobody seems to be really against this, so I'll merge :smile: :shipit:.

PR #543, opened 2016-01-25
Redirectable stdout and in

albertjan:
This PR adds the ability to redirect stdin and out. It handles stdin and out as "real" files. the sys module now has `stdin`, `__stdin__`, `stdout` and `__stdout__`. And I've added the `StringIO` module, so you have something to write to and read from.


albertjan:
oh yeah this also makes imports from files in directories work.

albertjan:
@skulpt/committers any of you have some time to look at this in the near future?

waywaaard:
PR is fine :+1:

eah13:
:+1: this is already in production on trinket :)

PR #544, opened 2016-01-31
Add keyword arg names for some turtle methods

### LINKED ISSUE #483: turtle Screen.setup lost its keyword arguments
setup should take keyword arguments for width and height.  I believe it did in the past as I'm having reports from teachers that student code that worked in the fall no longer works.

aajanki:
Without this patch some turtle methods accept only unnamed keywords arguments. For example, this works:

`turtle.Screen().setup(400, 400)`

but this fails with the error `got an unexpected keyword argument`:

`turtle.Screen().setup(width=400, height=400)`

Fixes #483

### DIFF
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -585,7 +585,9 @@ function generateTurtleModule(_target) {
                 {angle:this._angle, radians: this._radians}
             );
         };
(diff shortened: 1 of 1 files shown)


albertjan:
LGTM! @skulpt/committers I'll merge this, it looks sound to me.

PR #545, opened 2016-02-06
Meredydd susp try catch pr

albertjan:
replaces #518


albertjan:
@bnmnetp I can revert if you don't agree

bnmnetp:
No worries. My codehub client was showing me the merge but not the additional commit with the doc.

PR #549, opened 2016-02-27
Fix bug in input

albertjan:
This fixes a bug in the new implementation of `raw_input` that would break if no prompt was given.


meredydd:
Looks good to me (by inspection; it's simple and it has tests). I say merge.

PR #550, opened 2016-03-12
Prevent dict.__getattr__() from exploding

meredydd:
It was getting an unpleasant `AssertionError`, rather than the correct `AttributeError`. Fix it by remapping arguments correctly.

Also added my first new-style unit test for it (they're nice! Yes, I'm behind the times :-P)

### DIFF
--- a/src/dict.js
+++ b/src/dict.js
@@ -481,7 +481,7 @@ Sk.builtin.dict.prototype.__len__ = new Sk.builtin.func(function (self) {
 
 Sk.builtin.dict.prototype.__getattr__ = new Sk.builtin.func(function (self, attr) {
     Sk.builtin.pyCheckArgs("__getattr__", arguments, 1, 1, false, true);
-    return Sk.builtin.dict.prototype.tp$getattr.call(self, attr);
+    return Sk.builtin.dict.prototype.tp$getattr.call(self, Sk.ffi.remapToJs(attr));
 });
 
 Sk.builtin.dict.prototype.__iter__ = new Sk.builtin.func(function (self) {
--- a/src/dict.js
+++ b/src/dict.js
@@ -481,6 +481,7 @@ Sk.builtin.dict.prototype.__len__ = new Sk.builtin.func(function (self) {
 
 Sk.builtin.dict.prototype.__getattr__ = new Sk.builtin.func(function (self, attr) {
     Sk.builtin.pyCheckArgs("__getattr__", arguments, 1, 1, false, true);
+    if (!Sk.builtin.checkString(attr)) { throw new Sk.builtin.TypeError("__getattr__ requires a string"); }
     return Sk.builtin.dict.prototype.tp$getattr.call(self, Sk.ffi.remapToJs(attr));
 });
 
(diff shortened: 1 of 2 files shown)


rixner:
Should the type of attr be checked before calling remapToJs?  I think we probably want to explicitly fail, right?  I would advocate some tests that explicitly call ```__getattr__``` correctly and incorrectly...

meredydd:
`GenericGetAttr()` asserts that `typeof(name) == "string"`, which is the `AssertionError` that first kicked off this whole thing.

What should it be doing instead?

(In fact, if there were a general "check the type of this thing" method to make this easy, akin to `pyCheckArgs()`, I think that would be a Good Thing.)

rixner:
Philosophically, I've felt like we should check the types of Python objects passed to Python-facing functions before using them and/or converting them to Javascript.  Since ```GenericGetAttr``` is an internal function, it arguably should not need to type check anything.  I'm pretty sure that assertion is an artifact of us being fast and loose with type checking, so it got put in there at one point to help us all debug other internal functions that were calling it.

This is certainly not the only location that remains in Skulpt that has this sort of issue, where Python objects are not properly type checked.

(If you look in function.js, I wrote a whole bunch of type checking functions a long time ago - around the same time I wrote pyCheckArgs.  They are of the form ```Sk.builtin.check*```.)

meredydd:
Is this the sort of thing you were after?

I'm glad we have those `check*` functions in `function.js`, but I suspect that one of the reasons they don't get used is that they require a lot of boilerplate. For example, the updated `dict.__getattr__` I just supplied is now a 2:1 ratio of argument-checking to function calls, and they're not a concise two lines either. I mean, we should still all be doing it, but there's plenty of evidence across the codebase that people are being lazy :)

Do you think we could sensibly make a nicer 'all-in-one' function? I would quite like something like `pyCheckArgs`, but which can check types as well, so you can concisely express something like "This function takes a single argument, and it's a string", and get a suitable exception if you don't. (If we're really reaching, we might try to give it an API that doesn't require reifying the `arguments` object, which deoptimises the whole function. @LeszekSwirski has a better idea than me, but I believe that pushing `arguments` straight into `.apply()` keeps the optimiser happy...)

LeszekSwirski:
I believe that this is correct, passing `arguments` through to any other function than apply kills the optimiser.

rixner:
Yes, this seems better to me.

Theoretically, we could combine ```pyCheckArgs``` and argument type checking, but I think that would be a bit more difficult and convoluted than you might suspect.  Each argument could potentially allow multiple possible types and/or be optional.  Further, the exact wording of the error messages in Python are not always consistent.  So, I'm not sure if this would actually be any cleaner.

I did experiment with just passing ```arguments.length``` to ```pyCheckArgs``` a while ago.  I didn't see much improvement.  Though, perhaps that just means there are other things going on that are killing the optimizer, so this is a necessary first step?  It doesn't use anything other than the length, so this is certainly an easy improvement (other than the fact that we'd have to go fix a whole bunch of places in the code).

Anyway, I'm okay with accepting this.

Issue #295, opened 2014-10-24
v8 specific optimisations


LeszekSwirski:
Although optimising for v8 is a moving target, there are a couple of things we could do to improve performance on it in its current state (using information from https://github.com/petkaantonov/bluebird/wiki/Optimization-killers and other similar sources, along with `--trace-opt` on Skulpt itself). Optimisation killers include:
- Anything that leaks the `arguments` object to first copy it into an array. This includes:
  - Anything that uses `pyCheckArgs`
  - Anything that uses `Array.prototype.slice(arguments)`
- Anything that writes over function arguments, by writing first arguments to locals and then overwriting those locals:
  - In particular, `$wakeFromSuspension` clobbers function arguments
- Try/catch, which needs to be put in a separate function which then (presumably) calls our functions
  - Especially the big one in the switch code

It's hard to tell what sort of performance impact this will have without doing it (and it all needs to be done before there's a noticeable difference). However, for reference, I modified `callsim*` to correctly use the `arguments` object (copy into an array instead of slicing), and noticed a ~5-10% speed increase (on pystone) from that alone.

meredydd:
Of all of those, I'm going to vote for eliminating `try`/`catch` in generated code. Our profiling shows Not Good things. We're going to suffer a bit (because we'll need to explicitly store all the locals and temporaries in an object), but it's preventing the optimiser from even *trying* on any generated code.

I had a stab at it a couple of months ago, but it turned out to be much more invasive than I thought, so I threw it away and it's on the back burner for now.

LeszekSwirski:
Note that there's at least some movement in v8 to optimise try-catch:

http://v8project.blogspot.com/2015/07/v8-45-release.html
https://codereview.chromium.org/1694743003/

meredydd:
Well, that would be seven kinds of awesome.

PR #552, opened 2016-03-12
Improve optimisability of Sk.misceval.chain()

meredydd:
Realising the `arguments` makes the V8 optimiser bail out. While this isn't the most glaring performance bottleneck in Skulpt, it was contributing to some performance issues in Anvil.

### DIFF
--- a/src/misceval.js
+++ b/src/misceval.js
@@ -981,7 +981,13 @@ goog.exportSymbol("Sk.misceval.applyAsync", Sk.misceval.applyAsync);
  */
 
 Sk.misceval.chain = function (initialValue, chainedFns) {
-    var fs = arguments, i = 1;
+    var fs = new Array(arguments.length), i = 1;
+
+    for (i = 1; i < arguments.length; i++) {
+        fs[i] = arguments[i];
+    }
+
+    i = 1;
 
     return (function nextStep(r) {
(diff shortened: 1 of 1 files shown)


LeszekSwirski:
If you're looking for optimisation opportunities, there's #295 in general.

LeszekSwirski:
Also, since I'm here, this looks fine to me.

rixner:
Is something like this faster?

```var fs = Array.prototype.slice.call(arguments);```

Or does that also break the optimizer?

meredydd:
Unfortunately, my understanding is that that breaks the optimiser, because
the arguments object needs to be reified to pass it onto Array.slice()

rixner:
I see.  I would argue that you should put a comment in that code telling everyone that it's there to allow optimization.  Otherwise, someone might reasonably wonder what the point is and try to put it back.

If you can use ```apply```, it seems like there's probably some idiom there that could be used (like ```Array.apply(null, arguments)```?).  But, [this page](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers), at least, suggests your method.  (It also suggests a way to deal with ```try-catch```.)

albertjan:
Added in the comment.

rixner:
I think this is good to merge, then.

PR #553, opened 2016-03-12
Make the `in` operation suspendable

meredydd:
This PR makes it possible to use the `in` operator over suspendable generators.

It also introduces the `iterFor` function, which makes loops over suspending iterators a much more pleasant experience.

Testing it also required allowing unit tests to suspend.

### DIFF
--- a/skulpt.py
+++ b/skulpt.py
@@ -1036,7 +1036,9 @@ def rununits(opt=False, p3=False):
 var input = read('%s');
 print('%s');
 Sk.configure({syspath:["%s"], read:read, python3:%s});
-Sk.importMain("%s", false);
+Sk.misceval.asyncToPromise(function() {
+    Sk.importMain("%s", false, true);
+}).then(quit, function(e) { throw e; });
         """ % (fn, fn, os.path.split(fn)[0], p3on, modname))
         f.close()
         if opt:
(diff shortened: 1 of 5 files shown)


rixner:
What is the performance impact of this?

meredydd:
Unknown - do we have a standard profiling tool we can use for this sort of thing?

It's probably as good as it can be unless we manually inline `Sk.misceval.iterFor()` into `Sk.abstr.sequenceContains()`. (On the other hand, that loop was ugly enough when I tried it that I factored it out into `iterFor()`).

I don't think we can afford to break suspending generators just because someone used them in an `in` expression, so we do need something here.

meredydd:
Ping - any update on this?

@rixner - Correctness trumps performance, so unless there's an obvious Better Faster Way to Do It, I'd like to request this gets merged before someone else stubs their toe on it. (eg with the new debugger patch, any un-suspendable code is also un-debuggable)

rixner:
Is it actually "correctness" in terms of Python programs producing wrong results?  Or just that you can't suspend during such operations?

I currently have no power, so can't easily look to give instructions, but @mchat added a simple performance measurement infrastructure that you can easily use to at least tell us the slowdown.  It's runnable through skulpt.py.

albertjan:
Code:
```
import random
print int(random.random() * 100) in range(100)
```

Results on master:
`Average time over 100 iterations: 0.148655557632 seconds`

Results on this branch:
`Average time over 100 iterations: 0.14664254427 seconds`

@rixner merge?

rixner:
Looks good.  Thanks for running that.

PR #555, opened 2016-03-29
Fixes call to the PVector::normalize method in processing.

brianpmarks:
The `PVector::normalize` method was referencing an undefined variable and throwing an error any time it was called.


dsblank:
Looks good; thanks!

PR #556, opened 2016-03-30
disable failing random test

albertjan:
This test keeps failing the build, but not always, I'm not really comfortable with the random maths.

@waywaaard Do you have any idea why it doesn't always give the correct result?

I suggest we merge this and disable the test until we have a solution.


meredydd:
Agreed. This is getting really annoying. I'm +1 for this change unless it can be fixed (eg the tolerance increased a bit).

rixner:
I'm not sure this can really be fixed.  We can just make it fail less often, right?  I'm also for disabling it.

PR #557, opened 2016-04-05
changing input method in case integer value is entered in prompt.

rajnishsrivastava:
In its current implementation, the input method returns the same value as returned by the raw_input function. However, the input function is expected to return an integer value in case an integer is entered, else a string.

Please review this pull request. If changes are required in other places as part of best practices, let me know and I can work on it.

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -809,16 +809,16 @@ Sk.builtin.raw_input = function (prompt) {
 };
 
 Sk.builtin.input = function (prompt) {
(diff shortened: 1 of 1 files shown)


dsblank:
`input()` does more than just return an int; it wraps a full `eval()` around the string to return any type of item.

rajnishsrivastava:
Good point! I noticed the issue in input function in context of integer only, hence the implementation.

Let me think about the eval implementation and make some changes.

PR #559, opened 2016-04-10
Initial support for an interactive debugger

ssarangi:
This pull request adds the support for an interactive debugger. It uses the suspension framework underneath and has some features from the pdb debugger. Not all commands can be supported yet...
Supported:
1) run
2) step
3) put breakpoints
4) set ignore count
5) view all breakpoints
6) move up and down the stack
7) resume

Not supported:
1) moving the instruction pointer
2) conditional breakpoints.

I have tried my best to adhere to the coding guidelines but if you see anything which needs changing please don't hesitate to let me know. 

Please give me any feedback you might have on this PR.


waywaaard:
This is amazing. Cannot check this out. Do you have a live demo?

albertjan:
Wow this is awesome, it'll take me a while to grok it all though :smile:. I'll start looking into it asap.

rixner:
Glad to see someone working on this!  I look forward to looking through it!

ssarangi:
Looks encouraging to see the enthusiasm on this.. I have a live demo running on https://skulpt-ssarangi.c9users.io/doc/index.html. Hopefully this works since c9 tends to shut off the servers after sometime. I will look at why travis is failing but any feedback is welcome at this point :)

ssarangi:
So here are a series of steps to try it out...
Copy paste this code into the debugger.

`def foo():
    print("foo")
    
def goo():
    foo()
    print("goo")
    
def boo():
    goo()
    print("boo")
    
boo()`

This basically shows how stacks are maintained.. So here are a set of commands..

break 2
run
where
step
view locals
step
step

ssarangi:
Also since this is the first version, I think generators and yield statements won't work. I have tried out simple kernels for now and I think once that's stable then more complicated stuff can be added. Thanks to the suspension framework this wasn't as complicated as I initially expected it to be.

LeszekSwirski:
Just tried it out, _very_ cool.

ssarangi:
all thanks to the suspension framework. it is extremely well written and does most of the work anyway :)

codemercenary:
This is pretty awesome work.  An interactive debugger is an essential part of any language, great to see this finally getting added.

ssarangi:
Thanks... there are still quite a few features and unit tests which have to be written.. I will do them as this gets code reviewed.

ssarangi:
ping... just trying to see if anyone got a chance to review this any further...

albertjan:
Haven't forgotten about it :smile: will get to it next week.

skulpt.py
```diff
@@ -714,6 +714,7 @@ def dist(options):
         print ". Running tests on uncompressed..."
 
     ret = test()
+    ret = 0
```
albertjan: 😄 this should probably not be in there

skulpt.py
```diff
@@ -753,9 +757,10 @@ def dist(options):
     doc()
 
     try:
-        shutil.copy(compfn, os.path.join(DIST_DIR, "tmp.js"))
```
albertjan: Somewhere the build script relies on this file existing at some point, could you add these lines back in?

debugger/debugger.js
```diff
@@ -0,0 +1,310 @@
+/**
+ * Debugger support for skulpt module
+ */
+
+var Sk = Sk || {}; //jshint ignore:line
+
+function hasOwnProperty(obj, prop) {
+    var proto = obj.__proto__ || obj.constructor.prototype;
```
albertjan: this `__proto__` property is something google closure adds? I don't remember ever running into it in skulpt.

albertjan:
Could you add the files in `doc/static/debugger` to `.gitignore`?

albertjan:
Sorry I'm adding these comments in such a piecemeal way. If you look at the lines below 218 in skulpt.py, you'll see `debugger/**` isn't covered by jscs and jshint, could you add them there? you'll see there are a couple errors nothing too bad.

Maybe it's a good idea to leave the debugger_cm file in `doc/static/debugger` and debugger in `/debugger` there's no reason to lint the debugger_cm file (it will be a lot of work and non of the other interface stuff is linted).

Other than that, it looks very good to me, I'm going to do a bit more fiddling to to see if there obvious broken stuff. And if I can't find it I'd love to merge this so we can start fiddling with it irl asap. 

It's very nice there are no changes to the core code.

waywaaard:
I have not read through all of the code, but is the local variable listing working? Are there some fixes, that we could use for the `locals()`  builtin?

albertjan:
@LeszekSwirski @meredydd could you have a quick look at the suspension implementation? It works great, but I'd like to see this as an example implementation of the suspension stuff.

LeszekSwirski:
Once I have a spare minute, probably in a week or two, sure. I need to refresh my memory on suspension stacks first though.

rixner:
Can we put the fixes to locals and suspensions into a different pull request?  Can those be evaluated and pulled independently.

albertjan:
@rixner I think both are connected directly to the debugger, not the base implementation of locals and suspensions. @ssarangi amirite?

ssarangi:
@albertjan: thats correct. there are absolutely no changes to the core code at all. i didnt want to break anything there in the first version :)

rixner:
My misunderstanding, then.  I obviously need to find the time to take a closer look.

albertjan:
@ssarangi ping! :smile:

codemercenary:
Also ping here, too.

ssarangi:
Apologize for the delay... I will get the updated code out hopefully by the weekend.

albertjan:
@ssarangi awesome, I was just making sute you hadn't forgotten about it yet. It was a very friendly reminder 😄

ssarangi:
@albertjan : No definitely not.. :) I will have it very soon... sorry again

ssarangi:
@albertjan : I committed all the changes from the jslinting as well as moved the debugger_cm into the static/debugger folder. However, I also kept the debugger.js file in the static/debugger folder because the index.html includes it. Also added the suggestions for skulpt.py. 

Once this is merged, I have a couple more fixes which I need to do.

debugger/debugger.js
```diff
@@ -0,0 +1,310 @@
...
+    }
+    
+    var key = this.generate_breakpoint_key(filename, lineno, colno);
+    if (hasOwnProperty(this.dbg_breakpoints, key) &&
+        this.dbg_breakpoints[key].enabled === true) {
+        var bp = null;
+        if (hasOwnProperty(this.tmp_breakpoints, key)) {
+            delete this.dbg_breakpoints[key];
+            delete this.tmp_breakpoints[key];
+            return;
```
rixner: Shouldn't this explicitly return ```true``` or ```false```?

debugger/debugger.js
```diff
@@ -0,0 +1,310 @@
...
+            
+            if (this.suspension_stack.length === 0) {
+                this.print("Program execution complete");
+                return;
+            }
+            
+            var parent_suspension = this.get_active_suspension();
+            // The child has completed the execution. So override the child's resume
+            // so we can continue the execution.
+            parent_suspension.child.resume = function() {};
```
rixner: I'm having a hard time wrapping my head around why this is needed.  Can you explain?

rixner:
I finally had some time to look through this.  I like it!  I made two line comments, but otherwise, things look pretty good to me.

I think the fact that everything is outside the core of skulpt makes it pretty safe to merge something like this.  Even if we discover bugs later, they will not have disrupted the behavior of Skulpt when the debugger is not in use.  And having it merged will let a bunch of people use and stress test it to tease out any bugs.

debugger/debugger.js
ssarangi: You are right... Thanks for catching this. I fixed it and checked it in.

debugger/debugger.js
ssarangi: @rixner 
So suspension stacks are recursive by nature in the way they are being returned from the child. This example below will kind of explain what happens. 

def foo(a, b):
    a = a + 1
    print("foo")

def goo(c, d):
    foo(c+1, d+1)
    print("goo")

def boo(e, f):
    goo(e+1, f+1)
    print("boo")

boo(1, 2)

Now lets say we put a breakpoint in foo(1, 2)... So at this point we have a suspension stack starting from main -> boo -> goo -> foo..

Now once foo completes execution, then we come to goo. At this point, unless we override the resume function for foo, when goo's execution resumes you will see that it again starts executing foo. That is because whenever a suspension resumes it first run's the child's suspension (or resume method). That is what is happening. It's just like a stack call when I am unwinding the stack but apart from that I sort of mark that the child has completed execution so no need to re-execute it. 

I know its a slightly awkward explanation, but let me know if this makes sense..

ssarangi:
@rixner : Thank's for grokking through this. I agree. It would be easier to stress test this and file bugs on this. So let me know if I need to make any more changes for this to be merged. :)

bnmnetp:
OK, I've looked at it a bit too.  I don't understand it well enough, but I'm hoping I can learn enough from whats going on in here to make a pythontutor integration work.

As @rixner points out this doesn't look like it will affect any of the core functionality so I'll go ahead and accept it.

ssarangi:
Thanks a lot for getting this merged... I will wait for feedback for bugs and functionality from all of you.

ssarangi:
@bnmnetp : Thanks Brad... This needs to be merged into gh-pages too for the html changes and the debugger ui changes.

albertjan:
🎉🎉🎉

debugger/debugger.js
rixner: Okay, but I'm still not convinced that this is necessary.  When foo suspends, you should get a suspension back for main, which has a child suspension for boo, which has a child suspension for goo, which has a child suspension for foo.  Then, when you want to resume, you resume the main suspension, which recursively resumes the children first, so foo will be resumed first.  If goo later suspends after foo returns because it hits another breakpoint, you should get a new suspension for main, with a child for boo, with a child for goo.

Instead, you are resuming the foo suspension manually.  Which is what causes the "problem", where the goo suspension still has a stale foo child.  I'm not sure, but I believe this can cause problems when a function actually returns a value.  I think in some circumstances, your method could lose that return value.

So, rather than keeping around the suspension stack and managing it manually, it seems like you should let the suspension infrastructure do it for you (and maybe push it onto an array each time you get a suspension if you'd prefer to use it as an array, rather than the linked list you get, when allowing the user to navigate the call stack).

This was the source of my original question.  It seems like you should not have to clear out that resume function if you use it the way it was originally intended.  So, my question was really why are you doing it this way instead of using the suspension list.  I wanted to know if I am missing something, as I still can't wrap my head around why you need to do it this way...

(Or perhaps I am totally incorrect about how the suspension infrastructure currently works...)

ssarangi: @rixner I see... that makes sense. I might have misunderstood the whole thing. Let me rework it this way but I think I see your approach and point here.

PR #561, opened 2016-04-18
Remap bools correctly in Sk.ffi.remapToPy()

meredydd:
JS bools were not being turned into `Sk.builtin.bool`s. This patch fixes that.


rixner:
Looks fine to me.  I can't believe this is still there...

PR #562, opened 2016-04-21
For processing: Adds alpha arg to fill and green, blue, alpha args to colorMode

brianpmarks:
Processing's `fill` and `colorMode` didn't have full support for alpha and RGB/HSB.


dsblank:
Looks good to me; thanks!

PR #568, opened 2016-05-03
Make custom iterators suspendable, and fix StopIteration

meredydd:
`raise StopIteration` used to be a horrible hack; it is now an exception (as in Python) that is caught in the shim code between `tp$iternext` and user code.

Also makes `next()` suspendable (but not `__iter__`, yet)

### DIFF
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -918,7 +918,7 @@ Sk.abstr.iter = function(obj) {
             try {
                 ret = Sk.misceval.callsim(this.getitem, this.myobj, Sk.ffi.remapToPy(this.idx));
             } catch (e) {
-                if (e instanceof Sk.builtin.IndexError) {
+                if (e instanceof Sk.builtin.IndexError || e instanceof Sk.builtin.StopIteration) {
                     return undefined;
                 } else {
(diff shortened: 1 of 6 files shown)


albertjan:
@skulpt/committers LGTM merge?

rixner:
So, this looks like ```iternext``` still returns ```undefined``` to indicate the end of the loop?  So, now ```StopIteration``` is used mostly internally, but the Javascript from user Python code still expects the old semantics?  (I'm not saying this is a bad thing, I'm just trying to understand what's happening now.)

I think this results in some funny behavior with programs like:

```
lst = [1, 2, 3]
for item in lst:
    print item
    raise StopIteration
```

Obviously, what you have here is better than what we currently have, though!

rixner:
I see, my bad.  The problem is that our exceptions don't have proper string names.  Where Python would print the string value of the exception, we print ```undefined``` because we never defined the name.  This has nothing to do with the ```StopIteration``` you added, but would happen with any exception thrown inside that loop.

I'm good to merge this.

albertjan:
many thanks again @meredydd

PR #572, opened 2016-05-05
Add strftime() and strptime() implementations to time module

meredydd:
As far as I'm aware, these aren't available in the Closure library (they have some nifty date-time manipulation stuff that could be used to implement `strftime()` from scratch, but no `strptime()`), so I've bundled MIT-licenced implementations (for source see READMEs).


albertjan:
`Test threw exception in  TimeTestCase.test_sleep  (SuspensionError: Cannot call a function that blocks or suspends here on line 26)`

meredydd:
That's...weird. It's working for me. Will investigate, possibly on Monday.

PR #573, opened 2016-05-17
Support keyword arguments properly

meredydd:
Also removes the copy-and-pasted code between `method` and `func` - a method call just sticks `self` onto the beginning of its arguments and then calls the underlying function.


albertjan:
lol I was just lookin into this too :smile:

albertjan:
Looks good to me! @skulpt/committers merge at will! :smile:

waywaaard:
As far as I can tell, reduces the PR the complexity of handling of calling functions/methods. 

LGTM

PR #574, opened 2016-05-18
Fix tuple comparison to other types and fix failing tests.

rixner:
The recently added test_calling.py fails when called from CPython.  The problem is that ```*args``` is a tuple, but the assertions were expecting a list.  I think this highlights that we should run the unittests through CPython in travis so that we can test the tests, so to speak, before we merge them.

Fixing the tests exposed a bug.  The bugfix is simple, as we were not properly checking if we were comparing to a different type in the tuple ```tp$richcompare``` function.

### DIFF
--- a/src/tuple.js
+++ b/src/tuple.js
@@ -154,7 +154,8 @@ Sk.builtin.tuple.prototype.tp$richcompare = function (w, op) {
     var wl;
     var vl;
     var v;
-    if (!w.__class__ || !Sk.builtin.isinstance(w, Sk.builtin.tuple)) {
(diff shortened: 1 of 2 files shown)


rixner:
9 of the 32 unittest test files have some sort of failure/error in them when I run them through CPython 2.7 on my machine.  (Not including test_calling.py.)

I have not investigated how serious these problems are or are not.

albertjan:
Good catch! Yeah we should run the tests on cpython on travis. I'll see if I can look into that. This looks good btw.

PR #577, opened 2016-05-20
take timezone into consideration

albertjan:
Fixes failing test from #517


rixner:
This fixes the time test for me.

PR #578, opened 2016-05-20
Use test.test_support to be compatible with CPython.

rixner:
Updates to allow more unittest files to pass in CPython.

This one requires some thought.  I am not sure that the change to import.js is correct.  It doesn't break any of the other tests, but I'm not sure if the change should be here or if ```Sk.importModuleInternal_``` should be returning something different.  So, I'd like to have multiple people look this over and think about it.

There were attempts to get importing test_support working in #310 and #450, but they are out of date at this point, so I could not just use that solution, but it may be instructive for others to look at those ideas.

(Incidentally, I also uncommented some tests in test_index.py that now pass because we have implemented more features!)

### DIFF
--- a/src/import.js
+++ b/src/import.js
@@ -574,8 +574,8 @@ Sk.builtin.__import__ = function (name, globals, locals, fromlist) {
                 if (fromName != "*" && ret.$d[fromName] == null && (ret.$d[dottedName] != null || ret.$d.__name__.v == dottedName)) {
                     // add the module name to our requiredImport list
                     fromImportName = "" + name + "." + fromName;
-                    fromNameRet = Sk.importModuleInternal_(fromImportName, undefined, undefined, undefined, true, currentDir);
-                    ret["$d"][fromName] = Sk.abstr.gattr(fromNameRet, fromName, true);
+                    fromNameRet = Sk.importModuleInternal_(fromImportName, undefined, undefined, undefined, false, currentDir);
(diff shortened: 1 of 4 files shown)


albertjan:
I'm struggling to see what the change in the `__import__` will do, as far as I've been able to see, the variable you're assigning is already there. So when you `from test import test_support` by the time we reach `ret["$d"][fromName] = Sk.abstr.gattr(fromNameRet, fromName, false);`. `ret["$d"][fromName]` will already have what comes back from the `Sk.abstr.gattr`. That may not be true in all cases it's just what I saw, from my first exploration.

rixner:
What I was seeing is that ```ret["$d"][fromName]``` would get the result from ```Sk.importModuleInternal_```, which is the parent module.  So, I changed it.  It's possible that things were already set and I just needed to delete the line ```ret["$d"][fromName] = fromNameRet;```, which was setting the "test_support" name to map to the <module test>.  As I said, I'm not sure what the correct behavior in all cases should be...

rixner:
@albertjan, you were correct, I should just remove line 578 because it is unnecessary, rather than trying to change it.  I just did that and it seems to work for me.

bnmnetp:
I've built, and tested all the different import paths in my books, and I have no problems with this one.

Brad

rixner:
@albertjan are you good to merge this now, then?

albertjan:
👍

Issue #560, opened 2016-04-14
messy state with multiple time.sleep runs


eah13:
When I repeatedly run this code (as in run it again when it's in the middle of executing)

```
import time

for i in range(10):
  print(i)
  time.sleep(1)
```

I get muddled results, as if some part of the stack isn't getting cleaned:

```
0
4  
1 
1
2
3
4
5
6
7
8
9
```

The `4` in this case corresponds to re-running the program around the time that i = 4.  The `4` I can understand.  But what's up with that second `1`?

I reproduced in vanilla skulpt too but here's a trinket to easily repro yourself: https://trinket.io/python/3541e47727  Press run then press run again once the loop's gone thru a few iterations.

(Aside: I wonder if this may be related to some hard-to-diagnose Turtle state bugs we've seen that also seem to preserve odd state across repeated runs and now that I think about it probably also used time.sleep.  If thats the case I'll open an issue refing this one)  

Thoughts?  Perhaps I'm misdiagnosing what's going on.  Hopefully there's a magic bullet to clear all leftover state/stack stuff at the beginning of the new run and this is a small one.

LeszekSwirski:
That's a pretty interesting bug, nice catch. My guess for the second 1 is that it's a print from the previous invocation printing the new state of i. Perhaps fixing this is as simple as waiting for the previous invocation to drain before starting a new one?

rixner:
I think that the problem here is that there is a ```setTimeout``` timer that is active when you are trying to reset the state.  This timer is not exposed anywhere, so you can't kill it.  I don't think you can just wait for it to drain - what if the call was to sleep for 5 minutes?  To solve this, I suspect we need to expose the timer ID somewhere so that you can call ```clearTimeout``` on that timer.

This is a general problem with using ```setTimeout``` within Skulpt, as the timer is not exposed to the enclosing controller so that it can easily be killed when necessary.  I think a better solution is to add another option to ```Sk.configure``` where you can provide your own ```setTimeout``` equivalent for all the places that it is used (currently image, turtle, time, and misceval).  Then, it could default to just ```setTimeout``` for those who don't care (or maybe a function that checks if you are in a browser or not and does the appropriate thing), or you could provide a function that keeps the ID's so they can be cleared for those who want to be able to reset them all.  I'm not 100% certain, but it seems like there should only ever be one timer active at any given time, so you would just have to clear that one...

PR #579, opened 2016-05-24
Add infrastructure to allow Skulpt to be reset.

rixner:
This adds some infrastructure to allow the problem in #560 to be addressed.  Currently, if you want to reset Skulpt, you have to do so externally in the code that is driving Skulpt.  This PR allows that external code visibility to see any pending timers.  If you don't change anything, it defaults to behaving as it does now.  But, if you set ```setTimeout``` in the call to ```Sk.configure```, you can manage the timers.

The demo infrastructure doesn't have any sort of "reset" button, so here's a simple example of how one might use this:

```
var timers = [];

var timeout = function (func, delay) {
    var id;
    var wrapper = function () {
        var idx = timers.indexOf(id);
        if (idx > -1) {
            timers.splice(idx, 1);
        }
        func();
    };
    id = setTimeout(wrapper, delay);
    timers.push(id);
};

var clear_timeouts = function () {
    var idx;
    for (idx = 0; idx < timers.length; idx++) {
        clearTimeout(timers[idx]);
    }
    timers = [];
};
```

You would then set ```setTimeout``` in the call to ```Sk.configure``` to be ```timeout``` and you would call ```clear_timeouts``` when you want to reset Skulpt.  I am pretty sure there can only be one pending timer at any given time, but figured I'd give a more general example in case that changes in the future.  Obviously, one could choose to do this differently, as well.

I used the infrastructure in this PR to solve the problem described in #560 in CodeSkulptor (although the symptoms were different in CodeSkulptor, as the behavior you see depends on how you handle suspensions in the Skulpt driver), so I think these changes are sufficient, it is just up to the user to decide how to put the reset machinery into their driver.

This requires that anyone wanting to add a call to ```setTimeout``` in Skulpt in the future instead use ```Sk.setTimeout```.  I don't use turtle and image and they directly call ```window.setTimeout```, so I wasn't sure what behavior people want in those modules.  So, I left them alone, but they could be easily modified to use ```Sk.setTimeout```, as well so that they could be reset.

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -58,6 +58,16 @@ Sk.configure = function (options) {
     Sk.breakpoints = options["breakpoints"] || function() { return true; };
     goog.asserts.assert(typeof Sk.breakpoints === "function");
 
+    Sk.setTimeout = options["setTimeout"];
+    if (Sk.setTimeout === undefined) {
+        if (typeof setTimeout === "function") {
+            Sk.setTimeout = setTimeout;
+        } else {
+            Sk.setTimeout = function(func, delay) { func(); };
+        }
+    }
+    goog.asserts.assert(typeof Sk.setTimeout === "function");
+
     if ("execLimit" in options) {
         Sk.execLimit = options["execLimit"];
     }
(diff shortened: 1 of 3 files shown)


albertjan:
LGTM

eah13:
Looks good.  We'll definitely want to get Turtle and image hooked up like this, perhaps using this option value or perhaps a distinct one if we can think of a scenario when you'd want to define a different timeout for just the visual modules.

PR #580, opened 2016-05-26
Descriptors

bnmnetp:
This takes the work from the decorators branch, and gets it cleanly up to date with master.  I propose that we focus on getting the descriptor part to work in this PR, which I think is pretty close.  Then make a second PR when we are ready to get decorators fully working.

The great thing about this branch is that @waywaaard 's python only implementation of property works!  I would really like to get that into a builtin because I'm looking at some changes to the book where I would use properties.

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -1632,6 +1632,9 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
         out(scopename, ".$defaults=[", defaults.join(","), "];");
     }
 
+    if (decos.length > 0) {
+        out(scopename, ".$decorators=[", decos.join(","), "];");
+    }
 
     //
     // attach co_varnames (only the argument names) for keyword argument
@@ -1687,7 +1690,14 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
         }
     }
     else {
-        return this._gr("funcobj", "new Sk.builtins['function'](", scopename, ",$gbl", frees, ")");
+        var res;
+        if (decos.length > 0)
+        {
+            res = this._gr("funcobj", "Sk.misceval.callsim(", scopename, ".$decorators[0], new Sk.builtins['function'](", scopename, ",$gbl", frees, "))"); // scopename, ".$decorators[0](new Sk.builtins['function'](", scopename, ",$gbl", frees, "))";
+        } else {
+            res = this._gr("funcobj", "new Sk.builtins['function'](", scopename, ",$gbl", frees, ")");
+        }
+        return res;
     }
 };
 
--- a/src/compile.js
+++ b/src/compile.js
@@ -1693,7 +1693,7 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
         var res;
         if (decos.length > 0)
         {
-            res = this._gr("funcobj", "Sk.misceval.callsim(", scopename, ".$decorators[0], new Sk.builtins['function'](", scopename, ",$gbl", frees, "))"); // scopename, ".$decorators[0](new Sk.builtins['function'](", scopename, ",$gbl", frees, "))";
+            res = this._gr("funcobj", "Sk.misceval.callsimOrSuspend(", scopename, ".$decorators[0], new Sk.builtins['function'](", scopename, ",$gbl", frees, "))"); // scopename, ".$decorators[0](new Sk.builtins['function'](", scopename, ",$gbl", frees, "))";
         } else {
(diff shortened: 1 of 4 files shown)


waywaaard:
@bnmnetp albertjan made some modifications (no upstream PR yet) to implement pure python based built-ins.

For testing, here are pure Python versions for `classmethod` and `staticmethod`.

```python
class classmethod(object):
    "Emulate PyClassMethod_Type() in Objects/funcobject.c"

    def __init__(self, f):
        self.f = f

    def __get__(self, obj, klass=None):
        if klass is None:
            klass = type(obj)
        def newfunc(*args):
            debugger
            return self.f(klass, *args)
        return newfunc


class staticmethod(object):
    "Emulate PyStaticMethod_Type() in Objects/funcobject.c"

    def __init__(self, f):
        self.f = f

        return self.f

class test(object):
    def __init__(self, f):
        self.f = f
        
    @classmethod
    def cool_cm(cls, val):
        print cls, val
    
    @staticmethod
    def cool_sm(val):
        print val


t = test(2)
test.cool_sm(5)
test.cool_cm(4)
```python

**Comment by bnmnetp** [comment]
Thanks @waywaaard It would be nice to get a fix for that closure problem exercised in the classmethod get function.  But someone with more compiler experience is going to have to tackle that one.

@albertjan implementing a builtin in pure python?    Why are you holding back? 😄

File: src/compile.js [hunk 1]
```diff
@@ -1687,7 +1690,14 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
         }
     }
     else {
-        return this._gr("funcobj", "new Sk.builtins['function'](", scopename, ",$gbl", frees, ")");
+        var res;
+        if (decos.length > 0)
+        {
+            res = this._gr("funcobj", "Sk.misceval.callsim(", scopename, ".$decorators[0], new Sk.builtins['function'](", scopename, ",$gbl", frees, "))"); // scopename, ".$decorators[0](new Sk.builtins['function'](", scopename, ",$gbl", frees, "))";
```
albertjan: do we need to call with `callsimOrSuspend` here?

src/misceval.js
```diff
@@ -1187,6 +1187,14 @@ Sk.misceval.applyOrSuspend = function (func, kwdict, varargseq, kws, args) {
         //append kw args to args, filling in the default value where none is provided.
         return func.apply(null, args);
     } else {
+        fcall = func.__get__;
+        if (fcall !== undefined) {
+            fcall = Sk.misceval.callsim(func.__get__, func, func);
```
albertjan: and here?

src/object.js
```diff
@@ -37,6 +37,20 @@ Sk.builtin.object.prototype.GenericGetAttr = function (name) {
...
     dict = this["$d"] || this.constructor["$d"];
 
+    descr = Sk.builtin.type.typeLookup(tp, name);
+
+    // otherwise, look in the type for a descr
+    if (descr !== undefined && descr !== null && descr.ob$type !== undefined) {
+        f = descr.ob$type.tp$descr_get;
+        if (!(f) && descr["__get__"]) {
+            f = descr["__get__"];
+            return Sk.misceval.callsim(f, descr, this);
```
albertjan: and here?

albertjan:
Added some comments so we can suspend in decorators 😄

bnmnetp:
Yep, you are probably correct.  I'll give that a shot.

waywaaard:
@albertjan: Those changes should our current problems, right?

albertjan:
@waywaaard I believe so :)

rixner:
I am not convinced that this should have been merged into master yet.  I am fairly certain that the changes to misceval.js, for instance, are incorrect and will cause problems.

albertjan:
Ah ok I'm sorry. I'll revert when I get home.

albertjan:
I'm going to revert this per @rixner 's comment. We should find a test that makes this bug show.

rixner:
Unfortunately, I don't think the test suite really forces suspensions to be returned everywhere in order to confirm that all code handles suspensions correctly.  I think @meredydd just thinks it all through and fixes the mistakes the rest of us make... :smile:

meredydd:
The command you're looking for is `./m testdebug`. That runs the tests in debug mode, which generates suspensions between every line of code.

Perhaps we should add this to the `dist` tests?

meredydd:
(Of course, the debug suspensions are `optional`, so this won't verify that your code can handle all suspensions. But it's fairly good at making sure every function call that could get a suspension returned, does.)

meredydd:
Ah! I just checked out `skulpt.py`, and it looks like the `debug_mode` flag is not honoured for the new-style unit tests. I should submit a PR. In the meantime, use `./m rundebug <filename>.py`.

PR #587, opened 2016-06-14
Invokes setTimeout call in time.sleep with window as this.

### LINKED ISSUE #584: time.sleep returning instantly
time.sleep is resolving instantly in the lastest versions of skulpt

Try this code on skulpt.org:

```python
import time
start = time.time()
time.sleep(1)
finish = time.time()
print(finish-start)
```
I get ~`.001` instead of ~`1.001`

Perhaps related to #579?  It's as if `Sk.setTimeout` isn't getting set correctly in env.js

brianpmarks:
As far as I can tell this solves the `Illegal Invocation` error I was seeing in Chrome. May also resolve #584 but I'm not sure if `setTimeout` is overridden on skulpt.org.

I tested this in Chrome both with and without `setTimeout` passed into `Sk.configure`. I think other `Sk.setTimeout` calls may also need to be invoked this way.

### DIFF
--- a/src/lib/time.js
+++ b/src/lib/time.js
@@ -65,7 +65,7 @@ var $builtinmodule = function (name) {
         var susp = new Sk.misceval.Suspension();
         susp.resume = function() { return Sk.builtin.none.none$; }
         susp.data = {type: "Sk.promise", promise: new Promise(function(resolve) {
-            Sk.setTimeout(resolve, Sk.ffi.remapToJs(delay)*1000);
+            Sk.setTimeout.call(window, resolve, Sk.ffi.remapToJs(delay)*1000);
         })};
         return susp;
     });


rixner:
Does the one line change in env.js I suggested in #584 solve your problem?  If so, I think that's a better solution, as it will fix this everywhere it is used.  I was able to replicate the problem and that fixes it for me.

As a reminder, change line 64 in env.js (where ```Sk.setTimeout``` is set to ```setTimeout```) to:

```Sk.setTimeout = function(func, delay) { setTimeout(func, delay); };```

brianpmarks:
@rixner Ah, sorry, I missed that earlier comment. My apologies. Just tried your suggestion and yes, it does work for me.

brianpmarks:
Closing this PR in lieu of a global fix.

PR #588, opened 2016-06-14
Run setTimeout within the global context.

rixner:
Fix ```Illegal Invocation``` error introduced by #579.  Most (all?) browsers need ```setTimeout``` to run within the global context.


rixner:
Can we get this merged, as I believe master is currently broken for most everyone?  I was able to reproduce the problem and this fixes it for me.  It also fixes it for @brianpmarks .  It does not fix it for @eah13, but he says his problem predates #579.

PR #589, opened 2016-06-16
Implements dict.copy

waywaaard:
Adds method and original unit test.


waywaaard:
Based from wrong upstream/master. Sorry, the right PR comes later.

PR #590, opened 2016-06-16
Fixes #583 Broken Import for dotted names in from list

### LINKED ISSUE #583: something in import broke
OK, So this was working great until I discovered today that broke sometime in the last several months.

```
from unittest.gui import TestCaseGui
```
This used to work just fine, now I get the error:  ``ImportError: No module named unittest.gui.TestCaseGui``

Anyone with any ideas on when or how this broke it would be much appreciated.

Brad

waywaaard:
See issue #583 

With the fix both ways are working:

```python
from unittest.gui import TestCaseGui
```

and
```python
from unittest import gui as mystuff
```

I've added two unit tests, so that we do not break those again in future.

### DIFF
--- a/src/import.js
+++ b/src/import.js
@@ -559,6 +559,8 @@ Sk.builtin.__import__ = function (name, globals, locals, fromlist) {
         if (saveSk !== Sk.globals) {
             Sk.globals = saveSk;
         }
+
+        // There is no fromlist, so we have reached the end of the lookup, return
         if (!fromlist || fromlist.length === 0) {
             return ret;
         } else {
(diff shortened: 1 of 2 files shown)


bnmnetp:
OK, I just grabbed the PR and built a new distribution to try in my code.  The good news is that the import works, **sort of**

When the tests actually go to run, I'm not getting an error: ``Error: AttributeError: 'myTests' object has no attribute 'setup' on line 28``

So, setup is defined in TestCase, which TestCaseGui extends, and then MyTests extend TestCaseGui. I'm not sure if we have an inheritance tests for three levels or if the problem is somewhere else, I haven't had time to investigate any further.

bnmnetp:
Ah, now I understand @rixner renamed ``setup`` to ``setUp`` in TestCase which appears to be the correct way to spell it according to the doc.

bnmnetp:
With that figured out, this looks good to me.  I'll be submitting a separate PR that has changes to gui.py and fixes the incorrectly named call to setup.

waywaaard:
There is still one bug with import and packages.

Current folder layout:
```
|-test_lib/
|---__init__.py
|---test_lib.py
`---other.py
```

and inside the content of `__init__.py`:
```python
from .test_lib import Foo, Foo as Bar
from .other import FooBar, FooBarFoo
```

It fails importing inside `__init__.py` in the first line. The `test_lib.py` is imported and stored as a module on `Sk.sysmodules`, but no `$d` dictionary is created and the lib is stored as `$js`.

The same setup works when I rename the `test_lib.py` to `test.py`

rixner:
@waywaaard , do you want to fix that too, or have this merged and leave the test_lib problem as an open issue?  Also, does it make sense to fold the fix from #578 into this PR (the actual import fix there is only really a few lines, the rest of it is about the testing infrastructure, which should remain independent)?

bnmnetp:
I would really like to have this accepted, so that I can move on with other things that depend on it, so I would vote to make the test_lib independent.  I would be happy to test #578 so we can accept that one too.

rixner:
I am good with this. I was waiting for a response on test_lib, before merging.  I will merge this now.  I would appreciate someone vetting and merging #578, as well.

PR #593, opened 2016-06-24
Decorators 2

### LINKED ISSUE #18: decorators not implemented

albertjan:
Hi all here is the new decorators branch. But I still have a small issue. @meredydd  could you have a look at this:

```py
def decofunc(fn):
    return fn

@decofunc
def func():
    print "called"

func()
```

this fails because the `@decofunc` makes `func` into a suspension. And then it fails in `applyOrSuspend`.

I want to add the real tests from cpython for the property decorator.

closes #18 #426 #326

### DIFF
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -828,7 +828,7 @@ Sk.abstr.gattr = function (obj, nameJS, canSuspend) {
         ret = Sk.misceval.callsimOrSuspend(f, new Sk.builtin.str(nameJS));
     }
 
-    var ret = Sk.misceval.chain(ret, function(ret) {
+    ret = Sk.misceval.chain(ret, function(ret) {
         var f;
 
         if (ret === undefined && obj.tp$getattr !== undefined) {
@@ -842,7 +842,6 @@ Sk.abstr.gattr = function (obj, nameJS, canSuspend) {
                 }
             }
         }
-
         return ret;
     }, function(r) {
         if (r === undefined) {
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -828,20 +828,24 @@ Sk.abstr.gattr = function (obj, nameJS, canSuspend) {
         ret = Sk.misceval.callsimOrSuspend(f, new Sk.builtin.str(nameJS));
     }
 
-    ret = Sk.misceval.chain(ret, function(ret) {
+    var eventualresult = Sk.misceval.chain(undefined, function () {
+        return obj;
+    }, function(obj1) {
         var f;
+        objname = Sk.abstr.typeName(obj1);
 
-        if (ret === undefined && obj.tp$getattr !== undefined) {
-            ret = obj.tp$getattr(nameJS);
+        if (ret === undefined && obj1.tp$getattr !== undefined) {
+            ret = obj1.tp$getattr(nameJS);
 
             if (ret === undefined) {
-                f = obj.tp$getattr("__getattr__");
+                f = obj1.tp$getattr("__getattr__");
 
                 if (f !== undefined) {
                     ret = Sk.misceval.callsimOrSuspend(f, new Sk.builtin.str(nameJS));
                 }
             }
         }
+        
         return ret;
     }, function(r) {
         if (r === undefined) {
@@ -850,7 +854,7 @@ Sk.abstr.gattr = function (obj, nameJS, canSuspend) {
         return r;
     });
 
-    return canSuspend ? ret : Sk.misceval.retryOptionalSuspensionOrThrow(ret);
+    return canSuspend ? eventualresult : Sk.misceval.retryOptionalSuspensionOrThrow(eventualresult);
 };
 goog.exportSymbol("Sk.abstr.gattr", Sk.abstr.gattr);
 
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -828,24 +828,21 @@ Sk.abstr.gattr = function (obj, nameJS, canSuspend) {
         ret = Sk.misceval.callsimOrSuspend(f, new Sk.builtin.str(nameJS));
     }
 
-    var eventualresult = Sk.misceval.chain(undefined, function () {
-        return obj;
-    }, function(obj1) {
+    var ret = Sk.misceval.chain(ret, function(ret) {
         var f;
-        objname = Sk.abstr.typeName(obj1);
 
-        if (ret === undefined && obj1.tp$getattr !== undefined) {
-            ret = obj1.tp$getattr(nameJS);
+        if (ret === undefined && obj.tp$getattr !== undefined) {
+            ret = obj.tp$getattr(nameJS);
 
             if (ret === undefined) {
-                f = obj1.tp$getattr("__getattr__");
+                f = obj.tp$getattr("__getattr__");
 
                 if (f !== undefined) {
                     ret = Sk.misceval.callsimOrSuspend(f, new Sk.builtin.str(nameJS));
                 }
             }
         }
-        
+
         return ret;
     }, function(r) {
         if (r === undefined) {
@@ -854,7 +851,7 @@ Sk.abstr.gattr = function (obj, nameJS, canSuspend) {
         return r;
     });
 
-    return canSuspend ? eventualresult : Sk.misceval.retryOptionalSuspensionOrThrow(eventualresult);
+    return canSuspend ? ret : Sk.misceval.retryOptionalSuspensionOrThrow(ret);
 };
 goog.exportSymbol("Sk.abstr.gattr", Sk.abstr.gattr);
 
--- a/src/compile.js
+++ b/src/compile.js
@@ -1630,6 +1630,9 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
         out(scopename, ".$defaults=[", defaults.join(","), "];");
     }
 
+    if (decos.length > 0) {
+        out(scopename, ".$decorators=[", decos.join(","), "];");
+    }
 
     //
     // attach co_varnames (only the argument names) for keyword argument
--- a/src/compile.js
+++ b/src/compile.js
@@ -1685,6 +1685,13 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
         }
     }
     else {
+        var res;
+        if (decos.length > 0) {
+            out("$ret = Sk.misceval.callsimOrSuspend(", scopename, ".$decorators[0], new Sk.builtins['function'](", scopename, ",$gbl", frees, "));");
+            this._checkSuspension();
+            return this._gr("funcobj", "$ret");
+        }
+
         return this._gr("funcobj", "new Sk.builtins['function'](", scopename, ",$gbl", frees, ")");
     }
 };
--- a/src/import.js
+++ b/src/import.js
@@ -238,8 +238,10 @@ Sk.doOneTimeInitialization = function () {
 
     // compile internal python files and add them to the __builtin__ module
     for (var file in Sk.internalPy.files) {
-        var mod = Sk.importBuiltinWithBody(file.split(".")[0], false, Sk.internalPy.files[file], false);
-        Sk.importStar(mod, Sk.globals, Sk.globals);
+        var fileWithoutExtension = file.split(".")[0].split('/')[1];
+        var mod = Sk.importBuiltinWithBody(fileWithoutExtension, false, Sk.internalPy.files[file], false);
+        goog.asserts.assert(mod["$d"][fileWithoutExtension] !== undefined, "Should have imported name " + fileWithoutExtension);
+        Sk.builtins[fileWithoutExtension] = mod["$d"][fileWithoutExtension];
     }
 };
 
--- a/src/import.js
+++ b/src/import.js
@@ -238,7 +238,7 @@ Sk.doOneTimeInitialization = function () {
 
     // compile internal python files and add them to the __builtin__ module
     for (var file in Sk.internalPy.files) {
-        var fileWithoutExtension = file.split(".")[0].split('/')[1];
+        var fileWithoutExtension = file.split(".")[0].split("/")[1];
         var mod = Sk.importBuiltinWithBody(fileWithoutExtension, false, Sk.internalPy.files[file], false);
         goog.asserts.assert(mod["$d"][fileWithoutExtension] !== undefined, "Should have imported name " + fileWithoutExtension);
         Sk.builtins[fileWithoutExtension] = mod["$d"][fileWithoutExtension];
--- a/src/import.js
+++ b/src/import.js
@@ -235,6 +235,12 @@ Sk.doOneTimeInitialization = function () {
 
         proto[name] = new Sk.builtin.func(proto[name]);
     }
+
+    // compile internal python files and add them to the __builtin__ module
+    for (var file in Sk.internalPy.files) {
+        var mod = Sk.importBuiltinWithBody(file.split(".")[0], false, Sk.internalPy.files[file], false);
+        Sk.importStar(mod, Sk.globals, Sk.globals);
+    }
 };
 
 /**
@@ -539,6 +545,21 @@ Sk.importMainWithBody = function (name, dumpJS, body, canSuspend) {
     return Sk.importModuleInternal_(name, dumpJS, "__main__", body, canSuspend);
 };
 
+/**
+ * **Run Python Code in Skulpt**
+ *
+ * When you want to hand Skulpt a string corresponding to a Python program this is the function.
+ *
+ * @param name {string}  File name to use for messages related to this run
+ * @param dumpJS {boolean} print out the compiled javascript
+ * @param body {string} Python Code
+ * @param canSuspend {boolean}  Use Suspensions for async execution
+ *
+ */
+Sk.importBuiltinWithBody = function (name, dumpJS, body, canSuspend) {
+    return Sk.importModuleInternal_(name, dumpJS, "__builtin__", body, canSuspend);
+};
+
 Sk.builtin.__import__ = function (name, globals, locals, fromlist) {
     // Save the Sk.globals variable importModuleInternal_ may replace it when it compiles
     // a Python language module.  for some reason, __name__ gets overwritten.
@@ -617,5 +638,6 @@ Sk.importStar = function (module, loc, global) {
 
 goog.exportSymbol("Sk.importMain", Sk.importMain);
 goog.exportSymbol("Sk.importMainWithBody", Sk.importMainWithBody);
+goog.exportSymbol("Sk.importBuiltinWithBody", Sk.importBuiltinWithBody);
 goog.exportSymbol("Sk.builtin.__import__", Sk.builtin.__import__);
 goog.exportSymbol("Sk.importStar", Sk.importStar);
--- a/src/misceval.js
+++ b/src/misceval.js
@@ -1204,6 +1204,14 @@ Sk.misceval.applyOrSuspend = function (func, kwdict, varargseq, kws, args) {
         //append kw args to args, filling in the default value where none is provided.
         return func.apply(null, args);
     } else {
+        fcall = func.__get__;
+        if (fcall !== undefined) {
+            fcall = Sk.misceval.callsimOrSuspend(func.__get__, func, func);
+            //args.unshift(func);
+            if (fcall.tp$call) {
+                return fcall.tp$call.call(fcall, args);
+            }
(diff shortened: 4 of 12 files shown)


src/abstract.js
```diff
@@ -828,20 +828,24 @@ Sk.abstr.gattr = function (obj, nameJS, canSuspend) {
         ret = Sk.misceval.callsimOrSuspend(f, new Sk.builtin.str(nameJS));
     }
 
-    ret = Sk.misceval.chain(ret, function(ret) {
+    var eventualresult = Sk.misceval.chain(undefined, function () {
+        return obj;
```
albertjan: @meredydd I've also changed this because the obj is a suspension aswell. But it may be that I should have done it differently, because it sortof feels like the same kind of issue as I mentioned in the PR.

src/lib/image.js
```diff
@@ -117,13 +117,37 @@ $builtinmodule = function (name) {
...
+                index = (y * 4) * self.width + (x * 4);
+                red = self.imagedata.data[index];
+                green = self.imagedata.data[index + 1];
+                blue = self.imagedata.data[index + 2];
+                arr[i] = new Sk.builtin.tuple([new Sk.builtin.int_(red), new Sk.builtin.int_(green), new Sk.builtin.int_(blue)]);
+            }
+
+            return new Sk.builtin.list(arr);
+        });
+
```
meredydd: Why are these changes in here? Not sure what images have to do with decorators...

albertjan: yeah they need to be gone 😄 I'll remove them

src/misceval.js
```diff
@@ -1189,6 +1189,14 @@ Sk.misceval.applyOrSuspend = function (func, kwdict, varargseq, kws, args) {
         //append kw args to args, filling in the default value where none is provided.
         return func.apply(null, args);
     } else {
+        fcall = func.__get__;
+        if (fcall !== undefined) {
+            fcall = Sk.misceval.callsimOrSuspend(func.__get__, func, func);
+            //args.unshift(func);
+            if (fcall.tp$call) {
```
meredydd: This will fail if `func.__get__` returns a suspension. What are you trying to do here?

src/object.js
```diff
@@ -88,8 +102,28 @@ Sk.builtin.object.prototype.GenericSetAttr = function (name, value) {
...
+        // f = descr.ob$type.tp$descr_set;
+        if (!(f) && descr["__set__"]) {
+            f = descr["__set__"];
+            Sk.misceval.callsim(f, descr, this, value);
+            return;
+        }
+        // todo;
+        //if (f && descr.tp$descr_set) // is a data descriptor if it has a set
+        //return f.call(descr, this, this.ob$type);
+    }
```
meredydd: What's `descr` short for here? I'm lost...

src/property.py
```diff
@@ -0,0 +1,40 @@
...
+        self.fdel(obj)
+
+    def getter(self, fget):
+        return type(self)(fget, self.fset, self.fdel, self.__doc__)
+
+    def setter(self, fset):
+        return type(self)(self.fget, fset, self.fdel, self.__doc__)
+
+    def deleter(self, fdel):
+        return type(self)(self.fget, self.fset, fdel, self.__doc__)
```
meredydd: Ah, so you are implementing property descriptors, then decorators to generate them? I'll keep hunting, so don't do anything yet, but it might help to break this into two PRs if that wouldn't be too much of a pain.

src/misceval.js
albertjan: Well the problem we're having here is that `func` itself is a suspension, so it will throw an exception even before it gets there. this call was just a shot in the dark. Thats why I asked for your help 😄.

src/object.js
albertjan: descriptor

albertjan: they should probably have been a part of a different PR.

src/property.py
albertjan: I agree.

src/misceval.js
meredydd: So, why is `func` a suspension? I'm guessing that it's because it's decorated, and the decorator suspended?

If this happens, we shouldn't continue and make the value of the function variable a suspension. What should happen is that the decorator should be allowed to complete before that variable gets assigned. Can you show me where the `compile.js` output for decorators is?

src/compile.js
```diff
@@ -1687,7 +1690,14 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
         }
     }
     else {
-        return this._gr("funcobj", "new Sk.builtins['function'](", scopename, ",$gbl", frees, ")");
+        var res;
+        if (decos.length > 0)
+        {
+            res = this._gr("funcobj", "Sk.misceval.callsim(", scopename, ".$decorators[0], new Sk.builtins['function'](", scopename, ",$gbl", frees, "))"); // scopename, ".$decorators[0](new Sk.builtins['function'](", scopename, ",$gbl", frees, "))";
```
albertjan: here the decorator is called.

src/compile.js
```diff
@@ -1687,7 +1690,14 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
         }
     }
     else {
-        return this._gr("funcobj", "new Sk.builtins['function'](", scopename, ",$gbl", frees, ")");
+        var res;
+        if (decos.length > 0)
+        {
+            res = this._gr("funcobj", "Sk.misceval.callsimOrSuspend(", scopename, ".$decorators[0], new Sk.builtins['function'](", scopename, ",$gbl", frees, "))"); // scopename, ".$decorators[0](new Sk.builtins['function'](", scopename, ",$gbl", frees, "))";
```
meredydd: Ah, here. We should do this in two stages:
1.  `out("$ret = Sk.misceval.callsimOrSuspend(", scopename, ".$decorators[0], new Sk.builtins['function'](", scopename, ",$gbl", frees, "))");`

2. `checkSuspension()`

3. return this._gr("funcobj", "$ret")

albertjan:
@meredydd that appears to have fixed the code example. 😄

src/abstract.js
albertjan: I can revert this change now can't I?

albertjan:
and the tests @meredydd you're awesome :)

src/object.js
meredydd: Can you separate them? It would certainly be easier to review, although I appreciate it's a pain in the ass.

src/property.py
meredydd: Brill. Am happy to commit to reviewing both.

bnmnetp:
So, the current state is that the tests all pass in the non-closure compiled phase.  But after running the closure compiler, the unittests appear to get hung up in test_complex -- Travis kills them after 10 minutes.  They've been running for 15 on my macbook pro and still stuck in test_complex.

If I take test_complex out of the picture then all tests except for: `Tests failed in  TestGenericStringIO.test_input` pass.

Any ideas?

bnmnetp:
Just to update, test_complex does eventually finish after about 35 minutes.  But for some strange reason, test_input still fails.

waywaaard:
Testing complex takes ages with suspensions turned on for every line. Maybe I need to do some profiling to find the bottleneck.

bnmnetp:
Indeed, If I change `debug_mode` back to false then everything passes in a reasonable amount of time.

It seems to me that this would be a separate issue then.  Is it worth holding this one back over this or should we make a separate issue for the stringIO failing in the closure compiled phase with debug turned on?

rixner:
@bnmnetp, if you just check out master and do:

```
skulpt.py rundebug test/unit/test_StringIO.py
```

There is a failure.  So, I would argue that that problem should be investigated independently and has nothing to do with this pull request.  (I didn't look at the rest of this pull request, but just thought I'd point that out.)  Although, you should probably do this yourself and confirm that it's the same failure you are seeing...

albertjan:
Shall I turn debug_mode off then? And open a PR with that on, so we can add the changes to fix those issues to there?

bnmnetp:
@albertjan that would be my vote.  I would like to see this go in in the hope that we can get some implementation of property working next.

@rixner, I did that and saw the same failure.

albertjan:
@bnmnetp I've found the bug in input, and commented out the test in debug_mode but we really should run them. I'll try recompiling v8 the later versions are much faster.

bnmnetp:
@albertjan I agree they should be run, if they can be run fast enough to avoid having all our travis checks look like they have failed.

albertjan:
I agree, I'll look into building a newer version of v8 or using node to run skulpt (as node is just v8 with some stuff boulted on to it).

bnmnetp:
Yeah, I've often thought that it would be convenient if we could use node.

bnmnetp:
OK, @albertjan WOW!
So property is implemented in property.py and it just works!  😄 

We should probably remove the function definition with the unimplemented error from builtins.js ??

albertjan:
Ah good catch yeah we can remove that. Yes it "just works" any .py files placed into the src dir will be compiled during startup and assigned to `Sk.builtin.<name_of_file>`. I know we might want to change that behaviour in the future to assign to different things. But we can do that as and when needed.

src/compile.js
```diff
@@ -1687,7 +1690,16 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
         }
     }
     else {
-        return this._gr("funcobj", "new Sk.builtins['function'](", scopename, ",$gbl", frees, ")");
+        var res;
+        if (decos.length > 0)
+        {
+            out("$ret = Sk.misceval.callsimOrSuspend(", scopename, ".$decorators[0], new Sk.builtins['function'](", scopename, ",$gbl", frees, "));");
```
meredydd: Am I missing something obvious, or does this only support one decorator at a time?

src/misceval.js
meredydd: I'm either misreading Github's interface or this issue is still here - a call to `Sk.misceval.chain` should sort you out.

albertjan:
Hey @meredydd I only see the comment about only the first decorator, I think you are correct there.

rixner:
This is pretty cool, so I hate to be the bad guy and complain about this yet again, but this one really concerns me.  I'm seeing up to a 20% slowdown from this pull request on programs that don't even use decorators.  In my opinion, that's a _lot_.

bnmnetp:
@rixner My guess is that it is the descriptor protocol more than the decorators themselves.  Descriptors could potentially impact accessing any instance variable.  I'm somewhat sympathetic to the performance argument, but I think this PR opens up so much more of Python to us!  I think this one enables lots more pure python modules to be added to the library.

How we make the tradeoffs of new functionality against performance is one that we need more discussion around.

meredydd:
I feel like 20% is bigger than should be associated with a change like this. Can you point a profiler at it, @rixner ?

albertjan:
It's compiling the property decorator at start up. Thats probably why its slower.

rixner:
I strongly suspect the problem is the new, expensive call to ```typeLookup``` at the start of ```GenericGetAttr```.  ```GenericGetAttr``` is a performance critical function and this is going to slow it down every time.  I'll profile it and confirm.

rixner:
Yes, ```Sk.builtin.type.typeLookup``` has become the number 1 function in the profile.

rixner:
@albertjan, when I measure time, I actually measure the body of the code and ignore compilation/startup overheads.  So, while that may have also gotten slower, I am personally not so concerned about that.

albertjan:
@rixner alright, it's obviously part of the GenericGetAttr. I'll look into speeding it up.

albertjan:
Why don't we move the typelookup to after the normal lookup so it will only do that when there's no normal property on the object. That would help with performance. probably.

albertjan:
@rixner do you have an example program that shows the performance hit? So I can run it locally?

rixner:
The gist I referenced [here](https://github.com/skulpt/skulpt/pull/597#issuecomment-231252466) in the comments for #597 is a program that shows a big slowdown for me.

albertjan:
Ok I've moved the typeLookup to after the normal field lookup. My understanding of descriptors in python aren't good enough to see if we break anything with this. The Performance is now the same as the master. Best run on my machine is 5.73 for both this branch and the master.

bnmnetp:
Here is a simple example:

```
class RevealAccess(object):
    """A data descriptor that sets and returns values
       normally and prints a message logging their access.
    """

    def __init__(self, initval=None, name='var'):
        self.val = initval
        self.name = name

    def __get__(self, obj):
        print('Retrieving', self.name)
        return self.val

    def __set__(self, obj, val):
        print('Updating', self.name)
        self.val = val


class MyClass(object):
    x = RevealAccess(10, 'var "x"')
    y = 5

m = MyClass()
print(m.x)
print(m.y)
m.x = 20
print(m.x)
```

my output is:
```
('Retrieving', 'var "x"')
10
5
('Updating', 'var "x"')
('Retrieving', 'var "x"')
20
```

rixner:
@albertjan, I also now see roughly the same performance as master.  Thanks!

albertjan:
I've merged it with master now and I see a performance increase of 0.3s on average due to the fast chain path. I'm still having one issue with the script @bnmnetp posted. My output differs:

```
('Retrieving', 'var "x"')
10
5
('Updating', 'var "x"')
20

('Retrieving', 'var "x"')
10
5
('Updating', 'var "x"')
('Retrieving', 'var "x"')
20
```

bnmnetp:
@albertjan so you are missing the second retrieving print statement even though it appears to have updated properly. Hmm.

albertjan:
I have to move the setter above the normal property accessor. I'm still seeing the same performance.

bnmnetp:
@albertjan This is so great.  

Here's another little test program.  It was working earlier today but is broken on the latest.

```
class P:

    def __init__(self,x):
        self.x = x

    @property
    def x(self):
        return self.__x

    @x.setter
    def x(self, x):
        if x < 0:
            self.__x = 0
        elif x > 1000:
            self.__x = 1000
        else:
            self.__x = x

t = P(9)
print(t.x)
t.x = 100000
print(t.x)
```

I get ``UNCAUGHT EXCEPTION: ExternalError: TypeError: Cannot read property '0' of undefined on line 6``

Instead of
```
9
1000
```

bnmnetp:
Ah, I see that ``test/unit/test_decorators.py`` went missing or the unit tests would have caught that.

albertjan:
Right ... :) I'll see if I can find that file, and if I can fix the bug.

bnmnetp:
Hi @albertjan,

Any progress on this one?  Are we stuck with a tradeoff of a slow but working implementation, or fast but non-working?

albertjan:
@bnmnetp yeah pretty much, I'll add the tests back in, and return it to working state, see if we can optimise it in a different way.

albertjan:
@bnmnetp I accidentally removed some code from the compiler causing this problem we saw. Do you know where in the cpython test suite the decorators and descriptors are tested?

bnmnetp:
@albertjan At one point this PR contained the file test/unit/test_decorators.py

You should be able to recover it from the history.

albertjan:
yeah I have it 😄 thanks

albertjan:
I'm trying to get the `test_decorators.py` to work but it seems expertly crafted to show all skulpt's creavats. https://github.com/python/cpython/blob/2.7/Lib/test/test_decorators.py

bnmnetp:
If I'm reading my log correctly all of those tests were passing from a branch called trinketapp-decorators-2 at this point:  28e31ea039dff0e4c059d1221ceab9b81263f6ca

albertjan:
@bnmnetp they can't have been they use double decorators and the compile function in almost all of the tests. they use `func_name`. all of which aren't supported.  I've added your version of the tests back in. I've changed them a little to make sure they also at least compile on cpython.

albertjan:
Good find @bnmnetp I thought I'd lost them in the rebase. I don't know from which branch in the cpython repo I got them. I've added them next to the other tests in the repo, to be extra sure it works :smile:. If this all builds and @rixner is happy with the perf, i'd say it's ready to be merged :shipit:.

We should note that we currently support only one decorator.

rixner:
What is internalpython.js and the other seemingly associated stuff (like ```importBuiltinWithBody```) in here?  And what does that have to do with decorators?

albertjan:
internalpython.js is the file that contains the parts of the stdlib that are written in python, for now only property.py. importBuiltinWithBody compiles these parts.

rixner:
Perhaps I am missing something really obvious, but why is this necessary?  If you put property.py in the lib directory, it will be importable.  This already works, why do we need a second mechanism to do the same thing?

albertjan:
`property` is a builtin like `map` and `filter` they don't need to imported. Thats why we need to do it when we initialise skulpt.

albertjan:
they aren't imported into `__main__` either

rixner:
Ah, sorry!  I was missing something really obvious. :smile:

bnmnetp:
I just merged this against the current master, and ran all the tests. While I did not measure the performance impact, it was able to take into account the gains from #604 and so seemed quite speedy.

albertjan:
😄 thats good, we do take a small hit during initialisation, that we may want to move forward to compile time with #609.

bnmnetp:
Are we all happy (enough) with this one to merge it?

bnmnetp:
A bit more empirical, I just timed ``./skulpt.py rununits``

This PR
```
Passed:   292 Failed     0
       29.49 real        29.01 user         1.50 sys
```

master
```
Passed:   289 Failed     0
       25.58 real        25.14 user         1.44 sys
```

I think we should merge this to get the feature into the language, but this gives us more to work on in terms of looking for optimizations.

albertjan:
there are some optimisations on the main branch that aren't on this one so the numbers may be a bit skewed by that.

bnmnetp:
OK, unless I hear otherwise, I'm going to merge this later today.

I think we've identified `GenericGetAttr` as an important starting point for looking at Optimizations, but the descriptor protocol is an integral part of that and it would make no sense to seriously look at Optimizing that without this feature in place.

rixner:
I still think this is a large slowdown.

I concede, however, that I care more about not losing any further performance than I do about this particular feature.  If others feel differently, then so be it.

albertjan:
@rixner let me do a merge with master so we can objectivly compare the perf difference.

albertjan:
this branch:

rununits:
```sh
./m rununits  25.04s user 1.32s system 106% cpu 24.841 total
./m rununits  25.10s user 1.28s system 106% cpu 24.795 total
./m rununits  25.46s user 1.27s system 106% cpu 25.140 total
```
field.py
```sh
time elapsed: 5.60977315903
./m run field.py  6.40s user 0.13s system 100% cpu 6.478 total

time elapsed: 5.49832081795
./m run field.py  6.30s user 0.12s system 100% cpu 6.380 total

time elapsed: 5.55619883537
./m run field.py  6.37s user 0.13s system 100% cpu 6.454 total
```

master:

```sh
./m rununits  24.33s user 1.21s system 106% cpu 24.019 total
./m rununits  23.75s user 1.20s system 106% cpu 23.472 total
./m rununits  24.02s user 1.19s system 106% cpu 23.705 total
```

field.py:
```sh
time elapsed: 5.57643413544
./m run field.py  6.32s user 0.12s system 100% cpu 6.406 total

time elapsed: 5.67489504814
./m run field.py  6.46s user 0.12s system 100% cpu 6.533 total

time elapsed: 5.81251382828
./m run field.py  6.57s user 0.12s system 100% cpu 6.647 total
```

bnmnetp:
Wow, thanks @albertjan !  I'll bet even @rixner can live with that 😄

src/import.js
```diff
@@ -539,6 +547,21 @@ Sk.importMainWithBody = function (name, dumpJS, body, canSuspend) {
     return Sk.importModuleInternal_(name, dumpJS, "__main__", body, canSuspend);
 };
 
+/**
+ * **Run Python Code in Skulpt**
+ *
+ * When you want to hand Skulpt a string corresponding to a Python program this is the function.
```
LeszekSwirski: I thought this was the job of importMainWithBody?

rixner:
Yes, I can live with that.  Clearly, the overhead of this pull request is now quite low!

src/import.js
rixner: @albertjan, since both @LeszekSwirski and I were confused by this, I think it would help us and all future developers if this comment were improved to more clearly explain the purpose of this function and its relation to the other import functions...  The current comment is misleading, in my opinion.

PR #594, opened 2016-07-04
remapToPy() maps null to 'None' rather than '{}'

meredydd:
Currently, `remapToPy()` turns `null` into an empty dictionary (because `typeof (null) == "object"`). Make it map to `None` instead.


rixner:
This seems like a trivial change.  Unless anyone objects today I'll merge it later.

albertjan:
Merge away

PR #595, opened 2016-07-05
Better/More consistent handling of feedback.

bnmnetp:


rixner:
I think this still prints things backwards.  Quick example:

```
import unittest

class Tests(unittest.TestCase):
    def test_truth(self):
        self.assertTrue(False)

if __name__ == "__main__":
    unittest.main()
```

This outputs:

```
Fail: expected False got True Expected False to be True
```

I think it should print "Fail: expected True got False...", as currently, the two clauses of the output seem to contradict each other.

bnmnetp:
@rixner, I had actually meant to remove the redundant clause altogether.  I've done that now, and your example would simply read

``Fail:  Expected False to be True``

rixner:
@bnmnetp, this looks much better to me.   As far as I'm concerned, this is a welcome improvement!

PR #597, opened 2016-07-07
Normalise implicit calls

meredydd:
There are many implicit calls (to functions like `__len__`) triggered from internal 'dunder-named' function calls (like `tp$length`). This PR starts the work of normalising those calls - it uses `Sk.abstr.gattr()` to get the attributes (so they can be dynamically returned by `__getattr__` and friends), and makes some (such as `__call__` and `__len__`) suspendable.

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -243,11 +243,11 @@ Sk.builtin.len = function len (item) {
     Sk.builtin.pyCheckArgs("len", arguments, 1, 1);
 
     if (item.sq$length) {
-        return Sk.misceval.chain(item.sq$length(true), Sk.builtin.int_);
+        return Sk.misceval.chain(item.sq$length(), Sk.builtin.int_);
     }
 
     if (item.mp$length) {
-        return Sk.misceval.chain(item.mp$length(true), Sk.builtin.int_);
+        return Sk.misceval.chain(item.mp$length(), Sk.builtin.int_);
     }
 
     if (item.tp$length) {
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -242,16 +242,18 @@ Sk.builtin.round = function round (number, ndigits) {
 Sk.builtin.len = function len (item) {
     Sk.builtin.pyCheckArgs("len", arguments, 1, 1);
 
+    var int_ = function(i) { return new Sk.builtin.int_(i); };
+
     if (item.sq$length) {
-        return Sk.misceval.chain(item.sq$length(), Sk.builtin.int_);
+        return Sk.misceval.chain(item.sq$length(), int_);
     }
 
     if (item.mp$length) {
-        return Sk.misceval.chain(item.mp$length(), Sk.builtin.int_);
+        return Sk.misceval.chain(item.mp$length(), int_);
     }
 
     if (item.tp$length) {
-        return Sk.misceval.chain(item.tp$length(true), Sk.builtin.int_);
+        return Sk.misceval.chain(item.tp$length(true), int_);
     }
 
     throw new Sk.builtin.TypeError("object of type '" + Sk.abstr.typeName(item) + "' has no len()");
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -243,15 +243,15 @@ Sk.builtin.len = function len (item) {
     Sk.builtin.pyCheckArgs("len", arguments, 1, 1);
 
     if (item.sq$length) {
-        return new Sk.builtin.int_(item.sq$length());
+        return Sk.misceval.chain(item.sq$length(true), Sk.builtin.int_);
     }
 
     if (item.mp$length) {
-        return new Sk.builtin.int_(item.mp$length());
+        return Sk.misceval.chain(item.mp$length(true), Sk.builtin.int_);
     }
 
     if (item.tp$length) {
-        return new Sk.builtin.int_(item.tp$length());
+        return Sk.misceval.chain(item.tp$length(true), Sk.builtin.int_);
     }
 
     throw new Sk.builtin.TypeError("object of type '" + Sk.abstr.typeName(item) + "' has no len()");
--- a/src/misceval.js
+++ b/src/misceval.js
@@ -989,28 +989,43 @@ goog.exportSymbol("Sk.misceval.applyAsync", Sk.misceval.applyAsync);
  */
 
 Sk.misceval.chain = function (initialValue, chainedFns) {
-    // as per the discussion here: https://github.com/skulpt/skulpt/pull/552
-    // this is here for performance reasons. array.slice doesn't get optimized
-    var fs = new Array(arguments.length), i = 1;
+    // We try to minimse overhead when nothing suspends (the common case)
+    var i = 1, value = initialValue, j, fs;
 
-    for (i = 1; i < arguments.length; i++) {
-        fs[i] = arguments[i];
+    while (true) {
+        if (i == arguments.length) {
+            return value;
+        }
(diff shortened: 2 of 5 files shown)


rixner:
Is it really necessary to make simple things like ```len``` suspendable?  Introducing the call to ```chain``` in that path is going to really slow down ```len``` in the common case where something like ```sq$length``` is just going to immediately return a value.  I would much prefer if we could find a mechanism to only introduce that overhead when you are ultimately calling a user defined ```__len__``` function.  Is that possible?

I have the same concern with most of the dunder methods which are calling internal functions that don't need to suspend.

meredydd:
> Is it really necessary to make simple things like `len` suspendable?

Yes, for two reasons. For one thing, it turns up in real code (in the case that motivated this patch, it was a lazy-loaded collection). For another, suspensions are how Skulpt is implementing interactive debugging. So your question is equivalent to, "Should you be able to debug your own `__len__` functions?" - to which I'd say the answer is obvious.


> I would much prefer if we could find a mechanism to only introduce that overhead when you are ultimately calling a user defined `__len__` function. Is that possible?
Most of this PR only affects the call paths for user-defined `__len__` functions. I take it therefore that you are objecting to the modifications to `Sk.builtin.len()`?

The problem here is that the dunder methods' API allows them to return primitive numbers, so you need to `int()` them before they go back to Python. At the point of calling `sq$length`, we don't know whether a given object is user-defined or not. So it *might* return a `Suspension`. So we have to either check whether the object is user-defined, or whether the return value is a suspension - and the former is at least as expensive as the latter. (Testing for suspensions is just an attribute check these days.)

We could do the check immediately in `Sk.builtin.len`, rather than calling out to `Sk.misceval.chain()`. But in practice, doing this everywhere turned into an enormous amount of boilerplate, which is why I factored out that check-and-call patterin into `chain()` in the first place.

(By the way, `chain()` is nowhere near as heavyweight as I suspect you're fearing. It's not a `Promise` - it just checks whether its first argument is a `Suspension`, and if it's not, calls the first function immediately.

Which makes me think - would you be happer if I tweaked `chain()` to be even faster in the fast case, and made it only drop into the `nextStep()` lambda loop when it encounters its first `Suspension`? If the fast path through `chain` were something like:
```javascript
function chain(value, chainedFns) {
  for (var i=1; i < arguments.length; i++) {
    if (value.isSuspension) { break; } // uh-oh, slow case.
    value = arguments[i](value);
  }
  if (i == arguments.length) { return value; }

  // ... existing implementation continues here ...
```
...would you be more relaxed about patches like this?)

meredydd:
Oh, and I'll send up the bat-signal for @LeszekSwirski, because he's done a bunch of thinking in this neighbourhood and may see an option I've missed.

rixner:
I understand (and appreciate) what you are trying to accomplish. This is why I asked if it is possible to only introduce the overhead when you ultimately call into Python.  If the answer is no, then so be it.

I definitely appreciate all of the activity in Skulpt and the contributions that you and others have made to make it so much better.  However, many of these improvements have introduced overheads, and while each of these things is just a small overhead, a lot of small overheads add up.  I have definitely seen chain, for instance, show up high in profiles some times.  It gets called an awful lot.

As a frame of reference for where I am coming from, consider the following.  I use this [program](https://gist.github.com/rixner/f79dfa52a0385e5c82d147ae3d06c50c), which was written by one of the students in one of my classes years ago, as one of many performance benchmarks.  (The GUI part of the program is stripped out, leaving just the computation.)  Looking back in the repo, I see the following results (arbitrarily picked versions that seemed to be stable points):

816a848 (July 6, 2016): 9.5 seconds
e55c36e (May 6, 2015): 8.5 seconds
63c7239 (August 4, 2014): 4.3 seconds

For reference, this program runs in 0.13 seconds natively with CPython on the machine I ran these tests on.  (I do not expect Skulpt to get that performance, I include that for reference to understand the speed of the machine I used for the tests.)

This is neither the highest nor the lowest slowdown.  It all depends on what language features get exercised most, obviously.

I consider over a 2x slowdown to be pretty significant.  This is why I am wary of even small performance degradations since this overall slowdown comes from the aggregate effect of a bunch of seemingly small slowdowns.

(Yes, I do like the idea of the fast path in chain. :smile: )

meredydd:
Your wish is my command ;)

I've added a fast path to `chain`. (Pretty much what I sketched above, except that `for` is now a `while` to make it explicit that there is only ever one path out of that loop that isn't a return).

Overall, I do agree that a 2x performance degradation is worth raising the alarm over. I'm a bit slammed at the moment, but I'd definitely be up for running that program through a profiler and seeing if there are some easy wins.

(I have a terrible sinking feeling the big killer is going to be that using `try`/`catch` in a function kills the V8 optimiser for that function. This means that *none* of our generated Python gets optimised. I did make a couple of attempts, a few months ago, to extract the catch blocks from the main loop, but then you end up deciding how/whether to persist each temporary variable, and you're basically suspending on every function, which means doing liveness analysis, and you're basically reimplementing an optimiser again!)

rixner:
Thanks!

I have also tried a few times to deal with the try/catch over the past several months without success.  I did try to create a function roughly like this:

```
function tryCatch(body, error) {
   try {
       body();
   catch (x) {
       error(x);
  }
}
```

Then, I was passing in the while loop in a lambda as the body function and the catch clause in a lambda as the error function.  It's more complicated than this, since sometimes after an exception we just resume the loop, but you get the idea.  This did allow the compiled Python code to be optimized without needing to try to figure out how to deal with the temporary variables, but it didn't result in any real performance improvement.  The best that I could determine is that the fact that the body was now accessing everything in a closure instead of as local variables was negating the benefits.  So, I'm really not sure what the right approach is...

Honestly, at this point, I'd be happy if we could keep performance stable...

waywaaard:
Here are some hints about the try-catch optimization: https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#2-unsupported-syntax

rixner:
@waywaaard, that's basically the method I tried.  However, because of the way saving suspensions and waking from suspensions currently work, it's very difficult to actually do it without a closure, which is what I think is hurting the performance of that method.

meredydd:
It's not just suspensions; you've got to keep all your local variables around in case they get accessed after the `catch` (or do liveness analysis to figure out which ones to keep around, and only keep a subset...). This means the locals would have to go "off-stack" into an object on the heap in any case - at which point, that object is that thing you need when suspending anyway.

rixner:
@meredydd, that's basically the problem I had.  You need everything saved so that you can resume after an exception that allows the loop to continue and/or after a suspension.  Accessing it all in a closure proved to be expensive.  I haven't thought about whether there is a better way to share the data among the body, the catch, and the save/wake functions.  As you point out, your approach sounds problematic for other reasons.

meredydd:
In any case, the first order of business is to profile the darn thing and validate our hypotheses.
Shall we separate out this discussion into a separate performance thread somewhere (perhaps an issue?) It's getting a bit big for this pull request...

LeszekSwirski:
I recommend #295 as a good place for discussion ;)

albertjan:
Seems like it's ready to be merged to me with @meredydd 's last change.

albertjan:
@meredydd I fear I have caused a conflict. 

@skulpt/committers are we in agreement, can this be pulled in?

meredydd:
Now rebased up (subsequent changes in the additional two commits for easy reviewing), and tests passing on my end. If the build passes, I'm good to go.

Issue #591, opened 2016-06-20
Name conflicts with suspensions


rixner:
Currently, there are several internal variable names that are being used that don't have $'s in them.  This leads to name conflicts when you have Python variables of the same name.  This is a problem in particular with ```$saveSuspension```.  Take, for example, the following program:

```
def f():
    child = 1
    filename = "abc"
    lineno = 2
    colno = 3
    normal = 5

    print child, filename, lineno, colno, normal

f()
```

If you use "skulpt.py run" to run this program, it will print the correct values.  If you suspend, though, using "skulpt.py rundebug" you get the wrong results.  I don't know how many other names are problematic, but I don't think these four (child, filename, lineno, and colno) are the only problems.  The solution is probably "easy", in that all these internal names just need to be prefixed with a $ to become $child, $filename, etc...  There are just potentially a lot of places which need to be fixed.

meredydd:
Another candidate: `isSuspension` should become `$isSuspension`.

PR #598, opened 2016-07-07
More builtins

rixner:
This provides implementations of the ```callable```, ```next```, and ```iter``` builtins.

### DIFF
--- a/src/dict.js
+++ b/src/dict.js
@@ -178,46 +178,6 @@ Sk.builtin.dict.prototype.mp$del_subscript = function (key) {
     throw new Sk.builtin.KeyError(s.v);
 };
 
-Sk.builtin.dict.prototype.tp$iter = function () {
-    var ret;
-    var i;
-    var bucket;
-    var k;
-    var allkeys = [];
-    for (k in this) {
-        if (this.hasOwnProperty(k)) {
-            bucket = this[k];
-            if (bucket && bucket.$hash !== undefined && bucket.items !== undefined) {
-                // skip internal stuff. todo; merge pyobj and this
-                for (i = 0; i < bucket.items.length; i++) {
-                    allkeys.push(bucket.items[i].lhs);
-                }
-            }
-        }
-    }
-    //print(allkeys);
-
-    ret =
-    {
-        tp$iter    : function () {
-            return ret;
-        },
-        $obj       : this,
-        $index     : 0,
-        $keys      : allkeys,
-        tp$iternext: function () {
-            // todo; StopIteration
-            if (ret.$index >= ret.$keys.length) {
-                return undefined;
-            }
-            return ret.$keys[ret.$index++];
-            // return ret.$obj[ret.$keys[ret.$index++]].lhs;
-        },
-        tp$name    : "dict_keyiterator"
-    };
-    return ret;
-};
-
 Sk.builtin.dict.prototype["$r"] = function () {
     var v;
     var iter, k;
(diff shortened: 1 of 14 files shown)


albertjan:
Wow this is one very cool PR! I say merge!

albertjan:
@skulpt/committers merge at will forgot to mention in my earlier comment

src/dict.js
```diff
@@ -578,3 +542,58 @@ Sk.builtin.dict.prototype["viewvalues"] = new Sk.builtin.func(function (self) {
...
+                for (i = 0; i < bucket.items.length; i++) {
+                    allkeys.push(bucket.items[i].lhs);
+                }
+            }
+        }
+    }
+    this.$keys = allkeys;
+    this.tp$iter = this;
+    this.tp$iternext = function () {
+        // todo; StopIteration
```
meredydd: Don't think this is a TODO any more. `StopIteration` is handled by the dunder-to-Python adapters, and converted into `return undefined`, which is a much more efficient way of doing things in Javascript.

src/function.js
```diff
@@ -93,12 +93,24 @@ Sk.builtin.checkIterable = function (arg) {
...
+    // takes care of python function, methods and lambdas
+    if (obj instanceof Sk.builtin.func) {
+        return true;
+    }
+    // takes care of instances of methods
+    if (obj instanceof Sk.builtin.method) {
+        return true;
+    }
+    // go up the prototype chain to see if the class has a __call__ method
+    if (Sk.abstr.lookupSpecial(obj, "__call__") !== undefined) {
```
meredydd: This doesn't work in the corner case where `__call__` is supplied by `__getattr__`. (That does work in real Python.)

It's a pretty corner-y case, and should probably be fixed all at once rather than in this PR, but I wanted to make a note.

src/iterator.js
```diff
@@ -0,0 +1,102 @@
...
+        return obj;
+    }
+    objit = Sk.abstr.lookupSpecial(obj, "__iter__");
+    if (objit) {
+        return Sk.misceval.callsim(objit, obj);
+    }
+    this.sentinel = sentinel;
+    this.flag = false;
+    this.idx = 0;
+    this.obj = obj;
```
meredydd: Non-dollar-prefixed names could leak JS objects into Python here.

src/iterator.js
```diff
@@ -0,0 +1,102 @@
...
+    if (sentinel === undefined) {
+        this.getitem = Sk.abstr.lookupSpecial(obj, "__getitem__");
```
meredydd: Again, doesn't work with `__getattr__`-supplied `__getitem__`; should probably be fixed at all call sites of `lookupSpecial()`.

rixner:
I don't think this is an issue.  Our internal objects all have a ```v``` property, for instance, that is not leaked to Python.  The properties that are accessible to Python are all in the ```$d``` dictionary/object, aren't they?  If you look at ```GenericGetAttr```, it will only find things in the ```$d``` property of the object or the ```$d``` property of the constructor.  It will not allow you to see properties of the object itself.

Just to be sure, we ran some tests where we tried to read/set these properties from a Python program and were unable to do so.

The issue in #591 had to do with the fact that variable inserted into the compiled code didn't have $'s, so they could shadow Python variable names.  I don't think that applies here.

Obviously, we can just prefix them all to be extra safe, but I don't think it's necessary.

meredydd:
1) OK, good to know. I don't have sufficient understanding of that bit of the codebase, and had recalled something about looking up bare names on class objects, but am happy to accept your reassurance.

2) Absolutely - this should be fixed in `lookupSpecial` rather than this PR. Just wanted to have raised the issue.

Merging now.

PR #600, opened 2016-07-13
Rename ParseError to SyntaxError

meredydd:
One for discussion, I think!

Some user testing recently revealed that people were left scratching their heads over `ParseError`. It doesn't Google well, because Python calls it a `SyntaxError` instead. This PR proposes to rename it to `SyntaxError` to match other Python implementations.

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -267,16 +267,16 @@ Sk.abstr.setUpInheritance("OverflowError", Sk.builtin.OverflowError, Sk.builtin.
  * @extends Sk.builtin.StandardError
  * @param {...*} args
  */
-Sk.builtin.ParseError = function (args) {
+Sk.builtin.SyntaxError = function (args) {
     var o;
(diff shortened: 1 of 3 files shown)


rixner:
I'm all for this change.  Unless I'm misreading the diff, though, I think you now have ```Sk.builtin.SyntaxError``` defined twice, as it was already there.  Currently, ```SyntaxError``` is being used for errors that occur after parsing and ```ParseError`` is being used for errors that occur in parser.js.  

There's also a ```TokenError``` that I think is also just a ```SyntaxError```.  Perhaps that should be removed as well?

meredydd:
Yeah, that's what's making the tests fail. I'm happy to respin by removing TokenError as well - just wanted to take people's temperature on whether this was a welcome change before I got stuck in :)

rixner:
As I said, I'm for it. :smile:

meredydd:
Now ready to go, I think.

albertjan:
Merging then :smile:

PR #604, opened 2016-07-25
Minor v8 speedups

LeszekSwirski:
Trying to be a bit more compiler/v8 friendly. Timed in a browser with @rixner's [script](https://gist.github.com/rixner/f79dfa52a0385e5c82d147ae3d06c50c), modified to time each `update()` call separately instead of all of them together, and to aggregate these timings.

Timings before:
```
min time: 0.0622251033783
average time: 0.0676951956749
stddev: 0.00581978406985
```

Timings after:
```
min time: 0.0563948154449
average time: 0.0603578090668
stddev: 0.00456230040262
```

### DIFF
--- a/src/misceval.js
+++ b/src/misceval.js
@@ -211,6 +211,15 @@ Sk.misceval.swappedOp_ = {
     "NotIn": "In_"
 };
 
+Sk.misceval._tryCatch = function (func, catchHandler) {
+    try {
+        return func();
+    } catch (err) {
+        return catchHandler(err);
+    }
+};
+goog.exportSymbol("Sk.misceval._tryCatch", Sk.misceval._tryCatch);
+
 /**
 * @param{*} v
 * @param{*} w
@@ -447,7 +456,7 @@ Sk.misceval.richCompareBool = function (v, w, op, canSuspend) {
 
     vcmp = Sk.abstr.lookupSpecial(v, "__cmp__");
     if (vcmp) {
-        try {
+        return Sk.misceval._tryCatch(function() {
             ret = Sk.misceval.callsim(vcmp, v, w);
             if (Sk.builtin.checkNumber(ret)) {
                 ret = Sk.builtin.asnum$(ret);
@@ -469,15 +478,15 @@ Sk.misceval.richCompareBool = function (v, w, op, canSuspend) {
             if (ret !== Sk.builtin.NotImplemented.NotImplemented$) {
                 throw new Sk.builtin.TypeError("comparison did not return an int");
             }
-        } catch (e) {
+        }, function (e) {
             throw new Sk.builtin.TypeError("comparison did not return an int");
-        }
+        });
     }
 
     wcmp = Sk.abstr.lookupSpecial(w, "__cmp__");
     if (wcmp) {
         // note, flipped on return value and call
-        try {
+        return Sk.misceval._tryCatch(function() {
             ret = Sk.misceval.callsim(wcmp, w, v);
             if (Sk.builtin.checkNumber(ret)) {
                 ret = Sk.builtin.asnum$(ret);
@@ -499,9 +508,9 @@ Sk.misceval.richCompareBool = function (v, w, op, canSuspend) {
             if (ret !== Sk.builtin.NotImplemented.NotImplemented$) {
                 throw new Sk.builtin.TypeError("comparison did not return an int");
             }
-        } catch (e) {
+        }, function (e) {
             throw new Sk.builtin.TypeError("comparison did not return an int");
-        }
+        });
     }
 
     // handle special cases for comparing None with None or Bool with Bool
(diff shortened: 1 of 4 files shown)


waywaaard:
Have you tried to bypass the `setTimeout(func, 0)` with other tricks for a browser `nextTick` trick? Maybe using something like: https://github.com/kriskowal/asap

The timeout of 0 is for me basically **4ms**. Would be cool if we can lower it.

I have been doing some experiments with the postMessage. Maybe we can also try `requestAnimationFrame`.

LeszekSwirski:
I did not, but I should note that that's an orthogonal improvement to do with scheduling, while these are improvements to do with compilability.

That said, the test code here wasn't suspendable, so I don't think setTimeout ever even got invoked. I might be wrong about that though.

LeszekSwirski:
Interestingly, I (manually) got rid of all the try-catches in the generated code, and the speed improvement wasn't that great. It went down to:

```
min time: 0.053530216217
average time: 0.0585943579674
stddev: 0.00682532023716
```

The chrome profiler suggests that the generated functions were now (with the manually remove try-catches) finally being optimised, so something else is the real bottleneck. GenericGetAttr and str seem to come up a lot...

waywaaard:
Have you done some profiling in order to identify the bottlenecks?

LeszekSwirski:
I've tried to, to some extent, which is where I got that str and GenericGetAttr come up a lot, but profiling lines within them is tough because they're compiled and optimised by v8. I should add that the numbers I have are without clojure compilation.

albertjan:
@LeszekSwirski a 10% speed up on average is pretty good if you ask me 😄. Is this ready for merging?

meredydd:
Me likey. Merging.

rixner:
@waywaaard, doesn't that 0-delay ```setTimeout``` only get called for yield suspensions for which there is no user provided handler?  So one can already turn off yield suspensions if you are not using them or provide a handler that does something better if you need it to, right?  I use a ```setImmediate``` polyfill, for instance, and use that in cases where I need it.  Furthermore, given that it uses ```Sk.setTimeout``` now, you could provide an implementation that first checks for a 0 delay and uses some implementation of ```setImmediate``` in that case.

Skulpt has traditionally functioned without the need for any external libraries or modern browser features.  I believe Promises are the only exception to that.  Introducing something like ```postMessage``` would be another step away from that.  Presumably, it would be done in a way that falls back to ```setTimeout```, but still...  I'm not arguing against starting to incorporate more modern features (in a way that continues to function on older browsers), but I am arguing that doing so at least deserves some discussion among the community first.  Especially in this case, where I believe there is enough configurability that you can solve it without changing the core of Skulpt, so there is not an immediate need to fix something.

meredydd:
FWIW, Promises are polyfilled in (as a policy decision), so we don't depend on having those either.

rixner:
@meredydd, I know, I was just saying that this is the only thing we polyfill in order to allow us to use a modern feature...

waywaaard:
@rixner Of course, it is configurable. But every additional configuration makes ist more difficult for new people to use skulpt. 

And if we can get a noticable performance improvement, then we should really consider it.

meredydd:
@waywaaard Where do we call `setTimeout(0)`, except when we're testing suspensions (or in our Promise polyfill)?

albertjan:
in misceval for `Sk.yield` is the only other place.

src/misceval.js
```diff
@@ -447,7 +456,7 @@ Sk.misceval.richCompareBool = function (v, w, op, canSuspend) {
 
     vcmp = Sk.abstr.lookupSpecial(v, "__cmp__");
     if (vcmp) {
-        try {
+        return Sk.misceval._tryCatch(function() {
```
meredydd: Ahhhh bug. The function  How did this pass the test suite?

meredydd:
This PR has a nasty bug; I'm rolling it back.

LeszekSwirski:
Whoops, I'll fix that later.

meredydd:
I've just reverted d975afd564b0df2f0427603598f4af33eeda01e4, as that was the culprit. (Props for nicely separate commits that made that easy.)

To expand more, there were _conditional_ `return` statements inside that `try{}catch(){}` you turned into an `Sk.misceval._tryCatch`. If none of those `if`s matched, control flow was meant to carry on, but instead that commit caused us to return `undefined`.

LeszekSwirski:
Yeah, my bad, I'd assumed that once you get past if (vcmp), you're expected to return or throw. Do you have example code that fails that we can add to the tests?

meredydd:
It blew up when comparing a `dict` to `None` (although the context was several stack frames deep and triggered somewhat implicitly). I'm done for the day, so I suspect you'll be able to try replicating that sooner than I will.

PR #606, opened 2016-07-26
Fix suspension resumption after importing Python modules

DIJamner:
Fixes suspension resumption via $saveSuspension after importing an external Python module. Both the main script and the external module could be compiled to code assigning to "$scope0". As a result, when the main script imported the external script, it would overwrite $scope0. Then, when the main script suspended, it would resume in the external library.

To fix this, we simply name the generated function the same as the var. Thus, references within the function body are guaranteed to refer to it whether or not the global variable is overwritten.

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -2237,7 +2237,9 @@ Compiler.prototype.cmod = function (mod) {
     var modf = this.enterScope(new Sk.builtin.str("<module>"), mod, 0, this.canSuspend);
 
     var entryBlock = this.newBlock("module entry");
-    this.u.prefixCode = "var " + modf + "=(function($modname){";
+    // we name the function to guarantee that internal references to modf reference this function
+    // even if the global var is redefined
+    this.u.prefixCode = "var " + modf + "=(function " + modf + "($modname){";
     this.u.varDeclsCode =
         "var $gbl = {}, $blk=" + entryBlock +
         ",$exc=[],$loc=$gbl,$err=undefined;$gbl.__name__=$modname;$loc.__file__=new Sk.builtins.str('" + this.filename +


rixner:
This looks like you just added a name to an otherwise anonymous function.  I don't quite understand how this changes anything about ```$scope0```.  If everything is working correctly, an imported module should never get compiled to something named ```$scope0``` because ```Sk.gensymcount``` controls the number that is appended to the symbol name.  This is only reset to 0 via a call to ```Sk.resetCompiler```, which should only occur when you start a new program, not before importing an external module.

Can you provide a more concrete example of what is failing so that we can look at it further and better understand your proposed fix?  Thanks!

DIJamner:
After thinking about it for a bit, I think I see where my problem is. I'm currently using Skulpt on a project that compiles to a single .html file. This means that the "external libraries" I'm loading do not have their own locations. To get around the default behavior, I tried to preload my libraries into Sk.externalLibraryCache. However, since that only contains compiled JavaScript code, I was compiling my libraries separately. This led to the multiple $scope0s.

I just tried overriding Sk.loadExternalLibraryInternal_ in my project so that it looks up the files in a dictionary similar to Sk.builtinFiles.files and that solved my problem without need of the changes in this pull request.

It does worry me a bit though that if, for whatever reason, another piece of code were to overwrite $scopeX it might alter the behavior of my suspensions, which this pull request prevents. Do you think that should be a consideration?

Thanks for your assistance!

meredydd:
I'm in favour of this change.

@rixner - the issue is that if you compile two files separately, you could end up duplicating gensyms (ie, reusing a name like `$scope0` so it refers to different things in multiple compilations). Normally (I think) this isn't so much of a problem, as every symbol is defined (with a `var`) inside a JS function, so gensyms only need to be unique within a given compilation.

The exception is the outermost scope, which is defined into a shared namespace by the `eval()` in `Sk.importModuleInternal_`. This isn't a problem for `import` (it uses that var right away, so the fact that it might be redefined in a few milliseconds' time isn't a concern). However, it *is* a problem for the suspension code, which attaches attributes to the function object as a way of determining whether we are calling a function for the first time or resuming a suspension.

This PR fixes the issue by making sure that the function "knows its own name". If you define a recursive function like this:
```javascript
var f = function() { f(); }
```
...then to resolve `f` it looks up that var in the parent scope. If you change that var before you call the function, it won't recurse! Sample code:

```javascript
var f = function() { f(); }
var g = f;
f = function() { console.log("oops"); }

g();  // <-- This will print "oops"
```

However, if you tell the function its own name, it doesn't have to look up any vars in the outer scope, because it always knows:

```javascript
var f = function f() { f(); }
```

(If you run the above example with this as the first line, you'll get a stack overflow, because `f()` recurses as you'd expect it to).

Does this make sense?

@DIJamner  - As it happens, I'm noodling with ahead-of-time compilation right now (I have a PR cooking to add a `./m compile` command to the `skulpt.py` script). How are you currently performing your compilation?

rixner:
@meredydd, okay, I see.  This solves one limited part of the problem.  I have nothing against doing this, as it is simple and improves things.

However, I still don't believe this solves the underlying problem.  All of the ```$scopeXX``` symbols are at the top level.  The thing that is "unique" about ```$scope0``` is simply that it's guaranteed to conflict.  But, in reality, any of the ```$scopeXX``` symbols could conflict and this does nothing to address that.  You just might not see the problem, because the probability of a conflict is relatively low given that it depends on the number of symbols that get generated for each of the files.  When they do conflict, though, I'm reasonably sure that this will cause you other problems where you'll end up calling the wrong functions.

I think a real solution needs to change the name of the ```$scopeXX``` variables to something that's guaranteed to be unique.  Perhaps by including the module name?  Or by explicitly supporting ahead-of-time compilation by modifying ```gensym``` to be cognizant of it?

@DIJamner, if you don't actually need ahead-of-time compilation, you could ship your modules as Python, either by adding them to the libraries or by using ```Sk.read``` to return the appropriate Python file when it is imported.  It would then be compiled at that time.  Obviously, it would be nice if we supported ahead-of-time compilation regardless, though.

meredydd:
D'oh. Brainfart - thought they were defined inside each other. In which case, I think the correct response is to wrap everything in an anonymous function before it gets `eval`ed in `importModuleInternal_`, so they don't conflict.

meredydd:
(Or alternatively, to generate that wrapper function around the output in `Sk.compile`, which might be more sensible - put the containment in the same file that creates the problem :) )

DIJamner:
@meredydd I was using Sk.compile() to compile the Python and then storing it in SK.externalLibraryCache. It seemed a bit hackish, but Skulpt never overwrites the cache, so it worked.

@rixner I did switch to shipping Python modules more-or-less as you suggested. However, I also had to add this`Sk.loadExternalLibraryInternal_ = function (path, inject) { return Sk.read(path);  };`. There may be a better way to add file-less modules than my "all of the above" method, but when I was running through the debugger, it seemed to follow a variety of different control flows for importing my modules. I might look into this further.

As far as this pull request, @meredydd's fix in #610 is more general, so I'll close this one.

meredydd:
FYI, PR #609 would give us an official way of packaging arbitrary `.py` or `.js` files (with or without ahead-of-time compilation for the Python files) for something like your use case. Feel free to take a peek/comment; feedback welcome.

Issue #601, opened 2016-07-18
Python Import Module


dihego:
Hello,

I am in the attempt to import python module Exscript. So far all modules have been working out for me except this one. 
The module is available in my system and I can execute the same script thru CLI but not thru skulpt. 

Any thoughts...

waywaaard:
Hi,

skulpt is a Python Interpreter written in JavaScript and purely web-based. Therefore, a lot of complex standard libs/modules are missing. 

This is the difference between your local python installation and skulpt and why exscript does not work.

Issue #554, opened 2016-03-26
Importing simple python file into project


Daltron:
I am using the example on the Sculpt website (https://gist.github.com/bnmnetp/4650616). What I am wanting to do is import a simple Python class I can use, in this case Person.py 

I have already looked at #381 and am aware of the builtinRead(fileName) function but am not sure where to call it? 

If somebody could show me a step by step example, it would be greatly appreciated.

waywaaard:
It's a late response, but what you can do the following:

```javascript
var myCustomFiles = {'Person.py', 'print "hello"\nprint "test"'};

function builtinRead(x) {
    if (myCustomFiles[x] !== undefined) {
        return myCustomfiles[x];
    }

    if (Sk.builtinFiles === undefined || Sk.builtinFiles["files"][x] === undefined)
            throw "File not found: '" + x + "'";
    return Sk.builtinFiles["files"][x];
}
```

alexwenbj:
@ebertmi Hi,Michael,I custom my skulpt as you show:
`

var myCustomFiles = {'Person.py': 'print "hello"'};
--
  | window.builtinRead = builtinRead;
  | function builtinRead(x) {
  | if (myCustomFiles[x] !== undefined) {
  | return myCustomfiles[x];
  | }
  | if (Sk.builtinFiles === undefined \|\| Sk.builtinFiles["files"][x] === undefined)
  | throw "File not found: '" + x + "'";
  | return Sk.builtinFiles["files"][x];
  | }
`
but when I run the Python code:
`f = open("Person.py")`
It show me an error:
ExternalError: ReferenceError: myCustomfiles is not defined on line 1
Would you please tell me what is the problem?Thank you in advance.

PR #609, opened 2016-07-28
Support ahead-of-time compilation

meredydd:
Adds two commands: `compile` (which compiles a single .py file to a .js file that's ready to load), and `compileall` (compiles a directory of mixed .py/.js files into a JS file that adds its contents to `Sk.builtinFiles.files`, like `skulpt-stdlib.hs`).

This PR also adds a default `Sk.read()` implementation that uses `Sk.builtinFiles.files`, so as long as you've included `skulpt.js` and `skulpt-stdlib-js` (plus any additional JS files you've generated with `./m compileall`), module loading will work out-of-the-box.

This should help address complaints about how much of a pain it is to include extra Python modules in your Skulpt environment (eg issue #554, arguably #601, the underlying problem behind PR #606...)


rixner:
Should there be some way to check if the compile time options (such as ```debugging```, ```yieldLimit```, etc.) are the same as the runtime options when you use these compiled files?  And at least print some sort of warning?  Otherwise, if you mix aot and run-time compiled code, you might get unexpected results.  Or do you feel like this is a "use at your own risk" feature?  (In which case, perhaps that should be made really clear...)

rixner:
(I'm also not totally convinced that ahead-of-time compilation is the solution to the "pain" of including extra Python files.  Arguably, that's a deficiency of the documentation.  If you drop them in the ```lib``` folder, for instance, it "just works" - assuming, of course, that Skulpt supports the features that they are trying to use.  Just an aside, not an argument against the pull request.)

albertjan:
It may be something people are looking for, it doesn't break anything in our default workflow does it?

meredydd:
@rixner - absolutely. Ahead-of-time compilation is not the solution to the pain of bundling extra python files - `./m combineall` is.

(And it is a pain - right now the only way to add them to the build is to either copy them into the Skulpt source tree(!!), or roll your own `Sk.builtin.read()` implementation - which is what Anvil does, but it's a pain for new users.)

Should I separate out AOT compilation from `combineall`? I agree AOT has issues with build options. (Also, it creates really big `.js` files)

PR #610, opened 2016-07-28
Wrap compiler output in a function to avoid scope gensyms colliding

meredydd:
This should provide a more universal fix to the problem identified in #606 

@rixner, @DIJamner for review please


DIJamner:
This fixes the issue I had and is definitely more general than my solution. Looks good to me!

albertjan:
@rixner I would like your opinion on this one, it looks simple enough to me. Are we going to break old implementations with this change?

rixner:
@albertjan, yeah, this seems innocuous, but something about it has been nagging at me.  I will think about it and try to test it some more.  Have you confirmed that it works with your ```property``` infrastructure?

rixner:
I don't see any issues with this.

PR #612, opened 2016-08-02
fix comment for importBuiltinWithBody

albertjan:
@rixner yeah this comment was completely wrong I copied to much there 😄

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -1278,9 +1278,6 @@ Sk.builtin.next_ = function next_ (iter, default_) {
     return nxt;
 };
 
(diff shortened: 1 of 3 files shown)


rixner:
Looks much better to me!

rixner:
I'm also just now noticing that ```Sk.builtin.property``` is still in ```builtin.js``` and it throws a ```NotImplementedError```.  Shouldn't that be removed?

albertjan:
Oh yeah good one. I'll add that to this PR.

LeszekSwirski:
:+1:

PR #613, opened 2016-08-02
Copy module

rixner:
This is an implementation of the ```copy``` and ```types``` modules.  They are taken directly from CPython with minor modifications (and things commented out that Skulpt doesn't yet support).  There are also internal Skulpt updates to support the features the ```copy``` module requires, including the ```id``` builtin.

I'm not sure if we have already established the proper way to cite code from CPython.  Please speak up if you believe the copyright notices at the top of the relevant files are insufficient (or should be handled differently).  It does seem like perhaps the LICENSE file at least needs to be updated (regardless of anything to do with this pull request).

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -559,7 +559,8 @@ Sk.builtin.dir = function dir (x) {
         var internal = [
             "__bases__", "__mro__", "__class__", "__name__", "GenericGetAttr",
             "GenericSetAttr", "GenericPythonGetAttr", "GenericPythonSetAttr",
-            "pythonFunctions", "HashNotImplemented", "constructor"];
+            "pythonFunctions", "HashNotImplemented", "constructor", "__dict__"
+        ];
         if (internal.indexOf(k) !== -1) {
             return null;
         }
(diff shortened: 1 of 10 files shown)


albertjan:
Wow! So the importing of internal python files seems to work pretty well 😄. 

I think the headers are fine, and the LICENCE should be updated I agree.

rixner:
Currently, Scott Graham is the only listed copyright holder in the LICENSE file.  At this point, should it be all 47 contributors to the project?  Not sure how to get the years correct, though, other than to look over all of the commits, which seems like a big hassle...  Does anyone know if github (or any external tool) has any automatic way of generating the copyright names/years based on the commits?

rixner:
Arguably, though, committers only should be credited with copyright of files they created/modified, which further complicates the issue, since we don't have copyright notices in individual files.

albertjan:
the licence can just mention "Scott Graham and the committers".

rixner:
This generates a starting point that we could clean up (if people think it's worth doing more than @albertjan's suggestion):

https://github.com/klange/generate-copyright

bnmnetp:
That generates a pretty long list!

* Should we have notices in each file?  If so I think we should do that all at once in a separate PR.  It would be pretty easy to modify that script to spew out a minimal license message as well as contributors and then run it on all the files

* I've looked over the copy module stuff and I say: LGTM

albertjan:
I've merged this lets open a new issue for the LICENSE.

PR #615, opened 2016-08-04
Update license

bnmnetp:
OK, just for the sake of getting this started, I made these simple updates to the LICENSE file

I looked at a bunch of projects, and it seems most (including Python) do not have copyright notices for every file.  Lots of things in the standard lib do have comments that indicate their origins but no explicit claim of copyright.

I do think that when we basically just drop in a file from CPython or PyPy we ought to acknowledge those origins in the file.


albertjan:
I agree!

PR #616, opened 2016-08-04
classmethod

bnmnetp:
Same deal as for the staticmethod except this one does not work.  It is exposing some strange closure bug that needs someone smarter than me to track down.

running the unit tests will reveal:

```
Ran 1 tests, passed: 1 failed: 0

Test threw exception in  TestDecorators.test_classmethods  (ExternalError: TypeError: Cannot read property 'tp$name' of undefined on line 11)
Ran 3 tests, passed: 2 failed: 1
```

The line 11 in question is line 11 in classmethod.py

Any ideas???

### DIFF
--- a/src/classmethod.py
+++ b/src/classmethod.py
@@ -1,0 +1,12 @@
+class classmethod(object):
+    "Emulate PyClassMethod_Type() in Objects/funcobject.c"
+
+    def __init__(self, f):
+        self.f = f
+
+    def __get__(self, obj, klass=None):
+        if klass is None:
+            klass = type(obj)
+        def newfunc(*args):
+            return self.f(klass, *args)
+        return newfunc
--- a/src/compile.js
+++ b/src/compile.js
@@ -1471,9 +1471,14 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
         }
     }
     if (hasFree) {
-        funcArgs.push("$free");
-        this.u.tempsToSave.push("$free");
+        if (vararg) {
+            this.u.varDeclsCode += "$free = arguments[arguments.length-1];"
+        } else {
+            funcArgs.push("$free");
+            this.u.tempsToSave.push("$free");
+        }
     }
+
     this.u.prefixCode += funcArgs.join(",");
 
     this.u.prefixCode += "){";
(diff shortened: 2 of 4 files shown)


albertjan:
Hey @bnmnetp I've (accidentally) pushed a "temporary" fix for the problem to this branch. This shows what goes wrong.

albertjan:
Maybe some of the other @skulpt/committers can shed some light on a more performant and permanent solution to this.

bnmnetp:
That does seem like a band-aid on a problem that is happening further upstream.

albertjan:
It has something to do with the fact that it takes varargs, so the $free variable is usually placed at the end of the list, which is impossible for a function with varargs. I'm definitely not saying we ship it like this but it does show clearly what goes wrong 😄

meredydd:
What does `$free` represent here? (Don't have much context in this area of the code)

albertjan:
@meredydd this example explains it best I think:

working:

```py
class Test():
    def __init__(self):
        self.x = "testerdetest"
    
    def test(self):
        def plop():
            return self.x
        
        return plop

t = Test()

print t.test()();
```

broken:

```py
class Test():
    def __init__(self):
        self.x = "testerdetest"
    
    def test(self):
        def plop(*x):
            return self.x
        
        return plop

t = Test()

print t.test()(1);
```

the `$free` variable is usually added to the end of the parameter list, so when there are no parameters it can find the variables in `$free` fine, but when there are varargs is takes off the first arg here:

```js
/*   656 */             Sk.builtin.pyCheckArgs("plop", arguments, 0, Infinity, false, true);
/*   657 */             x = new Sk.builtins['tuple'](Array.prototype.slice.call(arguments, 1)); /*vararg*/
```

I didn't want to touch this as you've probably done this on puropse 😄. But this explains why adding the `self` as `$free` at the start fixes it.

albertjan:
After brooding on it for a while I figured out I could just slice off the end of the array when I have free vars.

bnmnetp:
Thanks @albertjan !  This problem crops up a lot when you write decorator functions and want to pass on *args, **kwargs.  I think Having this fixed will enable a bunch of good stuff.

albertjan:
Apparently I was too quick with my comment 😄

albertjan:
This should do the trick! 😄

bnmnetp:
Maybe... But it fixes a lot.  including this longstanding problem:

```
def call_counter(func):
    print('wrapping func')
    def wrap(*args, **kwargs):
        wrap.counter += 1
        return func(*args,**kwargs)
    wrap.counter = 0
    return wrap

@call_counter
def fact(n):
    if n < 1:
        return 1
    else:
        return n * fact(n-1)
#fact = call_counter(fact)

@call_counter
def fib(n):
    if n <= 1:
       return 1
    else:
       return fib(n-1) + fib(n-2)


print('calling fact')
print(fact(10))
print(fact.counter)

print(fib(20))
print(fib.counter)
```

I would love to accept this before I lock down the version of skulpt for the Fall semester.

waywaaard:
@albertjan This would allow us to implement the sense_stick function wrapping - at least I think so.

And we should recheck the datetime module and if we can now easily use the python version in skulpt

albertjan:
@bnmnetp let me have one last stab at it tonight 😄.

bnmnetp:
Thanks @albertjan !

Anyone else want to weigh in on this or are we good to accept?

PR #618, opened 2016-08-16
Delattr

rixner:
Implements ```delattr``` and cleans up ```setattr``` and ```hasattr``` a little bit.

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -789,18 +789,21 @@ Sk.builtin.getattr = function getattr (obj, name, default_) {
 };
 
 Sk.builtin.setattr = function setattr (obj, name, value) {
-
-    Sk.builtin.pyCheckArgs("setattr", arguments, 3, 3);
-    if (!Sk.builtin.checkString(name)) {
-        throw new Sk.builtin.TypeError("attribute name must be string");
+    // cannot set or del attr from builtin type
+    if (obj === undefined || obj["$r"] === undefined || obj["$r"]().v.slice(1,5) !== "type") {
+        Sk.builtin.pyCheckArgs("setattr", arguments, 3, 3);
+        if (!Sk.builtin.checkString(name)) {
+            throw new Sk.builtin.TypeError("attribute name must be string");
+        }
+        if (obj.tp$setattr) {
+            obj.tp$setattr(Sk.ffi.remapToJs(name), value);
+        } else {
+            throw new Sk.builtin.AttributeError("object has no attribute " + Sk.ffi.remapToJs(name));
+        }
+        return Sk.builtin.none.none$;
     }
-    if (obj.tp$setattr) {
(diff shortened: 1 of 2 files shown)


src/builtin.js
```diff
@@ -1233,9 +1256,37 @@ Sk.builtin.callable = function callable (obj) {
...
+                } else {
+                    throw e;
+                }
+            });
+        });
+        return ret;
+    } // cannot set or del attr from builtin type
+    if (obj["$r"]().v.slice(1,5) !== "type") {
+        if (obj.ob$type === Sk.builtin.type && obj[attr.v] !== undefined) {
+            obj[attr.v] = undefined;
```
albertjan: do we want to `delete obj[attr.v]` too/instead?

src/builtin.js
```diff
@@ -1233,9 +1256,37 @@ Sk.builtin.callable = function callable (obj) {
...
+            }, function(e) {
+                if (e instanceof Sk.builtin.AttributeError) {
+                    throw new Sk.builtin.AttributeError(Sk.abstr.typeName(obj) + " instance has no attribute '"+ attr.v+ "'");
+                } else {
+                    throw e;
+                }
+            });
+        });
+        return ret;
+    } // cannot set or del attr from builtin type
```
albertjan: could we achieve this with `delete` too maybe?

albertjan:
another great addition btw 😄

bnmnetp:
And what about this one?  It looks good to me.  @albertjan is an explicit delete better than setting to unknown?

albertjan:
@bnmnetp I think we can merge this too, no need to keep this hanging because of a couple of questions from me. We might want to log an issue that states that you can't delattr from builtin types.

bnmnetp:
OK, I'll merge this.  You can log the issue as you will state the problem better than I.

rixner:
Are you sure CPython allows you to delattr from builtin types?

Sorry we never responded to whether it would be better to use ```delete```.  I'm really not sure and I haven't had time to look at it.  @robyntorregrosa is tied up with the start of the semester, so neither has she.

albertjan:
`TypeError: can't set attributes of built-in/extension type 'int'`

you're right

src/builtin.js
robyntorregrosa: We shouldn't use `delete obj[attr.v]` here because JavaScript follows a different set of rules: "if a property with the same name exists on the object's prototype chain, the object will inherit that property from the prototype." (from [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete](url) Mozilla Developer Network Documentation). When I tried using `delete` it appears as though the property reference is not being completely deleted. This might be because the property is named elsewhere in the prototype chain and that reference is used to replace the one we are attempting to delete.

albertjan: Alright sounds logical. Thank you for your explanation. :smile:

Issue #620, opened 2016-08-16
lshift of 0 should not become long


OPiMedia:
```Python
type(0 << 1)  # is <type 'long'>, but must be <type 'int'>
```

Bug found by Sergey Korytnik, with CodeSkulptor
[http://www.codeskulptor.org/#user41_7649LLA6bW_1.py](http://www.codeskulptor.org/#user41_7649LLA6bW_1.py)

EDIT: In fact there are the same error with `a << 0`.
And _maybe_ the general test `tmp <= this.v` is completely wrong!
I don't understand why "overflow" in JavaScript may decide when the promotion to `long` is correct in Python.

albertjan:
This has been fixed. Thanks @OPiMedia for pointing it out.

PR #621, opened 2016-08-16
Corrected Bug #620: lshift of 0 should not become long

OPiMedia:
Simple correction of this bug, with test cases.

### DIFF
--- a/src/int.js
+++ b/src/int.js
@@ -633,7 +633,7 @@ Sk.builtin.int_.prototype.nb$lshift = function (other) {
                 throw new Sk.builtin.ValueError("negative shift count");
             }
             tmp = this.v << shift;
-            if (this.v !== 0 && shift != 0 && tmp <= this.v) {
+            if (this.v !== 0 && shift !== 0 && tmp <= this.v) {
                 // Fail, recompute with longs
                 return new Sk.builtin.lng(this.v).nb$lshift(other);
             }
--- a/src/int.js
+++ b/src/int.js
@@ -637,7 +637,9 @@ Sk.builtin.int_.prototype.nb$lshift = function (other) {
                 // Fail, recompute with longs
                 return new Sk.builtin.lng(this.v).nb$lshift(other);
             }
-        } else if (tmp !== undefined) {
+        }
+
+        if (tmp !== undefined) {
             tmp = /** @type {number} */ (tmp);
             return new Sk.builtin.int_(tmp);
         }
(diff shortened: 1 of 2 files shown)


albertjan:
Hi! Thanks for your contribution! Looks good, but I don't really understand the logic of the branch test for when to do a long calc.

bnmnetp:
@albertjan I was looking at this too, and the logic is sound.  the if statement is testing for an error condition.  When you left shift the number should get bigger.  If it doesn't get bigger or stays the same then the shift didn't work right.  Except when you are shifting 0.

Unless I'm messed up, then LGTM.

OPiMedia:
@albertjan I test 3 special cases with `0`.
And I added these 3 special cases with `0L` to be sure not to have broken the correct `long` behavior.
P.-S.: I have corrected the #620 message.

albertjan:
Alright cool 😄 

@bnmnetp, @OPiMedia  this logic operates on the assumption that the `<<` operator in js will return a smaller number when it fails. But it doesn't because it will push the 32 bit number though numerous bytes before coming to a solution which maybe smaller but also could be bigger.

This shows what happens clearly:

```js
9<<31
> -2147483648
9<<32
> 9
9<<33
> 18
```

which gives very strange results try `9<<100` for example.

albertjan:
@OPiMedia your solution is correct, but looking at this opened a "new" can of worms.

OPiMedia:
This test seem to me a general correct way to check if the _JavaScript_ calculation failed:
```Python
if (tmp >> shift !== this.v) {
```
But it causes a problem with 'test/run/t367.py' that don't understand.

albertjan:
But like I said this is not because of your code, but because of another problem in the code that we should address. And if we do the `0` shift should get fixed too. 

```py
print 9<<35
print 9<<67
print 9<<3
```
returns:
```
72
72
72
```

albertjan:
@rixner how do you feel about leaving out the javascript `<<` as it only does 32 bit number and will produce wrong values, and moving to:

```js
function lshift(num, shift) {
    return num * Math.pow(2, shift);
}
```

it's slower, but the loops we'll have have to jump through a LOT of hoops checking stuff if we want to keep the `<<`.

rixner:
The question is how much slower is it?  If it's noticeably slower, I'd be opposed.  However, I think it's relatively easy to optimize.  There are only so many shift values (10s?) for which this works before you overflow Javascript numbers and receive an incorrect result, so it seems like you could precompute the multipliers pretty easily and so something like the following might work:

```
function lshift(num, shift) {
    if (shift < X) {
        return num * shift_multiplier[shift];
    } else {
        // slow path
    }
}
```

I doubt multiplication is that much slower than << given all of the other overheads to get down to this function.  It's the ```Math.pow``` that's likely to be the noticeable part.

Unfortunately, the semester is about to begin for me, so I'm not going to have a lot of time for Skulpt in the coming week...

PR #622, opened 2016-08-19
Leftshift fix

albertjan:
Hey @OPiMedia thanks for pointing out the error in skulpt, it unearthed a bigger problem that this solves. 

@rixner I haven't done your perf script on this. But I will. 

ref #620 #621


albertjan:
On master:

`./m run leftshiftperf.py  29.80s user 2.01s system 100% cpu 31.746 total`

On this branch:

`./m run leftshiftperf.py  25.40s user 1.95s system 100% cpu 27.311 total`

Code:

```py
import sys
for i in range(-10000, 10000):
    print i
    for j in range(0, 64):
        print i << j
```

albertjan:
both pystone and field.py don't have leftshifts 😄

rixner:
LGTM!

SergeyKorytnik:
I would suggest to look at the int_lshift function from CPython implementation before residing to the array lookup.

The basic idea is that left shift result can be an integer if and only if it is revertible.
That is if right shift will restore the original value.

if( this.v === 0 || shift === 0)
  return this
if( shift >= 32 )
{
  reside to long shift...
}
tmp = this.v << shift;
if( tmp >> shift != this.v )
{
  reside to long shift...
}
return tmp;

albertjan:
@SergeyKorytnik probably 😄 

I like the idea of right shifting the result and checking if its still the same number. The only problem we have is, skulpts ints are 53bits. Because all numbers in javascript are ieee 753 64 bit floating point numbers. (unless you do binary operations on them). We use this cutoff point because above that we have to use a biginteger implementation, that is a lot slower. 

We've discussed making the integers 32bits before, but we would only after careful examination and refactoring of the code see the performance improvements that would be associated with the 32bit numbers. And it would add another division in our number system at 53 bits. 

we could add the `>>` `<<` check in but are shifts really that much quicker then multiplications? What I wanted to show was that we aren't getting much worse performance wise.

albertjan:
your script on this branch:
```sh
./m run leftshiftperf.py  1.23s user 0.08s system 101% cpu 1.285 total
./m run leftshiftperf.py  1.22s user 0.08s system 101% cpu 1.282 total
./m run leftshiftperf.py  1.24s user 0.08s system 101% cpu 1.301 total
```

on master:

```sh
./m run leftshiftperf.py  2.43s user 0.08s system 101% cpu 2.479 total
./m run leftshiftperf.py  2.59s user 0.09s system 101% cpu 2.645 total
./m run leftshiftperf.py  2.56s user 0.09s system 101% cpu 2.620 total
```

Wow unexpected, I did expect to see a little perf hit with this script but the perf gain is even more pronounced here.

Which may be because of the broken implementation, wrongfully going down the long left shift path.

SergeyKorytnik:
I was not aware about 53 bit ints in skulpt.

Yes. The performance test results are unexpected. I can't explain them.
I tried a hypothesis that the performance gain is totally due to the shortcut for  0 shifts.
But my experiments <http://www.codeskulptor.org/#user41_NCpbaGqnSDaPc0p_0.py> showed that the hypothesis is wrong.  I will try to think more about it.

It was a good lesson that I should  test my suggestions  myself prior to suggesting anything.
Sorry for taking your time.

albertjan:
No no I agreed with you I had the exact same expectations :). Don't worry
about it.

bnmnetp:
So if I'm following correctly.  This is ready to merge.

albertjan:
I think so yes. 😄

PR #624, opened 2016-08-22
make it possible to have an input fun that takes a prompt

### LINKED ISSUE #623: No prompt displayed for input and raw_input
I just got a bug report on this from users of my books.   I don't know when this happened, but it looks like a few things have changed.  Sk.inputfun looks like it is no longer setable  but more importantly, the prompt for input and raw_input has disappeared.

I need to get this back asap.  @albertjan it looks like you were the last person working on input and raw input.  Any advice on how to get this working again??

albertjan:
By setting `Sk.inputfunTakesPrompt` to true, the prompt will get sent to the input function, instead of to the stdout. 

fixes #623

### DIFF
--- a/support/run_template.html
+++ b/support/run_template.html
@@ -23,19 +23,38 @@ function showjs(text) {
 }
 
 
-function builtinRead(x)
-{
-    if (Sk.builtinFiles === undefined || Sk.builtinFiles["files"][x] === undefined)
+function builtinRead(x) {
+    console.log(x);
+    if (Sk.builtinFiles === undefined || Sk.builtinFiles["files"][x] === undefined) {
         throw "File not found: '" + x + "'";
-    return Sk.builtinFiles["files"][x];
+    }
+
+    var file = Sk.builtinFiles["files"][x];
+
+    if (!file) {
+        return new Promise(function (accept, reject) {
+            function reqListener () {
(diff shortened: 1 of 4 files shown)


albertjan:
@bnmnetp this should solve your problem with `window.prompt`

support/run_template.html
```diff
@@ -23,19 +23,38 @@
...
 }
 
 function runit(myDiv) {
    var prog = document.getElementById(myDiv+"_code").value;
    var mypre = document.getElementById(myDiv+"_pre");
 
-   Sk.inputfun = function() {
-       return new Promise(function (resolve) { resolve(window.prompt()); });
+   Sk.inputfun = function(prompt) {
+       return new Promise(function (resolve) { resolve(window.prompt(prompt)); });
```
bnmnetp: I think this  same pattern be applied to env.js where the base version of Sk.inputfun is defined?

It seems to work ok without the promise wrapper but I feel like it would be more correct.

albertjan: I put the promise there to test if it would work with a promise. the window.prompt function is synchronous so it's safe to use it without a promise.

bnmnetp:
Great, I'm going to accept this so I can get it into production ASAP.

PR #630, opened 2016-08-31
Fix `brun` command

meredydd:
20d27cce65da744d9d810ff7be34ce8785d7fa59 broke the `./m brun` command, by requiring a `root` substitution in `run_template.html`. This patch fixes it by adding a trivial `root=""`.


meredydd:
@albertjan to review, please, as you wrote the original commit

albertjan:
you're right, that slipped in accidentally.

meredydd:
:grinning:

albertjan:
Can you in return look at #619

PR #632, opened 2016-09-08
return .instance from ensureAnonymous

### LINKED ISSUE #631: Anonymous turtle is not working?
Maybe it is just my faulty memory but I thought that

```
from turtle import *
forward(10)
```

used to work?  I see evidence of an anonymous turtle in the code but can't make it work.

Brad

albertjan:
Fixes #631.

But we should really try and figure out whats going on here. But @bnmnetp you want to ship this to runestone probably 😄.

### DIFF
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -1502,7 +1502,7 @@ function generateTurtleModule(_target) {
             _anonymousTurtle = _module.Turtle();
         }
 
-        return _anonymousTurtle;
+        return _anonymousTurtle.instance;
     }
 
     function getTarget() {


bnmnetp:
@albertjan That sure seems to fix it!  And as far as I can tell does not have any impact on anything else.

albertjan:
Yeah this fix is in the function that makes the anonymous version of the turtle when you do `from turtle import`

bnmnetp:
Yeah, and I just looked at the diff of the last time that line was changed.  Prior to f9ed0551 ensure Anonymous did not return a value!  So this line:

```
                self       = classMethod ? ensureAnonymous() : args.shift(),
```

So, prior to f9ed0551 ensureAnonymous didn't return anything and the above was guaranteed to always use args.shift()

Not sure that is helpful. but may explain why it changed behavior at some point.

albertjan:
Yeah it's probably been broken for quiet a while.

albertjan:
I think it's safe to merge this.

PR #634, opened 2016-09-19
Move `__getattr__` and friends into instance `tp$getattr()` and friends

meredydd:
They never had any business being in the global `Sk.abstr.gattr()` (or `sattr`) function in the first place. Their presence was causing errors accessing any class attribute of a class with `__setattr__`, or any new-style class (which inherits a `__setattr__` from `object`). This is one of those "I can't believe it lasted that long" bugs.

### DIFF
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -819,31 +819,11 @@ Sk.abstr.gattr = function (obj, nameJS, canSuspend) {
         throw new Sk.builtin.AttributeError("'" + objname + "' object has no attribute '" + nameJS + "'");
     }
 
-
     if (obj.tp$getattr !== undefined) {
-        f = obj.tp$getattr("__getattribute__");
+        ret = obj.tp$getattr(nameJS, canSuspend);
     }
 
-    if (f !== undefined) {
-        ret = Sk.misceval.callsimOrSuspend(f, new Sk.builtin.str(nameJS));
-    }
-
-    ret = Sk.misceval.chain(ret, function(ret) {
-        var f;
-
-        if (ret === undefined && obj.tp$getattr !== undefined) {
-            ret = obj.tp$getattr(nameJS);
-
-            if (ret === undefined) {
-                f = obj.tp$getattr("__getattr__");
-
-                if (f !== undefined) {
-                    ret = Sk.misceval.callsimOrSuspend(f, new Sk.builtin.str(nameJS));
-                }
-            }
-        }
-        return ret;
-    }, function(r) {
+    ret = Sk.misceval.chain(ret, function(r) {
         if (r === undefined) {
             throw new Sk.builtin.AttributeError("'" + objname + "' object has no attribute '" + nameJS + "'");
         }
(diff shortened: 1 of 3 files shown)


albertjan:
very awesome! 😄 just jscs to make happy.

```
$ jscs src/*.js
Missing semicolon after statement at src/type.js :
   267 |                    }
   268 |                })
   269 |            };
---------------------^
   270 |
   271 |            getf = Sk.builtin.object.prototype.GenericGetAttr.call(this, "__getattribute__");
1 code style error found.
```

albertjan:
this'll probably speed things up too.

meredydd:
Right. JSCS is now happy, and I've PRed all three waypoints on my journey to a working `datetime` module. Please review :)

albertjan:
Did a quick perf test on @rixner 's field.py went from 5.8/5.6 to 5.3/5.3. This looks good to me!

meredydd:
Is that a code review +1 as well as a performance review?

meredydd:
Remind me of the policy - do we need another +1 from @skulpt/committers to merge this?

bnmnetp:
Yep, except for "trivial" changes we agreed to have two @skulpt/committers other than the author give it the thumbs up.   I'll try to look at these this morning.

bnmnetp:
OK, I applied this one and #635 and ran through a bunch of my examples.

All of the decorator and descriptor stuff works well.

However, The simple turtle test:

```
from turtle import *

forward(100)
```

Now fails with the error:  ``ExternalError: TypeError: undefined is not an object (evaluating 'a.slice') on line 3``

meredydd:
Hey, sorry for the dumb question, but what's the simplest way to get up a turtle environment to test this out? Would like to reproduce this.

meredydd:
...of course, moments after sending that, I find `src/example/turtle.html`. Stand by...

bnmnetp:
@meredydd Any thoughts on this one yet?  @albertjan you fixed it the last time, can you imagine any connection between the anonymous turtle and the error I'm getting after applying this PR??

albertjan:
@bnmnetp it works for me.

albertjan:
@bnmnetp do you have any local changes I'm not aware off?

bnmnetp:
@albertjan did you also apply OK, #635 ?    I guess that must be the culprit.

My only custom changes are to unittest.gui  which are pretty far from this example.

Brad

meredydd:
Apologies for the radio silence; I'll have some cycles to spend on this one soon, I promise!

meredydd:
Wow, that took longer to get round to than I thought.

In any case, I have managed to replicate this failure on current master. (I needed to replace the script in `turtle.html` with @bnmnetp's simpler version - the "draw an octagon" default script works just fine for me.)

I therefore suspect this has nothing to do with this PR, and should be handled separately.

@albertjan, @bnmnetp, can I get a +1 on this? I'm settling down for another round of bugfixes, and it would be good not to be working so far out on a branch :)

albertjan:
I'm ok with merging this @bnmnetp we'll fix this before the next release.

rixner:
This all looks fine to me.  I don't ever use turtle, so it seems like @bnmnetp should agree to the merge, though.

PR #635, opened 2016-09-20
Support `__new__` in object construction

meredydd:
This PR supports the `__new__` method when constructing objects.

In order to make the tests pass, I also had to clean up our support for static/class methods and hand-rolled Python descriptors (I've made them use the proper descriptor system, which was already in use for methods, and removed the extraneous checks in function invocation).

This PR depends upon #634 because it also touches the attribute-lookup mechanism.

### DIFF
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -819,31 +819,11 @@ Sk.abstr.gattr = function (obj, nameJS, canSuspend) {
         throw new Sk.builtin.AttributeError("'" + objname + "' object has no attribute '" + nameJS + "'");
     }
 
-
     if (obj.tp$getattr !== undefined) {
-        f = obj.tp$getattr("__getattribute__");
+        ret = obj.tp$getattr(nameJS, canSuspend);
     }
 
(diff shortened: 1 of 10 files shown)


albertjan:
This one looks good to me too! Awesome addition.

rixner:
This seems to be a 10+% slowdown for me on code that isn't using ```__new__```.   I would say that the common case (by far) is code that doesn't use ```__new__```, so I'd obviously prefer to see no slowdown for such code.

rixner:
I rebased this against master and played around with it some more.  It seems not to be slowing things down now.  I don't know if there was a bad interaction with something prior to the recently accepted pull requests (or a good interaction with one of them), I made a mistake, or what.

Regardless, I withdraw my objections and I think both this and #636 should be merged.

meredydd:
Thanks! Your role as Skulpt's unofficial Performance Watchdog is also very much appreciated - 5% here, 5% there, and before you know it we're crawling.

(And yes, pulling in external modules is a definite goal!)

PR #641, opened 2016-09-28
corrected error on regex with assignments

amr66:
extended the regex expression for assignments, so python augmented assignment like += will work.


albertjan:
Loving it 😄 
screen shot 2016-09-28 at 16 22 08

Could you also change it in `doc/static/repl.js`? (I know the cardinal sin in computer world duplicated code, and I'm not gonna make up excuses :trollface:)

albertjan:
Looks good @skulpt/committers merge at will.

albertjan:
@skulpt/committers I'll merge this if no one objects today.

Issue #642, opened 2016-09-28
urllib needs to be updated to use suspensions.


bnmnetp:
As I'm starting to see deprecation warnings from Chrome about non async xmlhttp sends, we need to get the request object of urllib updated to use Suspensions pretty quick.  There is an ominous date of October 16 in the message.

PR #643, opened 2016-09-29
make urllib use suspensions

albertjan:
@bnmnetp Here you go :) #642


bnmnetp:
Yowza!  Thanks @albertjan !

PR #658, opened 2017-01-03
Make closures work with keyword argument matching

meredydd:
When a Skulpt function has free variables, it takes an extra argument (the parent's `$cell`) to look them up in.

Unfortunately, we were adding that extra argument *before* we matched keyword arguments with named arguments, which was leading to odd errors. This PR produces the correct behaviour, by adding the closure argument after keyword matching has been performed. Test included.

### DIFF
--- a/src/function.js
+++ b/src/function.js
@@ -229,22 +229,6 @@ Sk.builtin.func.prototype.tp$call = function (args, kw) {
     var name;
     var numargs;
 
-    // note: functions expect 'this' to be globals to avoid having to
-    // slice/unshift onto the main args
-    if (this.func_closure) {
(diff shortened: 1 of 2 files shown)


albertjan:
thanks :) @skulpt/committers lets merge.

albertjan:
@meredydd I think we should merge all your PR's most of them have been mulled over long enough. Can you recommend the order in which they should be merged?

meredydd:
All my PRs from last autumn are stacked - if you merge the datetime one, that will bring in  the attribute fix and support for `__new__` for free.

The two I opened today are standalone, and can be merged in any order.

rixner:
This looks good to me.

PR #659, opened 2017-01-04
Add type checking to set methods.

rixner:
Makes sure that the correct number and types of arguments are passed.

### DIFF
--- a/src/set.js
+++ b/src/set.js
@@ -212,40 +212,46 @@ Sk.builtin.set.prototype["issuperset"] = new Sk.builtin.func(function (self, oth
 });
 
 Sk.builtin.set.prototype["union"] = new Sk.builtin.func(function (self) {
-    var S, new_args;
+    var S, i, new_args;
 
     Sk.builtin.pyCheckArgs("union", arguments, 1);
 
-    S = Sk.builtin.set.prototype["copy"].func_code(self),
-      new_args = Array.prototype.slice.call(arguments); //copy array
+    S = Sk.builtin.set.prototype["copy"].func_code(self);
+    new_args = [S];
+    for (i = 1; i < arguments.length; i++) {
+        new_args.push(arguments[i]);
+    }
 
-    new_args[0] = S;
     Sk.builtin.set.prototype["update"].func_code.apply(null, new_args);
     return S;
 });
 
 Sk.builtin.set.prototype["intersection"] = new Sk.builtin.func(function (self) {
-    var S, new_args;
+    var S, i, new_args;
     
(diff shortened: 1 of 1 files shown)


src/set.js
```diff
@@ -193,39 +207,55 @@ Sk.builtin.set.prototype["issubset"] = new Sk.builtin.func(function (self, other
...
-    var S = new Sk.builtin.set(self);
-    for (i = 1; i < arguments.length; i++) {
-        Sk.builtin.set.prototype["update"].func_code(S, arguments[i]);
-    }
+    var S, new_args;
+
+    Sk.builtin.pyCheckArgs("union", arguments, 1);
+
+    S = Sk.builtin.set.prototype["copy"].func_code(self),
+      new_args = Array.prototype.slice.call(arguments); //copy array
```
meredydd: Generally worth avoiding `slice.call(arguments)` where possible - it kills the optimiser. (Annoying, because it's so much nicer than the manual walk method). That said, I'm not the performance nut here, so if you're OK with it, I am ;)

meredydd: Also, is this line a miss-indentation?

src/set.js
```diff
@@ -193,39 +207,55 @@ Sk.builtin.set.prototype["issubset"] = new Sk.builtin.func(function (self, other
...
 
 Sk.builtin.set.prototype["intersection"] = new Sk.builtin.func(function (self) {
-    var S = Sk.builtin.set.prototype["copy"].func_code(self),
-        new_args = Array.prototype.slice.call(arguments); //copy array
+    var S, new_args;
     
+    Sk.builtin.pyCheckArgs("intersection", arguments, 1);
+
+    S = Sk.builtin.set.prototype["copy"].func_code(self),
+      new_args = Array.prototype.slice.call(arguments); //copy array
```
meredydd: Indentation again (surprised `jscs` isn't raising hell over this).

Actually, also a trailing comma on the previous line - I think this got copied and pasted out of the `var` declaration above, which explains both the indent and the comma.

meredydd:
Good stuff. (Playing with the new GitHub reviews feature; apologies for verbosity.)

Two formatting nits, and provided there are tests for the functions whose implementations you've altered, I'm +1 for merge.

rixner:
Yes, there are already tests for all of these functions.  I will clean up the bad indentation/commas (which did come from moving the assignments out of the declaration, as you suspected). The ```slice.call``` stuff was already there, I'll fix that too, while I'm at it.

rixner:
Cosmetic things should be fixed now.

bnmnetp:
Feel free to merge.

PR #660, opened 2017-01-12
Make dunder lookup use tp$getattr() again. Gives us better error messages.

meredydd:
tl;dr Fix "this object is not callable"/"not iterable" error messages.

This mostly reverses commit ae9fed464, which tried to accommodate dynamically provided dunder-functions by using Sk.abstr.gattr(). The correct fix (implemented in commit 2e43d47 / PR #634) was to make tp$getattr() do dynamic lookup properly. Using Sk.abstr.gattr() just made things complicated, and in particular made `__call__`/`__iter__` lookup failures come out as generic AttributeErrors rather than TypeErrors with helpful messages.

Reversing this means that we get better error messages (like "this object is not callable/iterable") again.

### DIFF
--- a/src/type.js
+++ b/src/type.js
@@ -325,27 +325,27 @@ Sk.builtin.type = function (name, bases, dict) {
             return canSuspend ? r : Sk.misceval.retryOptionalSuspensionOrThrow(r);
         };
         klass.prototype.tp$call = function (args, kw) {
-            return Sk.misceval.chain(Sk.abstr.gattr(this, "__call__", true), function(callf) {
+            return Sk.misceval.chain(this.tp$getattr("__call__", true), function(callf) {
+                if (callf === undefined) {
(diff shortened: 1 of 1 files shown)


waywaaard:
How should it be implemented? Is this the right behavior or just a fix for the newly introduced errors?

meredydd:
The behaviour was, and is, correct - this patch brings no functional changes. The only change is that the code is a bit simpler and the errors are more human-friendly.

meredydd:
@waywaard poke?

albertjan:
Looks good to me.

waywaaard:
I am not allowed to vote^^. Sorry for the late reply.

bnmnetp:
@waywaaard I am happy to grant you voting/committing rights.  Last time I asked I did not get a response...

waywaaard:
@bnmnetp I am sorry to hear, that you did not receive a response. Though, I am not sure where you send the message.

PR #662, opened 2017-01-25
Change dependencies to devDependencies

amasad:
Non of the dependencies are actually needed at runtime


waywaaard:
Correct, those are only required for building skulpt. `npm install` installed dev and normal deps anyways, so there should be no problem for pulling this in.

PR #665, opened 2017-02-03
Change all uses of performance to window.performance

### LINKED ISSUE #625: Cannot use time module ("performance" not defined)
I've received a few second-hand reports of a Skulpt-based application failing with `ExternalError: ReferenceError: Can't find variable: performance on line 50`. This doesn't occur on new versions of Firefox (e.g. 48) or Chrome (e.g. 51), but it sounds like it does occur on some older (unknown) versions of Safari, and possibly also some versions of Internet Explorer.

I suspect the problem is that those browsers don't support `window.performance`, and Skulpt is dying when it tries to use it. In `time.js`, there are a few checks along the lines of

``` js
if (performance && performance.now) {
}
```

which should probably be

``` js
if (window.performance && window.performance.now) {
}
```

j-f1:
Fixes #625.

### DIFF
--- a/src/lib/time.js
+++ b/src/lib/time.js
@@ -1,9 +1,6 @@
 /*
  implementation of the Python time package.
 
- For higher res time we could use following apart from new Date:
- window.performance.now()
-
  notes:
  - struct_time is a structseq but structseq does not implement methods: 'n_fields', 'n_sequence_fields', 'n_unnamed_fields' yet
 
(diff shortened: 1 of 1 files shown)


waywaaard:
Seems good for me. (I suggested the change :P)

The PR also removes some trailing whitespaces.

j-f1:
@waywaaard Should I put the trailing whitespace back?

albertjan:
that won't be necessary

waywaaard:
@albertjan You can also make "real" reviews now on github. See the **Files Changed** Tab.

PR #670, opened 2017-02-15
handle promise in suspension correctly, use internal suspension data to resume

albertjan:
When you return a value from the reject handler of a promise the resulting promise is `resolved` not `rejected`. 

These changes make sure you can reject your `Sk.input` promise implementation.

### DIFF
--- a/test/unit/test_input.py
+++ b/test/unit/test_input.py
@@ -1,0 +1,32 @@
+__author__ = 'Albert-Jan'
+
+import unittest
+
+class InputFunTests(unittest.TestCase):
+
+    def setUp(self):
+        jseval("""
+            var oldInputFunTakesPrompt = Sk.inputfunTakesPrompt;
+            var oldInputFun = Sk.inputfun
+            Sk.inputfunTakesPrompt = true;
+            Sk.inputfun = function (prompt) { 
+                return new Promise(function (resolve, reject) {
+                    resolve(Sk.builtin.str(prompt + "testing"));
+                });
+            }
+        """)
+
+
+    def test_input_fun_should_return_promt_asynchronously(self):
+        res = input(">>> ")
+        self.assertEqual(res, ">>> testing")
+
+
+    def tearDown(self):
+        jseval("""
+            Sk.inputfunTakesPrompt = oldInputFunTakesPrompt;
+            Sk.inputfun = oldInputFun;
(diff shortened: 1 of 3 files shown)


meredydd:
Haven't tested, but on inspection this looks like The Right Thing to me. (Very similar to a function we use in fact - should we perhaps create a promise-to-suspension adaptor in `Sk.misceval` - sort of the inverse of `asyncToPromise()`? Not for this PR though.)

rixner:
This has been sitting around for a while, so it seems like someone should test it and merge it.

To that end, I was trying to setup a simple test of this, but found that there are other problems in ```$readline```, that made my testing ineffective.  On line 129 of file.js, it tries to remap ```prompt``` to JS.  But, on line 166, it calls ```$readline``` with prompt set to ```undefined```.  Things, unsurprisingly, blow up because of this, as ```remapToJs``` does not work with an ```undefined``` input.  I don't use this infrastructure, so I am unclear how one was supposed to use this, but as is, it seems fundamentally broken, as you can't get past the ```remapToJs``` call to run any of the other code.  The fix appears simple enough, but I don't want to make it, since clearly other people use this code successfully (?) so I must be missing something.

albertjan:
@rixner yeah you're right this is broken, not many people readline-ing from files I guess. I'll look into this a bit further. I'm probably the one that broke it too 😄

test/unit/test_input.py
```diff
@@ -0,0 +1,46 @@
...
+                    if (prompt === "error") {
+                        throw new Sk.builtin.ValueError("aarrrggg");
+                    }
+                    resolve(Sk.builtin.str(prompt + "testing"));
+                });
+            }
+        """)
+
+
+    def test_input_fun_should_return_promt_asynchronously(self):
```
j-f1: `prompt`?

albertjan: yeah :)

albertjan:
@skulpt/committers this doesn't address the error @rixner spotted but it does demonstrate the promise change is working correctly. I will address the `readline` from a file error soon.

meredydd:
Ah, crumbs. That might have been a premature merge. My bad if so. (But at least it was a strict improvement on what came before...)

albertjan:
haha no I think we're alright with this one 👍

PR #671, opened 2017-02-16
fix reserved words in get and set attr

albertjan:
Use the same reserved words mangler for `get/setattr` as the compiler uses. So we can use `get/setattr` to get fields like `default` and `delete` from objects.

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -771,13 +771,14 @@ Sk.builtin.hash = function hash (value) {
 };
 
 Sk.builtin.getattr = function getattr (obj, name, default_) {
-    var ret;
+    var ret, mangledName;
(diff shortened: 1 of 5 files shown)


rixner:
Are there tests for this?  If not, can you add some?

albertjan:
Of course.

rixner:
This looks good.

My only concern here is the "unmangling".  Should we have a function (```unfixReservedWords``` or something) to do that in case we need to change the mangling in the future?  That way, the fix and "unfix" could be in the same place, so one didn't need to hunt through the code to find places that needed to change?

This is a legitimate question of mine.  I'm not saying this shouldn't be merged without that change.

albertjan:
I agree threre is currently no such function but it's easily added, I'll do
that :).

rixner:
I say merge.

PR #674, opened 2017-03-17
add antigravity module

albertjan:
I'm sorry for the lack of unit tests, antigravity can only be proven in geneva by CERN, we lack the necessary funding to do our own tests. Maybe we can borrow the LHC some time.

Because we want to mimic cpython as closely as possible I felt it's very important to have this module as well. 

We didn't even throw a NotImplemented exception when you tried to `import antigravity`.


waywaaard:
Just a brief note, any test that would try to import that lib locally would fail due to the `window` object access.

albertjan:
I promise to update the antigravity module when we officially switch to node 😄.

albertjan:
@bnmnetp do we want to merge this? Might be fun right when students find this corner of python?

PR #680, opened 2017-05-13
Implement `try`/`finally`

meredydd:
This feature has been missing for a while!

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -1245,7 +1245,6 @@ Compiler.prototype.ctryexcept = function (s) {
             this.vexpr(handler.name, "$err");
         }
 
-        // Need to execute finally before leaving body if an exception is raised
         this.vseqstmt(handler.body);
 
         // Should jump to finally, but finally is not implemented yet
@@ -1259,15 +1258,36 @@ Compiler.prototype.ctryexcept = function (s) {
 
     this.setBlock(orelse);
     this.vseqstmt(s.orelse);
-    // Should jump to finally, but finally is not implemented yet
     this._jump(end);
+
     this.setBlock(end);
 };
 
 Compiler.prototype.ctryfinally = function (s) {
-    out("/*todo; tryfinally*/");
-    // everything but the finally?
-    this.ctryexcept(s.body[0]);
+
+    var finalBody = this.newBlock("finalbody");
+    var exceptionHandler = this.newBlock("finalexh")
+    var exceptionToReRaise = this._gr("finally_reraise", "undefined");
+    this.u.tempsToSave.push(exceptionToReRaise);
+
+    this.setupExcept(exceptionHandler);
+    this.vseqstmt(s.body);
+    this.endExcept();
+    // Normal execution falls through to finally body
+    this._jump(finalBody);
+
+    this.setBlock(exceptionHandler);
+    // Exception handling also goes to the finally body,
+    // stashing the original exception to re-raise
+    out(exceptionToReRaise,"=$err;");
+    this._jump(finalBody);
+
+    this.setBlock(finalBody);
+    this.vseqstmt(s.finalbody);
+    // If finalbody executes normally, AND we have an exception
+    // to re-raise, we raise it.
+    out("if(",exceptionToReRaise,"!==undefined) { throw ",exceptionToReRaise,";}");
+    // Else, we continue from here.
 };
 
 Compiler.prototype.cassert = function (s) {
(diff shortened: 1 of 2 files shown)


albertjan:
WOW! This is sooo great! 😄

albertjan:
I'm sorry I didn't see this until now. I will do a thorough review tonight! Man this is like one of the last compiler hurdles!

meredydd:
This was the easy one - the real bugbear is `with`. Did you see my previous
email?

albertjan:
I don't think so... Where did you send it? But now we have the finally block setup, I figured with would be easy. Since we have all the building blocks.

meredydd:
Ah, fine - I'll resend. (It's a parser/AST problem and I'm not so familiar with that code.)

Let's keep this thread for the `try`/`finally` review. I think the tests work, and the compilation is straightforward, so hopefully we can get this merged quickly.

albertjan:
Found one little niggle (skulpt): 

```
JavaScript debugger enabled
Python 2.6(ish) (skulpt, Tue May 16 2017 21:38:53 GMT+0100 (BST))
[v8: 3.25.30] on a system
>>> 
>>> def a():
... 	try:
... 		return 1
... 	finally:
... 		return 2
... 
>>> a()
1
>>> 
```

(python2)
```
Python 2.7.13 (default, Dec 18 2016, 07:03:39) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
>>> def a():
...     try:
...             return 1
...     finally:
...             return 2
... 
>>> a()
2
>>> 
```

Any return from the `try` causes the finally not to be called:

```
JavaScript debugger enabled
Python 2.6(ish) (skulpt, Tue May 16 2017 21:42:04 GMT+0100 (BST))
[v8: 3.25.30] on a system
>>> def a():
... 	try:
... 		return 1
... 	finally:
... 		print "2"
... 
>>> a()
1
>>> 
```

meredydd:
Urk - well spotted! I'll have a think about the most elegant way to siphon off the "returning" state. (I'm thinking of a `$returning` variable containing the return type, and a stack of currently-active `finally` blocks like that of exception handlers. How does that sound to you?)

albertjan:
try finally should always return from the finally block so yeah if you add a $returning variable that get returned or overwritten and returned from the finally, might be difficult to change the behaviour of return for just the finally block. 

What does cpython do in case?

meredydd:
So, the knowledge about whether we're in a 'finally' block is
statically available to the compiler, so we could only generate the
convoluted code for returns within finally. I'd say the obvious first
step, though, would be to do the dynamic check every time.

meredydd:
My previous comment was silly. Far better to do the Right Thing straight away using compiler's static knowledge. Added commits to do exactly that.

@albertjan for a review please?

meredydd:
Ack, DO NOT MERGE. Just realised I didn't cover the `break`/`continue` case. This may require re-working my implementation of `return` tracking. Sigh.

meredydd:
Right, there we go. @albertjan for review, please :)

albertjan:
nice one! 😄 I'll look at it tonight.

rixner:
I have no idea if the tests have covered all of the possible cases, but the overall infrastructure looks good to me. @meredydd, this is ready to merge, right?  If so, I will do so.

meredydd:
Yep, should be good to go! I tried to cover all the possible cases raised
by the documentation, so here's hoping...

On 13 Jul 2017 19:16, "Scott Rixner" <notifications@github.com> wrote:

I have no idea if the tests have covered all of the possible cases, but the
overall infrastructure looks good to me. @meredydd
<https://github.com/meredydd>, this is ready to merge, right? If so, I will
do so.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/skulpt/skulpt/pull/680#issuecomment-315159674>, or mute
the thread

albertjan:
🎉

eah13:
🍻 Cheers, mate!  This was a huge improvement to Skulpt

eah13:
@bnmnetp you're the keeper of versions.  This and the other major improvements seem `0.12` worthy to me.  Happy to create the release if you (& any anyone else with a druthers here) agree.

meredydd:
@eah13 You're welcome!

If we're having a release, I'd like to get `with` (#696) in as well. It both fixes a terrible parser explosion *and* adds a widely-used piece of functionality.

PR #686, opened 2017-06-07
fix args.length check, add tmp.js to .gitignore

d4rkb1ue:
turtle.js args.length check missing the a `.length` thus can't throw an error on args missing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ test/namedtests.js
 Vagrantfile
 v8.log
 src/internalpython.js
+/tmp.js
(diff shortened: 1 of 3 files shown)


albertjan:
Hey thanks for your contribution, the build is failing because it's using npm5 which doesn't work with our antquated version of node. Could you change this line: https://github.com/skulpt/skulpt/blob/master/.travis.yml#L6 to say :`- npm install npm@4 -g`? The build should be fixed then... I hope.

j-f1:
@albertjan What version of Node are you using? Why can’t you upgrade?

albertjan:
We should be able to upgrade but that would mean figuring some stuff out, which is kind of out of scope of this PR.

albertjan:
Thanks @d4rkb1ue

PR #687, opened 2017-06-07
[WIP] Install and use the latest LTS Node before testing

j-f1:


albertjan:
:tada: much better

j-f1:
It’s not actually LTS, though — just the latest stable.

albertjan:
that's good enough for me, when we switch to building on node instead of v8 we'll have to rethink this. 😄 thanks

PR #688, opened 2017-06-07
Proposal to add Python 3 unit tests to skulpt

rixner:
This pull request adds a "test3" command to skulpt.py.  It also adds a directory (unit3) to the test directory with Python 3 unit tests.  Initially, these tests are taken from our Python 2 unit tests.  The included tests all pass in CPython 3.  They do not all pass in Skulpt.

We did not change the Python 2 unit test directory to be called unit2, but that might make sense.

I see this as a jumping off point to start testing Skulpt in Python 3 mode, and to start exposing bugs.  The test3 command is not run automatically anywhere, so this should have no impact on normal use.  Presumably, dist should eventually run the Python 3 tests, too (perhaps optionally?), but at least not until they all pass.

I am submitting this largely to get feedback from the community to make sure we are headed in a direction everyone is comfortable with before going too far down this path.

### DIFF
--- a/skulpt.py
+++ b/skulpt.py
@@ -211,17 +211,20 @@ if os.environ.get("CI",False):
 
 #jsengine = "rhino"
 
-def test(debug_mode=False):
+def test(debug_mode=False, p3=False):
     """runs the unit tests."""
     if debug_mode:
         debugon = "--debug-mode"
     else:
         debugon = ""
-    buildNamedTestsFile()
-    ret1 = os.system("{0} {1} {2} -- {3}".format(jsengine, ' '.join(getFileList(FILE_TYPE_TEST)), ' '.join(TestFiles), debugon))
+    ret1 = 0
     ret2 = 0
     ret3 = 0
     ret4 = 0
+    if not p3:
+        buildNamedTestsFile()
+        ret1 = os.system("{0} {1} {2} -- {3}".format(jsengine, ' '.join(getFileList(FILE_TYPE_TEST)), ' '.join(TestFiles), debugon))
+
     if ret1 == 0:
         print "Running jshint"
         base_dirs = ["src", "debugger"]
@@ -237,31 +240,8 @@ def test(debug_mode=False):
         ret3 = os.system(jscscmd)
         #ret3 = os.system(jscscmd)
         print "Now running new unit tests"
-        ret4 = rununits(debug_mode=debug_mode)
+        ret4 = rununits(p3=p3, debug_mode=debug_mode)
     return ret1 | ret2 | ret3 | ret4
-
-def test3(debug_mode=False):
-    """runs the unit tests in python 3."""
-    if debug_mode:
-        debugon = "--debug-mode"
-    else:
-        debugon = ""
-    print "Running jshint"
-    base_dirs = ["src", "debugger"]
-    for base_dir in base_dirs:
-        if sys.platform == "win32":
-            jshintcmd = "{0} {1}".format("jshint", ' '.join(f for f in glob.glob(base_dir + "/*.js")))
-            jscscmd = "{0} {1} --reporter=inline".format("jscs", ' '.join(f for f in glob.glob(base_dir + "/*.js")))
-        else:
-            jshintcmd = "jshint " + base_dir + "/*.js"
-            jscscmd = "jscs " + base_dir + "/*.js --reporter=inline"
-    ret1 = os.system(jshintcmd)
-    print "Running JSCS"
-    ret2 = os.system(jscscmd)
-    print "Now running unit tests"
-    ret3 = rununits(p3=True, debug_mode=debug_mode)
-    return ret1 | ret2 | ret3
-
 
 def parse_time_args(argv):
     usageString = """
@@ -1334,9 +1314,11 @@ def main():
     if cmd == "test":
         test()
     elif cmd == "test3":
-        test3()
+        test(p3=True)
     elif cmd == "testdebug":
-        test(True)
+        test(debug_mode=True)
+    elif cmd == "test3debug":
+        test(debug_mode=True, p3=True)
     elif cmd == "dist":
         dist(options)
     elif cmd == "regengooglocs":
--- a/skulpt.py
+++ b/skulpt.py
@@ -240,6 +240,29 @@ def test(debug_mode=False):
         ret4 = rununits(debug_mode=debug_mode)
     return ret1 | ret2 | ret3 | ret4
 
+def test3(debug_mode=False):
+    """runs the unit tests in python 3."""
+    if debug_mode:
+        debugon = "--debug-mode"
+    else:
+        debugon = ""
+    print "Running jshint"
+    base_dirs = ["src", "debugger"]
+    for base_dir in base_dirs:
+        if sys.platform == "win32":
+            jshintcmd = "{0} {1}".format("jshint", ' '.join(f for f in glob.glob(base_dir + "/*.js")))
+            jscscmd = "{0} {1} --reporter=inline".format("jscs", ' '.join(f for f in glob.glob(base_dir + "/*.js")))
+        else:
+            jshintcmd = "jshint " + base_dir + "/*.js"
+            jscscmd = "jscs " + base_dir + "/*.js --reporter=inline"
+    ret1 = os.system(jshintcmd)
+    print "Running JSCS"
+    ret2 = os.system(jscscmd)
+    print "Now running unit tests"
+    ret3 = rununits(p3=True, debug_mode=debug_mode)
+    return ret1 | ret2 | ret3
+
+
 def parse_time_args(argv):
     usageString = """
 
@@ -1067,7 +1090,13 @@ def shell(fn):
 
 
 def rununits(opt=False, p3=False, debug_mode=False):
-    testFiles = ['test/unit/'+f for f in os.listdir('test/unit') if '.py' in f]
+    if p3:
+        unit_dir = 'test/unit3'
+        p3on = 'true'
+    else:
+        unit_dir = 'test/unit'
+        p3on = 'false'
+    testFiles = [unit_dir + '/' + f for f in os.listdir(unit_dir) if '.py' in f]
     jstestengine = jsengine.replace('--debugger', '')
     passTot = 0
     failTot = 0
@@ -1076,10 +1105,6 @@ def rununits(opt=False, p3=False, debug_mode=False):
             os.mkdir("support/tmp")
         f = open("support/tmp/run.js", "w")
         modname = os.path.splitext(os.path.basename(fn))[0]
-        if p3:
-            p3on = 'true'
-        else:
-            p3on = 'false'
         f.write("""
 var input = read('%s');
 print('%s');
@@ -1308,6 +1333,8 @@ def main():
 
     if cmd == "test":
         test()
+    elif cmd == "test3":
+        test3()
     elif cmd == "testdebug":
         test(True)
     elif cmd == "dist":
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
 
     "build": "./skulpt.py dist",
     "build-min": "./skulpt.py dist -u",
-    "test": "./skulpt.py test"
+    "test": "./skulpt.py test",
+    "test3": "./skulpt.py test3"
   },
   "repository": {
     "type": "git",
--- a/test/unit3/exiting_module.py
+++ b/test/unit3/exiting_module.py
@@ -1,0 +1,5 @@
+__author__ = "leszek"
+
+if __name__ != "__main__":
+    # Exit when this module is imported
+    exit()
--- a/test/unit3/simpleMath.py
+++ b/test/unit3/simpleMath.py
@@ -1,0 +1,14 @@
+__author__ = 'millbr02'
+
+
+import unittest
+
+
+class SimpleMath(unittest.TestCase):
+    def testOne(self):
+        x = 2 + 2
+        self.assertEqual(x, 4)
+        self.assertEqual(type(x), int)
+
+if __name__ == '__main__':
+    unittest.main()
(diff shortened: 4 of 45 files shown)


meredydd:
I like the direction this is headed in. However, I don't like the copy-and-paste duplication. How much is currently in common between the v2 tests and the v3 tests?

If there's a lot in common, could I suggest we have  a`test-common/` directory or something for tests that should pass on both 2 and 3? Then we can start breaking out 2- and 3-specific tests only when necessary, rather than having to update every test twice when we make a generic fix.

rixner:
Unless/until we actually run CPython on the unit tests, I am wary of having common tests.  Someone will then write a test in a "common" test file that is correct only for Python 2 (or 3), but it will then be run as if it were correct on both.  Unfortunately, there are so many tests failing in CPython 2 right now, that we couldn't just move to a system where we validate the tests with CPython easily.

If others agree that we should have common tests, then I won't fight it too strongly, though.  But, I'd like to see us be less cavalier about the validity of tests that we accept into master moving forward, regardless.

rixner:
So, I just looked at test_list.py from your recent pull request.  All passes in CPython2.  One test fails in CPython3 (a comment notes the necessary change).  How would we handle a case like this?  There would be a common set of list tests in the common directory, and then a small file in each of the python 2 and python 3 unit test directories with different versions of that one test?  Requiring three list test files?

meredydd:
Oh, that's a pain! Yes, I agree that's a danger, and I agree that running the tests in CPython is The Right Thing To Do (for everything). I've just checked and most of the current tests do pass in CPython2, so it might be worth the effort of fixing them up. Perhaps we should do the CPython tests automatically, but not have them break the build yet.

To take the example of testing the `list` class, it's a pain however you slice it. If you keep the common code in one place, then you will indeed end up with three files to test one data type (common tests, v2-specific tests and v3-specific tests). If you copy-and-paste everything, then you still have two files to test one data type (v2 and v3), but there is an unknown amount of duplication between them. If you fix a bug that exists in both (as I just did), you'll end up having to modify two slightly-divergent sets of tests. Which, as you say, people will forget to do...

Ultimately, I think it's a question of whether we're going to continue supporting python2 for a substantial length of time. If we are, I think it's worth factoring out the common tests, to avoid the unholy mess described above. But if we're going to migrate steadily over to Python3, then we can basically abandon the Python2 tests in place, where they can rot gently and guard against regressions of things that used to work.

(Anvil will want to continue running our current users' code indefinitely, but we'd probably be on v3 already if Skulpt supported it. My guess is that Trinket are the same. The danger is having an intermediate state in which v2 starts rotting before v3 is shippable. On the basis that this transition will take longer than any of us expect, I therefore vote for factoring out the common tests, cause I anticipate we'll be maintaining them for a while.)

rixner:
I think both strategies have pros and cons.  They have different annoyances and advantages.  I'm not convinced your suggesting is really any better (nor do I believe it's really any worse).

I don't think the Python 2 tests will "rot", even if we split them up the way I am suggesting. Most of us still actively use Python 2 and will make sure the tests are kept up. Editing tests in two places is not necessarily a bad thing.  It forces you to think about Python 3 and how the behavior is the same/different. With either system, there is still the "danger" that we'll just write Python 2 (or 3) tests and ignore the other anyway. And editing two files is really no more effort than thinking about which of the three places your new test belongs (and you will have to write it twice anyway if Python 2 and 3 behave differently).

How do other people feel?

dsblank:
What about leaving them in one file, but marking them with a decorator?

https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures

albertjan:
Why don't we branch off, like cpython did? Having a compiler that will run py3 code or py2 code based on a flag is, not likely going to work for long after we start changing the compiler.

eah13:
tl;dr: we'll be moving to Python 3 over time but need to support legacy code, ideally with the same js file. Using CPython unittests as guideposts and potentially being able to selectively specify which should pass should be a boon and could be a feature of our dev toolkit. 

For Trinket's part, we've got a 'compatibility' mode that tries to run Python 2 & 3 code, defaulting to Python 3 where there's ambiguity (such as `/`l). This is important to us so that users can run code they find online without errors. They can engage strict 2 or 3 interpretation with the appropriate hashbang. The compatibility mode is in our fork and if others feel it's useful we could contribute it upstream. The hashbang detection is in our wrapper I believe but could also migrate up if others use it. 

Anyways, we have a large number of users who have Python 3 compatibility as a key requirement in choosing a tool. We've had to increase the visibility/documentation around the hashbang recently. 

Compatibility mode may get quite complex and going forward the design goal will be to run Python 3 code with a fallback to Python 2 syntax if possible instead of throwing errors if e.g. someone uses a print statement instead of a function. 

Compatibility mode also argues against a hard fork of Skulpt for Py3, unless that fork has compatibility mode. If we do hard fork we might lose the easy choice of 2/3 we can offer users now, but that's not a dealkiller if it paves the way for more complete Py3 support in the medium/long term. 

Compatibility mode aside, better Python 3 conformity for common educational modules like Turtle is a high priority for us. There are a number of things the module's just missing, and Python 3 added some cool new stuff. So medium term we'll spend some time there. 

Once @albertjan's compiler​ improvements are ready we'll be deploying those into our Python 3 strict mode and of course our compatibility mode should inherit them as well, with fallbacks for old style syntax where that's practical. 

Like Anvil, we'll need to support existing user code without breakage and this may get complex in places. Our upgrade path will probably be to one day flip over to strict Python 3 as the default with compatibility as an option. 

Long way of saying: we're definitely down with and will contribute to increasing Python 3 support. If no one else wants it, compatibility mode will live in our fork. 

I've long thought that a testrunner that could give us a sense of what CPython tests are passing is a good idea, so I support that heartily. What's more, a way of 'mandating' which unittests should pass would give us a way of ratcheting up completeness. So the process for implementing a new Py3 feature would be to specify in Skulpt's tests what output to expext from specific CPython unittests. Unsupported tests could still be run in a FYI mode, but over time more and more of Python's tests would become part of our CI suite. There's a not insignificant amount of overhead in scaffolding this and maintaining it, but a good portion of it could be automated. For instance, the 'FYI' mode of running tests could identify new passes and suggest adding them to the 'mandatory' set of passing tests/expected output we use for automated testing on github. 

My $0.02.

meredydd:
I'm busy today, so can't give this a proper treatment, but basically, yes, what Elliott said. We can't abandon our Python2 users and we don't currently have the bandwidth for a full migration, so a fork would be deeply unpleasant for us and we would be unable to contribute to the Python3 branch for the short term (which will end up being longer because we will have vital-for-Anvil fixes in the P2 branch and it all goes downhill from there).

@albertjan How big are the compiler changes going to be? Should we be dual-hosting compilers? How invasive is Trinket's "compat mode"?

meredydd:
Particularly +1 on @eah13's suggestion that we run CPython tests, with waivers that allow some files to fail, and then fix up those files and withdraw the waivers one by one.

rixner:
Even if the compiler changes are somewhat small and contained (I'm skeptical), they are not the only issue. You need to actually change the types to some degree.  Substituting ```__next__``` for ```next```, and so on.  While this all can be done, it's going to require some scaffolding to allow you to add/remove the right methods dynamically.  Furthermore, I know I harp on this a lot, but without more infrastructure, there will be a performance hit.  Dynamically deciding if you're going to return a list or a generator from some common functions, for instance.  To mitigate that, whole functions might need to be swapped in and out at runtime.  Again, this can be done, but it might quickly become unwieldy.

I also want to continue to support Python 2.  So, this is not me abandoning Python 2 and arguing to just drop everything and move to Python 3.  I don't, however, have the need or desire to dynamically switch back and forth with the same binary.  It seems easy enough to load two binaries and dynamically select which one to use if one wanted that capability...

I am opposed to the compatibility mode idea, simply because that doesn't exist in Python. It makes Skulpt no longer a subset of the language.  I am fine with non-standard libraries, but not non-standard language features.

Certainly, branching will complicate infrastructure updates and the addition of general language features that apply to both versions. They would have to be done twice, which probably won't happen. But, there are also a lot of benefits to branching in terms of not forcing the community to spend time building switching infrastructure that could be spent adding Python 3 features.

eah13:
I suppose Skulpt's relationship to Python 3 will be unavoidably complex, especially while community members are still Python 2 first.  In the meantime, we should continue to accommodate as many use cases as possible provided they can coexist and not degrade performance. As always, I'm very grateful for everyone's input and leadership on this!  Not something that anyone could address or solve in a vacuum.

@rixner a question for you (any anyone else for whom Python 2 support is key): when or under what circumstances would you move your project into using Python 3?  It will lose official PSF support [in 2020](http://legacy.python.org/dev/peps/pep-0373/) and is already on its final minor version.  If there is a date or criteria we can identify here, we can plan on having a relatively full alignment of requirements at that point, even if there's some divergence in the meantime. 

RE compatibility mode:  I'm totally OK if the community doesn't want it, but I don't think I correctly explained what this would be or mean *as a contributed feature*.  The feature contribution I envision would be an optional config setting, not the default, and wouldn't impact anyone's Skulpt deploys unless they specifically invoke it.  Trinket and others that want it could invoke the mode as needed and potentially set as default via a custom wrapper, just like those of use that implement non-alert based input or other config-level options. Just wanted to clarify the proposal if that influences anyone's feelings. The benefit to us would be simpler merges in both directions, which would benefit the community insofar as it would then be easier/faster for us to contribute upstream.  But of course we'll keep contributing and pulling in improvements regardless.

Due to the complexity of this issue, we could totally schedule another call if there's appetite and that seems most efficient.

As for this PR itself, I'm in favor of a merge unless there are any low hanging imrpovements anyone can think of.

rixner:
@eah13, I am actively using both Python 2 and Python 3 right now.  So, in that sense, I am moving to Python 3 immediately.  The question is when am I going to drop Python 2.  I suspect in 1-2 years (whether Skulpt as a project is ready or not).  By "drop", I mean stop actively updating.  I will probably leave a Python 2 snapshot in time up indefinitely.

I am not allowing dynamic switching between 2 and 3.  I have different deployments that set the python3 flag once and never touch it again.

rixner:
There does not seem to be a consensus, so I'm not sure what to do at this point.  My brief summary:

@rixner (and maybe @eah13): create a separate directory of Python3 unittests that are independent of Python2.

@meredydd: Refactor the tests into three directories, Python2 only, common, and Python 3 only.

@albertjan: create a Python3 branch

I'm sure this is an over-simplification of our positions, and I suspect we all have some flexibility.  However, I think @meredydd and @eah13 are both definitely opposed to the Python3 branch idea?

@bnmnetp, do you have an opinion on this?

At this point, I would like to make forward progress on Python3, but don't want to structure things in a way that forces me to redo/refactor a lot of stuff later to get everyone to agree to accept it into Skulpt. So, I'd really appreciate some sort of conclusion here - regardless of what it is. My personal order of preference is:

1. Independent Python3 test directory.
2. Python3 branch.
3. Three directory structure with Py2/common/Py3 tests.

Mostly, I have 3 ranked where it is because I think it's a lot of busy work for a short term goal that is going to be more trouble than it's worth in the long term.  But, I will do any of the three (or something else) that we can all agree on, as my opinion on this is not set in stone or anything.

albertjan:
Right lets nog get hung up on this :). We can merge this. That leaves @merredyd's option on the table. I think I misjudged your scope I think what I meant is that if we do a compiler update we won't be able to keep the old compiler/parser infrastructure around. I've been looking into that path. For testing python with the py3 flag on this is fine. 

I second adding a step to the build where we run the tests on cpython.

rixner:
We're going to have to replace the compiler at some point, though, right? Does that really mean they can't co-exist?  Doesn't it just mean that we shouldn't try to make the same compiler do both?  But, rather have both compilers in Skulpt and dynamically select which one to use based on the python3 flag?

Again, I don't care if Py2 and Py3 diverge, but others want them to peacefully co-exist, so I think that's potentially a viable option to make us all happy.  We may potentially want to add an Sk3 object or something to hold all the Python3-only infrastructure so the namespaces don't conflict...

albertjan:
I guess you are right, you could have a seperate tokenizer - parser - compiler for py3 in skulpt thats turned on with a flag.

albertjan:
@meredydd the compiler changes are invasive enough to make it very difficult to merge them and do flags the ast is pretty different. The compat mode is just a specify hash of option per py3 setting, instead of doing all of them in one bool. I would say something that would be nice upstream, because it paves the way for `from __future__ import bla`

rixner:
The new batch of commits add more tests and comment out failing tests (presumably to be uncommented as we fix the feature).  So, now running ```skulpt.py test3``` should pass.

j-f1:
@rixner Who’s @sgj1?

rixner:
@sgj1 is a student who works for me.  I have had several students who have worked with me over the years on Skulpt. I vet their contributions before creating pull requests with their work.  I am implicitly taking responsibility for the work along with them, while they get proper attribution for their work.  Why would that matter?

albertjan:
I don't think @j-f1 was worried about the quality of work :) and was just curious. @rixner's students have been very valuable for skulpt.

j-f1:
@rixner Cool, just curious 😀

albertjan:
@rixner more changes incoming here or are we ready to do a review/merge?

rixner:
There are no more updates coming at this point.

skulpt.py
```diff
@@ -240,6 +240,29 @@ def test(debug_mode=False):
         ret4 = rununits(debug_mode=debug_mode)
     return ret1 | ret2 | ret3 | ret4
 
+def test3(debug_mode=False):
```
albertjan: Couldn't this be a flag on the `test` function?

rixner: Yes.  It was unclear if they would diverge further in the future, but I guess there is a lot of repeated code this way.

albertjan:
Alright I'm good with this to get merged @skulpt/committers have at it.

rixner:
@skulpt/committers, can we get this merged or decide on an alternative structure?  Thanks!

bnmnetp:
I say merge.

rixner:
Can I just push the merge button then?

bnmnetp:
Feel free

albertjan:
Just merged it! 😄

PR #691, opened 2017-06-13
Implement explicit dunder-functions in list class

meredydd:
We didn't have explicit dunder-functions (eg `__setitem__`/`__getitem__`) for the list class. This made subclassing it a pain. Now we do.

### DIFF
--- a/src/list.js
+++ b/src/list.js
@@ -293,6 +293,11 @@ Sk.builtin.list.prototype.sq$contains = function (item) {
     }
     return false;
 };
+
+Sk.builtin.list.prototype.__contains__ = function(self, item) {
+    return Sk.builtin.list.prototype.sq$contains.call(self, item);
+};
+
 /*
  Sk.builtin.list.prototype.sq$inplace_concat = list_inplace_concat;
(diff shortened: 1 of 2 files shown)


src/list.js
```diff
@@ -293,6 +293,11 @@ Sk.builtin.list.prototype.sq$contains = function (item) {
     }
     return false;
 };
+
+Sk.builtin.list.prototype.__contains__ = function(self, item) {
```
albertjan: shouldn't this also be a `Sk.builtin.func`?

meredydd:
Right you are, both times. I feel like I owe you a drink for each stupid mistake of mine you spot in code review, but at this rate that would get messy quickly...

src/list.js
meredydd: Right you are, both times. I feel like I owe you a drink for each stupid mistake of mine you spot in code review, but at this rate that would get messy quickly...

albertjan: haha no worries. thats what these things are for right 😄 and I think if you'd look into what I do I soon be giving you beer instead

meredydd: Right. Fixed now. Are we good to merge?

PR #693, opened 2017-06-26
Platform Module

rixner:
This pull request adds a rudimentary platform module.  Perhaps most interestingly ```platform.python_implementation()``` returns ```"Skulpt"```, allowing you to check if you are running in Skulpt or another interpreter.

Also, a ```brun3``` command is added to skulpt.py to run Python 3 in the browser.

The Python versions are arbitrarily set to 2.7.0 and 3.2.0 for ```platform.version()```.  Not sure if those are the best values or not.

### DIFF
--- a/src/lib/platform.js
+++ b/src/lib/platform.js
@@ -1,0 +1,72 @@
+var $builtinmodule = function(name){
+    var mod = {};
+    var inBrowser = (typeof window != "undefined") && (typeof window.navigator != "undefined");
+    mod.python_implementation = new Sk.builtin.func(function() {
+        // This checks that the correct number of arguments were passed
+        Sk.builtin.pyCheckArgs("python_implementation", arguments, 0, 0);
+        return new Sk.builtin.str("Skulpt");
+    })
+    mod.node = new Sk.builtin.func(function() {
+        Sk.builtin.pyCheckArgs("node", arguments, 0, 0);
+        return new Sk.builtin.str("");
+    })
+    mod.version = new Sk.builtin.func(function() {
+        var vers;
+        Sk.builtin.pyCheckArgs("version", arguments, 0, 0);
+        if (Sk.python3) {
+            vers = "3.2.0";
+        }
+        else {
+            vers = "2.7.0";
+        }
+        return new Sk.builtin.str(vers);
+    })
+    mod.system = new Sk.builtin.func(function() {
+        var sys;
+        Sk.builtin.pyCheckArgs("system", arguments, 0, 0);
+        if (inBrowser) {
+            sys = window.navigator.appCodeName;
+        }
+        else {
+            sys = "";
+        }
+        return new Sk.builtin.str(sys);
+    })
+    mod.machine = new Sk.builtin.func(function() {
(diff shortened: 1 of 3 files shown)


src/lib/platform.js
```diff
@@ -0,0 +1,72 @@
+var $builtinmodule = function(name){
+    var mod = {};
+    var inBrowser = (typeof window != "undefined") && (typeof window.navigator != "undefined");
+    mod.python_implementation = new Sk.builtin.func(function() {
+        // This checks that the correct number of arguments were passed
+        Sk.builtin.pyCheckArgs("python_implementation", arguments, 0, 0);
+        return new Sk.builtin.str("Skulpt");
+    })
```
albertjan: we need some `;` for these 😄

src/lib/platform.js
```diff
@@ -0,0 +1,72 @@
...
+    mod.node = new Sk.builtin.func(function() {
+        Sk.builtin.pyCheckArgs("node", arguments, 0, 0);
+        return new Sk.builtin.str("");
+    })
+    mod.version = new Sk.builtin.func(function() {
+        var vers;
+        Sk.builtin.pyCheckArgs("version", arguments, 0, 0);
+        if (Sk.python3) {
+            vers = "3.2.0";
```
albertjan: why did you choose 3.2?

src/lib/platform.js
```diff
@@ -0,0 +1,72 @@
...
+            appVers = window.navigator.appVersion;
+        }
+        else {
+            appVers = "";
+        }
+        return new Sk.builtin.str(appVers);
+    })
+    mod.architecture = new Sk.builtin.func(function() {
+        Sk.builtin.pyCheckArgs("architecture", arguments, 0, 0);
+        return new Sk.builtin.str("('64bit', '')");
```
albertjan: is there any reason to not say 53bit here? because our long/int split is there?

src/lib/platform.js
rixner: Because 3.3 introduced Syntax changes that we don't have (like yield).  Perhaps 3.0.0 is a better choice, though...  Not sure.

albertjan: No no that makes sense. :). I didn't know.

src/lib/platform.js
rixner: That number is the bit width supported by the hardware for conventional python implementations.  If we consider the Javascript engine to be the "hardware" here, then its bit width is 64...  Floats are still 64 bits, which is possibly more important to know if you are querying that number.  (Since ints and longs will still work properly regardless.)

albertjan: Alright lets keep it at 64bit I guess that if students get to this corner they're about ready to graduate to cpython ;)

albertjan: Alright lets keep it at 64bit I guess that if students get to this corner they're about ready to graduate to cpython ;)

albertjan:
LGTM @skulpt/committers merge if you agree

bnmnetp:
I agree.

PR #694, opened 2017-06-26
Super

albertjan:
Well here we go, a first pass at super, I've probably forgotten some stuff here and there so a proper review would be greatly appreciated. 

This implements bits of the `super` type:

- python 2 most used version: `super(Type, self)` gives a proxy to the first parent of `Type`
- python 3 like python 2 `super()` where `Type` and `self` are inferred from the scope. (not entirely the same implementation as py3 but I don't think that should stop us now

What doesn't work:

- `super(Type)` to get an unbound super you can attach to an object and use descriptors
- `class superduper(super): pass` this will actually compile but because `tp$getattr` doesn't get inherited it doesn't work as a super class.
- `super(Type, OtherType)` might be easy to fix but haven't looked into it
- multiple inheritance, I just get the first type from the MRO 

What I also fixed:

- nested classes with `$cell` variables
- decorators in nested classes


j-f1:
You might want to change the author email on the first two commits to your `@trinket.io` one so they’re associated with your GitHub account, or you could [add](https://github.com/settings/emails#new_user_email) the one you used there (`@curit.com`) to your GitHub account.

albertjan:
Oh wank, yeah I'll look into it.

albertjan:
Ohh I'm seeing double now 😄

j-f1:
Now your committer emails are `@skimlinks.com` 😀

albertjan:
But thats in my account on GH 😄

meredydd:
Just chiming in to say that I've seen this - you just posted the PR as I was finishing up trip to the US (got home yesterday), so I will need a moment to come up to speed!

eah13:
Special thanks to @rik-cross and @lawsie for requesting this and to the Raspberry Pi Foundation for funding it.  And, of course, a huge thanks to @albertjan for doing it in such a way as to make it easily contribute-able upstream.

Rik and Laura have been testing this along the way and this is now in production on trinket.  Code examples they used:

https://trinket.io/python/1723cfac72
https://trinket.io/python/aea629feef 
https://trinket.io/python/13973c0f94

Bonus, though it was outside of the original scope: accessing a class variable of a super() class is also supported:

https://trinket.io/python/19d90c64cb 

Known limitations: Multiple inheritance is still a no-go

https://trinket.io/python/a4acf1a8a3

The above should all be testable in vanilla skulpt by collapsing the code into a single file, an exercise left to the reviewer :)

Should enable some neat object oriented coding once it's merged!

eah13:
Todo: Looks like `object` doesn't have a dunder init:

https://trinket.io/python/2f18037e30

This probably won't affect much user code but test code like this needs it :)

bnmnetp:
Any chance of finishing this off and making it merge ready?  That would seem to be a good place to do a release.   Along with a fix for #697 ?

albertjan:
Yeah I was planning to look into that one this evening time permitting, I can rebase and clean this up so we can get this merged too. Might be tomorrow though 😄

bnmnetp:
Awesome! Thanks.

albertjan:
@bnmnetp this is ready

albertjan:
But is hasn't been reviewed yet

src/compile.js
```diff
@@ -525,6 +525,15 @@ Compiler.prototype.ccall = function (e) {
     }
     else {
         out ("$ret;"); // This forces a failure if $ret isn't defined
+        if (Sk.python3 && e.func.id && e.func.id.v === "super" && args.length == 0) {
+            // make sure there is a self variable
+            // note that it's part of the js API spec: https://developer.mozilla.org/en/docs/Web/API/Window/self
+            // so we should probably add self to the mangling
+            // TODO: feel free to ignore the above
+            out("if (typeof self === \"undefined\" || self.toString().indexOf(\"Window\") > 0) { throw new Sk.builtin.RuntimeError(\"super(): no arguments\") };")
```
rixner: This is ugly!  It really seems like the right solution is to start mangling self.

src/compile.js
```diff
@@ -525,6 +525,15 @@ Compiler.prototype.ccall = function (e) {
...
+            args.push("self.ob$type");
```
rixner: I think CPython uses the compile-time context to see the name of the class this is being called from within, not using the run-time type of self.  I can't immediately think of a reason why this would result in different behavior, though.  And it doesn't seem like we have enough context in the compiler at this location.  But, is this definitely equivalent?

src/typeobject.js
```diff
@@ -0,0 +1,148 @@
...
+
+Sk.abstr.setUpInheritance("super", Sk.builtin.superbi, Sk.builtin.object);
+
+/**
+ * Get an attribute
+ * @param {string} name JS name of the attribute
+ * @param {boolean=} canSuspend Can we return a suspension?
+ * @return {undefined}
+ */
+Sk.builtin.superbi.prototype.tp$getattr = function (name, canSuspend) {
```
rixner: This is maybe the third place this code appears (object, type, here).  The type one is slightly different, but not by much.  Is it time to refactor things to create a generic get attr that's called from all of these places?

src/typeobject.js
```diff
@@ -0,0 +1,148 @@
...
+        /* a is not completely initilized yet; follow tp_base */
+        return Sk.builtin.type_is_subtype_base_chain(a, b);
+    }
+}
+
+/**
+ * @constructor
+ * Sk.builtin.superbi
+ */
+Sk.builtin.superbi = function superbi (a_type, self) {
```
albertjan: it was already called `superbi`

src/typeobject.js
albertjan: Yeah, that might be a good idea, I'll see how cpython does this, and see if we can do the same.

src/compile.js
```diff
@@ -525,6 +525,15 @@ Compiler.prototype.ccall = function (e) {
     }
     else {
         out ("$ret;"); // This forces a failure if $ret isn't defined
+        if (Sk.python3 && e.func.id && e.func.id.v === "super" && args.length == 0) {
```
albertjan: It's not the same as cptyhon, cpyhton get the first variable of the stack frame when calling `super`. But we don't have stackframes, and javascript doesn't allow us to inspect the scope. I saw no other obvious way of solving this. The only caveat with this solution is that the first variable of the function needs to be called `self`. I did this because I have no way of determining if the `super` function is called from a bound or unbound function.

src/compile.js
albertjan: It looks for `__class__` in the current `cell`. I figured this was close enough 😄 .

src/compile.js
albertjan: I think so too.

src/typeobject.js
rixner: Like I said, I wasn't blaming you for this, I know it was already called ```superbi```.  I was just commenting that maybe we should fix it now since you're fixing it up anyway.  This is not a deal breaker for me, but it does seem like it would be nice to move things towards consistency.  Your call, as far as I'm concerned.

src/compile.js
rixner: Fair enough.  Can we add a clear comment here, then, about the limitation that it will not work if the Python code violates the ```self``` convention?

src/compile.js
rixner: Ah, that makes sense.   Agreed this is good, then!

src/compile.js
rixner: It's a pretty quick and easy change to do so, right?  Unless you run into an issue doing so, I'd like to see that as part of this pull request...

rixner:
@bnmnetp what program did you get the call stack size exceeded error on?  I originally tried some simple examples and it worked properly for those.  So, perhaps you managed to find a broken edge case?

bnmnetp:
@rixner, @albertjan   Here's a simple example that shows that it breaks when you have multiple levels of inheritance.

```
class LogicGate:

    def __init__(self,n):
        self.name = n
        self.output = None



class BinaryGate(LogicGate):

    def __init__(self,n):
        super().__init__(n)

        self.pinA = None
        self.pinB = None



class AndGate(BinaryGate):

    def __init__(self,n):
        super().__init__(n)



b1 = BinaryGate("foo")
print("binary gate constructed")
g1 = AndGate("G1")

```

The BinaryGate is constructed fine. but I get the stack size error when I go to make the AndGate.

rixner:
@albertjan, I think that @bnmnetp just showed us the case where ```self.ob$type``` is not the same as ```__class__```. :smile:

albertjan:
I'll address this tonight if I'm allowed computer time by my girlfriend.

albertjan:
@rixner Hah yeah you're right... because we pass in `self` from the child `self.ob$type` isn't the type of the current class. Good catch @bnmnetp. I'll have to figure out how to do this.

Dunes:
I had a quick experiment with your implementation. You also have another problem with your implementation. Wrote a short test to demonstrate (both py2 and py3 compatible).

https://trinket.io/python/9671104043

```python
#!/bin/python3

from unittest import TestCase, main

class TestSuper(TestCase):

    def test_works_with_diamond_inheritance(self):
        # given 
        class A(object):
            def f(self):
                return 'A.f'

        class B(object):
            pass

        class C(B, A):
            def f(self):
                # super should inspect all classes after current in the mro
                # looks like super only currently inspects B and object
                return super(C, self).f()
        
        # then
        self.assertIn(A, C.__mro__) # success
        self.assertEqual('A.f', C().f()) # fails with attribute error

if __name__ == "__main__":
    main()
```

albertjan:
Hey @Dunes thanks, I did know about this limitation. It shouldn't be too hard to implement in skulpt but I felt to get this to work for most people quickly leaving out multiple inheritance would be the first thing to go.

src/typeobject.js
albertjan: I'd like @meredydd 's input on this one as he's the resident `getattr` specialist. cpython has a special implementation for this in super, but that's probably because it's more difficult with multiple inheritance and `mro` walking etc.

src/compile.js
albertjan: I couldn't figure out how python3 got that `__class__` variable, so I figured I'd venture out and ask a stack overflow question see: https://stackoverflow.com/a/45522873/117242 

I'm now trying to figure out how to walk the ast from `cclass` to annotate it where we need the `__class__` variable inserted into the scope of the function call.

albertjan:
With some guidance from @meredydd over the weekend at pycon uk, we found a solution to the `__class__` cell in python3.

ebertmi:
@albertjan When is this ready to merge?

albertjan:
As far as I’m concerned yeah :)

bnmnetp:
I would love to see this merged.

sheeeeep:
Is there any progress with this PR?

bnmnetp:
OK,  @rixner are you OK with merging this one to clear the way for @albertjan to do some merging back from the Trinket fork?

albertjan:
I'm going to drink a beer to celebrate this 🍺

bnmnetp:
I almost sent it back because when I ran the tests one test in time_tests.py failed.  

It turns out that there is a bug, but not related to this PR, and fairly obscure I think.

I am on an island in the Atlantic Time Zone that does NOT observe daylight savings time, this caused test_strftime to be off by an hour.  I just moved myself into EDT and all was well.

albertjan:
Ohhh ugh no. I'm not surprised there are issues with the str time thing 😄

rixner:
I was on a computer-free vacation for a week.  I've been happy to merge this since the ```__class__``` issue was resolved.  I don't know how it feel off the radar for so long!  Sorry @albertjan!

PR #696, opened 2017-07-11
Implement the `with` statement

meredydd:
This PR supports the `with` statement (PEP 343) in Skulpt. Context managers ahoy!

Many thanks to @albertjan for the parser patch that made this possible.

(This PR builds on my #680 `try`/`finally` implementation, as `with` uses many of the same mechanisms)

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -1245,7 +1245,6 @@ Compiler.prototype.ctryexcept = function (s) {
             this.vexpr(handler.name, "$err");
         }
 
-        // Need to execute finally before leaving body if an exception is raised
         this.vseqstmt(handler.body);
 
         // Should jump to finally, but finally is not implemented yet
@@ -1259,15 +1258,36 @@ Compiler.prototype.ctryexcept = function (s) {
 
     this.setBlock(orelse);
     this.vseqstmt(s.orelse);
-    // Should jump to finally, but finally is not implemented yet
     this._jump(end);
+
     this.setBlock(end);
 };
 
 Compiler.prototype.ctryfinally = function (s) {
-    out("/*todo; tryfinally*/");
-    // everything but the finally?
-    this.ctryexcept(s.body[0]);
+
+    var finalBody = this.newBlock("finalbody");
+    var exceptionHandler = this.newBlock("finalexh")
+    var exceptionToReRaise = this._gr("finally_reraise", "undefined");
+    this.u.tempsToSave.push(exceptionToReRaise);
+
+    this.setupExcept(exceptionHandler);
+    this.vseqstmt(s.body);
+    this.endExcept();
+    // Normal execution falls through to finally body
+    this._jump(finalBody);
+
+    this.setBlock(exceptionHandler);
+    // Exception handling also goes to the finally body,
+    // stashing the original exception to re-raise
+    out(exceptionToReRaise,"=$err;");
+    this._jump(finalBody);
+
+    this.setBlock(finalBody);
+    this.vseqstmt(s.finalbody);
+    // If finalbody executes normally, AND we have an exception
+    // to re-raise, we raise it.
+    out("if(",exceptionToReRaise,"!==undefined) { throw ",exceptionToReRaise,";}");
+    // Else, we continue from here.
 };
 
 Compiler.prototype.cassert = function (s) {
--- a/src/compile.js
+++ b/src/compile.js
@@ -1271,34 +1271,9 @@ Compiler.prototype.ctryexcept = function (s) {
     this.setBlock(end);
 };
 
-Compiler.prototype.ctryfinally = function (s) {
-
-    var finalBody = this.newBlock("finalbody");
-    var exceptionHandler = this.newBlock("finalexh")
-    var exceptionToReRaise = this._gr("finally_reraise", "undefined");
-    var thisFinally, nextFinally;
-    this.u.tempsToSave.push(exceptionToReRaise);
-
-    this.pushFinallyBlock(finalBody);
-    thisFinally = this.peekFinallyBlock();
-    this.setupExcept(exceptionHandler);
-    this.vseqstmt(s.body);
-    this.endExcept();
-    // Normal execution falls through to finally body
-    this._jump(finalBody);
-
-    this.setBlock(exceptionHandler);
-    // Exception handling also goes to the finally body,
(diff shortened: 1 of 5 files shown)


src/errors.js
```diff
@@ -536,4 +536,15 @@ Sk.abstr.setUpInheritance("StopIteration", Sk.builtin.StopIteration, Sk.builtin.
 goog.exportSymbol("Sk.builtin.StopIteration", Sk.builtin.StopIteration);
 
 
+// TODO: Extract into sys.exc_info(). Work out how the heck
+// to find out what exceptions are being processed by parent stack frames...
```
albertjan: Yeah lets not worry about this too much, fat chance we'll ever have stack frames. 😄

src/compile.js
```diff
@@ -1259,15 +1266,136 @@ Compiler.prototype.ctryexcept = function (s) {
...
+    } else {
+        // If this was a nonlocal return, we need to work out where to go.
+        // If it's a return, we always continue the finally chain.
+        // If it's a break or continue, we need to work out whether our parent
+        // finally-block is at the same break/continue depth as us (ie whether there's
+        // a loop/breakable-block boundary between us and it).
+
+        nextFinally = this.peekFinallyBlock();
+        
+        out("if($postfinally!==undefined) {",
```
albertjan: You've been able to break things up in to quite understandable bits, but this thing... 😄  did you find something in the python docs or code that described this? 

The nested `continue/break` block that are out of sync that is the definition of an edge case if there ever was one 👍

src/compile.js
```diff
@@ -1259,15 +1266,136 @@ Compiler.prototype.ctryexcept = function (s) {
...
+                    (nextFinally.breakDepth == thisFinally.breakDepth) ? "|| $postfinally.isBreak" : "",
+                    (nextFinally.continueDepth == thisFinally.continueDepth) ? "|| $postfinally.isContinue" : "", ") {",
+
+                        "$blk=",nextFinally.blk,";continue;",
+                "} else {",
+                    "$blk=$postfinally.gotoBlock;$postfinally=undefined;continue;",
+                "}",
+            "}");
+    }
+}
```
albertjan: we need a `;` here

src/compile.js
```diff
@@ -1888,13 +2016,33 @@ Compiler.prototype.cclass = function (s) {
...
+    if (this.u.breakBlocks.length === 0) {
+        throw new SyntaxError("'break' outside loop");
+    }
+    gotoBlock = this.u.breakBlocks[this.u.breakBlocks.length - 1];
+    if (nextFinally && nextFinally.breakDepth == this.u.breakBlocks.length) {
+        out("$postfinally={isBreak:true,gotoBlock:",gotoBlock,"};");
+    } else {
+        this._jump(gotoBlock);
+    }
+}
```
albertjan: and here 😄

albertjan:
Dude this is awesome. I've reviewed both the `try/finally` and the `with` functionality and I can't find fault with it. Other then a `repr` issue is with Errors but thats for another PR.

albertjan:
@skulpt/committers This is one of the big things! Lets get on this. I've tested with this script: 

```
class Banana(object):
    def __enter__(self):
        print "enter"

    def __exit__(self, *args):
        print "exit", args

def test():
    with Banana():
        print "inside"
        raise ValueError()

try:
    print "res", test()
except:
    print "errrrr"

def plop():
    count = 0
    while count < 10:
        count += 1
        try:
            print "try"
            if count > 7:
                print "break"
                break
        finally:
            if count > 7:
                return 5

print plop()

print repr(ValueError())
```

src/errors.js
meredydd: Indeed not. My guess is a stack of global exception state, populated by the `catch` handle, which `sys.exc_info()` can then read. One for another PR.

src/compile.js
meredydd: This is the best I've got: https://docs.python.org/2.7/reference/compound_stmts.html#finally

Relevant section:
> When a return, break or continue statement is executed in the try suite of a try...finally statement, the finally clause is also executed ‘on the way out.’

The challenge is executing a `break` or `continue` in a nested `finally` block. What I'm doing here is working out whether the  *which* `finally` blocks need to execute for a given `break` (or `continue`). I compare the `breakDepth` of the inner and outer `try` blocks (that is, how many nested `break`able blocks surround them). If the number is the same, then the `break` will always go through both `finally` blocks. If they're different, we only need to run the inner one before resuming.

(Aside: Looks like you're right - `break` and `continue` contexts are always exactly the same. They're tracked separately in `compile.js`, so I had a brainfart and just assumed Python had a `switch` statement or something else `break`able but not `continue`able. So I suppose we could get away with only tracking one of them.)

src/compile.js
meredydd: done

src/compile.js
meredydd: done

src/compile.js
meredydd: (Updated to match the fact that `break` and `continue` depths are always the same: I now just track `breakDepth`, and I no longer distinguish between `isBreak` and `isContinue` in `$postfinally`.)

meredydd:
I've pushed an update addressing your comments (including improved comments on that tricky control-flow section).

albertjan:
I'm going to merge this too @rixner because there are fixes in this for try/finally. Unless you have serious doubts.

rixner:
Can I take a careful look first?

albertjan:
Yeah yeah ofcourse! :)

albertjan:
Awesome! 😄

ludwiktrammer:
The most recent version on NPM doesn't include this, does it? Is there a roadmap to when the new release might appear?

albertjan:
@ludwiktrammer we don't actually own the package on npmjs.com @daluege does, who ever he may be. Skulpt also doesn't really play well with node (not yet atleast).

We definitely need to do a release.

But while we're getting around to it, you can always get the latest built version of skulpt from here: https://github.com/skulpt/skulpt-dist

j-f1:
If you’re interested, you could get yourself the `@skulpt` namespace so you could publish packages under `@skulpt/*`.


> Skulpt also doesn't really play well with node (not yet atleast).

npm is kind of a misnomer — it’s used for both Node and the browser.

Issue #697, opened 2017-07-18
Turtle fails with internal error when not using named turtle


deansheather:
Using the following turtle code yields an internal error. Tested using the `simpleskulpt.html` file on the project homepage. Naming the turtle (using the default code in `simpleskulpt.html`) doesn't seem to produce this problem. I copied the error from Firefox console, I apologize for the formatting.

This code previously worked on Trinket.io yesterday.

```py
from turtle import *
forward(100)

print "Hello World" 

```

```
wrapped method failed  skulpt.min.js%20line%20994%20%3E%20eval:2082:21

wrapperFn@http://www.skulpt.org/static/skulpt.min.js line 994 > eval:2079:17
Sk.builtin.func.prototype.tp$call@http://www.skulpt.org/static/skulpt.min.js:155:366
Sk.misceval.applyOrSuspend@http://www.skulpt.org/static/skulpt.min.js:272:391
Sk.misceval.callsimOrSuspend@http://www.skulpt.org/static/skulpt.min.js:263:93
$scope0@http://www.skulpt.org/static/skulpt.min.js line 994 > eval:20:104
@http://www.skulpt.org/static/skulpt.min.js line 994 > eval:39:1
x@http://www.skulpt.org/static/skulpt.min.js:994:373
Sk.importModuleInternal_@http://www.skulpt.org/static/skulpt.min.js:993:346
Sk.importMainWithBody@http://www.skulpt.org/static/skulpt.min.js:996:358
runit/myPromise<@file:///home/dean/Downloads/simpleskulpt.html:37:15
Sk.misceval.asyncToPromise/<@http://www.skulpt.org/static/skulpt.min.js:265:41
Sk.misceval.asyncToPromise@http://www.skulpt.org/static/skulpt.min.js:264:49
runit@file:///home/dean/Downloads/simpleskulpt.html:36:20
onclick@file:///home/dean/Downloads/simpleskulpt.html:1:1
  skulpt.min.js%20line%20994%20%3E%20eval:2083:21

ExternalError: TypeError: instance[method] is undefined on line 2  simpleskulpt.html:43:8

```

eah13:
[...]

PR #698, opened 2017-07-18
skulpt function should be called with Sk.misceval.callsim

albertjan:
This should address #697

### DIFF
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -1497,7 +1497,7 @@ function generateTurtleModule(_target) {
 
     function ensureAnonymous() {
         if (!_anonymousTurtle) {
-            _anonymousTurtle = _module.Turtle();
+            _anonymousTurtle = Sk.misceval.callsim(_module.Turtle);
         }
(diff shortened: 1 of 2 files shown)


bnmnetp:
Great, I'll pull this one and the super PR onto my test branch and give them a whirl.

bnmnetp:
Yep, this fixes the problem, and doesn't break anything else AFAIK.

This one should be an easy accept.

albertjan:
i don't think this code is used from anywhere else

meredydd:
LGTM by inspection. Merging.

meredydd:
FYI @albertjan, the canonical GitHub way of doing a pull request that closes an issue is to use a magic phrase like `fixes #697` or `closes #697` in a commit message. Then, when-and-only-when that commit is merged into master (ie when the PR is accepted), the issue will be closed with a link to the PR and commit responsible.

This avoids prematurely closing issues while the fix is still unmerged, while still making sure you don't have to remember it yourself!

PR #700, opened 2017-07-20
Support constructing a dict with kwargs

meredydd:
This change supports supplying keyword arguments to the `dict()` constructor.

By defining a `tp$call` slot on the `Sk.builtin.dict()` function, I could separate out the Python and JS call paths, and add kwarg handling neatly.


bnmnetp:
Nice!

LGTM

PR #703, opened 2017-07-26
Various Python 3 updates

rixner:
This pull request incorporates fixes for Python 3 in various places.

The builtins oct, pow, len, round, and format are updated.

There are fixes/improvements to ```__contains__``` and ```__format__```.

There is also some infrastructure added to switch method names between Python 2 and Python 3.


rixner:
Anyone?

PR #704, opened 2017-07-28
Add operators &,|,-,^ to set type

srdjan-bozovic:
Added operators equivalent to functions union, intersection, difference, symmetric_difference as described in https://docs.python.org/2/library/sets.html


rixner:
Can you please add some unit tests for this as well? (It probably makes the most sense to put them in test/unit/test_set.py)

PR #708, opened 2017-08-02
Implement object.__getattribute__; tidy up attribute getting

meredydd:
Skulpt's implementation of `object.__getattr__` was incorrect (`object.__getattr__` does not exist in CPython).

In CPython, `object.__getattribute__` is the default implementation of the "get attribute" slot (`tp$getattr` in Skulpt), and it falls back to `__getattr__` if it doesn't find anything. I have updated the Skulpt implementation to do the same thing.

This means that the `type.js` implementation of `tp$getattr` just got a whole lot simpler - and we can actually use `object.__getattribute__` from Python code now.

In addition, `object.__getattribute__` now actually throws an `AttributeError` instead of returning `undefined` (which is not a legal Python value, and confuses the heck out of calling code) - which was the original bug that prompted this change. The yak hair is piled deep around my ankles this afternoon ;)

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -1049,26 +1049,7 @@ Sk.builtin.hasattr = function hasattr (obj, attr) {
         if (obj.tp$getattr(attr.v)) {
             return Sk.builtin.bool.true$;
         } else {
-            special = Sk.abstr.lookupSpecial(obj, "__getattr__");
-            if (special) {
-                ret = Sk.misceval.tryCatch(function () {
-                    var val = Sk.misceval.callsim(special, obj, attr);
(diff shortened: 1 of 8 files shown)


albertjan:
It's amazing how something as simple as getting an attribute, needs this much work 😄. Great job @meredydd. Thank you for going all the way down the rabbit hole to fix this.

meredydd:
I've just submitted #726 on top of this; ideally this would get merged first.

albertjan:
@skulpt/committers Lets get these merged! 😄

ebertmi:
This can be merged.

PR #709, opened 2017-08-03
Re.compile mapped to existing re functions

### LINKED ISSUE #705: spoofing re.compile
Had a short chat with Elliott Hauser, who recommended I share this idea with the community.

The support for "re" is somewhat broken right now - specifically re.compile doesn't exist.

Elliot's solution, which I think is a great idea, would be to spoof re.compile - i.e. delivering no runtime performance optimization whatsoever, but it wouldn't break code that calls .compile either.

Please feel free to contact me if you have questions about the idea.

eah13:
This includes:

* `re.compile`
* `RegexObject` and the methods I could reuse code for: `.findall()`, `.search()`, `.split()`, and `.match()`.  
* `re.purge()` got a stub so we can run code that uses it but we don't cache so no need to purge
* A bunch of CPython tests.  These led me to do a few random things like `__str__` and `__repr__`

As I culled stuff from the CPython file I left failing tests behind a `if include_failing` statement if they looked like things we'd get to the next time someone looks at re.  Seemed like a good thing to do but I can scrap this if the consensus is that it's messy.

Made some style changes to the file here and there, trying to follow our guidelines. Hoisted vars, double quotes, semicolon hygiene, etc. LMK if anything seems amiss.  I'm not a Javascript native.

I left plenty of stuff out of scope in the interests of moving quickly. For instance, you'll see todos for `pos` or `endpos` arguments.  ~~Another big one is support for named arguments in the methods.  They work when placed in the correct order but not when explicitly named like `flags=re.I`.  My initial attempts to do them the same way I did for the functions didn't work so I bailed.~~ Edit: latest commit fixes this.  I also couldn't figure out how to get the `.flags` and `.pattern` attributes set up when the RegexObject gets instantiated.  I've never had to do either of these in Skulpt before so any pointers or a link to an example in the codebase would be welcome.

Someone, someday will implement `re.sub`.  I'm not that one and/or this is not the day.

Fixes #705


src/lib/re.js
```diff
@@ -30,137 +32,129 @@ var $builtinmodule = function (name) {
...
 
         return new Sk.builtin.list(result);
-    });
+    };
+
+    _split.co_varnames = ["maxsplit", "flags"];
+    _split.co_numargs = 4;
+    _split.$defaults = [ new Sk.builtin.int_(0), new Sk.builtin.int_(0) ];
+
+    mod.split = Sk.nativejs.func(_split);
```
albertjan: I've literally never seen this `Sk.nativejs.func` before. 😄 Any idea what it does?

albertjan:
Looks good to me btw. Nice to see another bunch of cpython tests come in.

src/lib/re.js
eah13: Nope.  It was there so I left it.  Maybe that's related to the problems I was having with named args.

eg. 

```
print(w.split('Words, words, words.', maxsplit=1))
```
gives `TypeError: <native JS>() got multiple values for keyword argument 'maxsplit' on line 25 in main.py`

albertjan: I'd try making them `Sk.builtin.func` see if that works. If it does I think it'd be a good idea to remove it all together. But thats for another PR.

eah13: Doing so seems to make ~200 test files that utilize Exceptions fail immediately with `EXCEPTION: TypeError`. Wat.

rixner: This was the original way of allowing keyword arguments.  Look at ```native.js```.  It was done differently later, probably because people were unaware that it was there.  So instead of using/improving it, other strategies were employed.  Remember the ```re``` module was written a long time ago and hasn't needed to be touched...

bnmnetp:
@skulpt/committers Remember to use the Review/approve button when giving your thumbs up to a PR.  -- Thanks

src/lib/re.js
meredydd: Specifically, the way to do it now is `co_varnames`.

rixner: ```Sk.nativejs.func``` sets ```co_varnames``` properly.

albertjan: Oh right I remember this now! We had a whole discussion about the regexes back then. My memory failing me.

meredydd: Oh, wow! I can't decide whether that's too grotty to live, or the best thing I've ever seen that's going to save me hours of work and should be expanded to argument type checking :D

rixner: Thanks?  I think? :smile:

albertjan:
@rixner about nativejs funcs, I think it's really fancy but it was only used for these two functions. Lets open a new ticket, to flesh out some docs and some of the missing features. And if we agree it's should be the way to do skulpt funcs from javascript, we should refactor it to be used everywhere.

eah13:
Agreed RE nativejs.  In addition, I think we'd need to update the developer docs.  AFICT, the source is the only place to learn about it.

speaking of docs, might be a good idea to update HACKING.md with examples of default parameters for module functions and methods.  The `sorted` example was a tad confusing since it didn't indicate that the attributes had to go on an underlying js function, before the `new Sk.builtin.func` call, and I think most contributions will not be made directly to builtins at this stage of the game.  Could add that to this PR or if we need to discuss builtin.func vs nativejs I can do it elsewhere.

rixner:
@albertjan, I think the project has long since gone in a different direction.  If the only calls to it get removed, the native.js file needs to be removed as well.  Otherwise, it is just dead code that will be sitting around in the distribution.

eah13:
I rebased this branch onto master to resolve the conflict and now the history looks a little messed. I'm still not sure why or when github shows rebased commits in the diff sometimes even though master already has them. Anyone who knows please educate me. But we should be good - my commits are shown here at the bottom.  

As suggested by @rixner, last commit removes native.js and its usage in `skulpt.py` and a test html file since it's now unused.

This code has been in production on trinket for a few days with no issues https://trinket.io/python/735698a2ed

bnmnetp:
I tried the following simple example program:

```
import re

x = re.compile('[a]+[b]+')
print(x.match('aaabbbbbb'))
```

When I try to run this I get
```
UNCAUGHT EXCEPTION: ExternalError: TypeError: Cannot read property 'v' of undefined on line 4
```

bnmnetp:
I tried the same example above on trinket and it runs.  But not on a clean master of skulpt.

bnmnetp:
@eah13 I've tried to trace this down, but it appears that by the time it gets to line 375 the string parameter is gone. So it calls slice and tries to remap nothing once the cannot read property v of undefined.  What happens to the string parameter along the way is unknown to me.

eah13:
Thanks for looking into it. I'll try to solve this tomorrow.

ebertmi:
any progress on this one?

albertjan:
@eah13 I'm gonna close this and open a new PR since the history on this one is completely broken 😄

PR #710, opened 2017-08-03
with open('file') ad file

albertjan:
Thanks to @meredydd we can now have this nice thing. Also this fixes a bunch of old bugs in file.

### DIFF
--- a/src/file.js
+++ b/src/file.js
@@ -24,22 +24,8 @@ Sk.builtin.file = function (name, mode, buffering) {
     } else if (this.name === "/dev/stderr") {
         this.fileno = 2;
     } else {
-        if (Sk.inBrowser) {  // todo:  Maybe provide a replaceable function for non-import files
-            this.fileno = 10;
-            elem = document.getElementById(name.v);
-            if (elem == null) {
-                throw new Sk.builtin.IOError("[Errno 2] No such file or directory: '" + name.v + "'");
-            } else {
-                if (elem.nodeName.toLowerCase() == "textarea") {
-                    this.data$ = elem.value;
-                } else {
-                    this.data$ = elem.textContent;
-                }
-            }
-        } else {
-            this.fileno = 11;
-            this.data$ = Sk.read(name.v);
-        }
+        this.fileno = 11;
+        this.data$ = Sk.read(name.v);
 
         this.lineList = this.data$.split("\n");
         this.lineList = this.lineList.slice(0, -1);
@@ -115,19 +101,24 @@ Sk.builtin.file.prototype["isatty"] = new Sk.builtin.func(function isatty(self) 
 
 Sk.builtin.file.prototype["read"] = new Sk.builtin.func(function read(self, size) {
     var ret;
-    var len;
+    var len = self.data$.length;
+    var l_size;
     if (self.closed) {
         throw new Sk.builtin.ValueError("I/O operation on closed file");
     }
-    len = self.data$.length;
+
     if (size === undefined) {
-        size = len;
+        l_size = len;
+    } else {
+        l_size = Sk.ffi.remapToJs(size);
     }
-    ret = new Sk.builtin.str(self.data$.substr(self.pos$, size));
+
+    ret = new Sk.builtin.str(self.data$.substr(self.pos$, l_size));
     self.pos$ += size;
     if (self.pos$ >= len) {
         self.pos$ = len;
     }
+
     return ret;
 });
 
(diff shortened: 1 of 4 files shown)


src/file.js
```diff
@@ -179,29 +179,39 @@ Sk.builtin.file.prototype["readlines"] = new Sk.builtin.func(function (self, siz
...
-        self.pos$ = offset;
-    } else {
-        self.pos$ = self.data$ + offset;
+    if (whence == 0) {
+        self.pos$ = l_offset;
+    } else if (whence == 1) {
+        self.pos$ = self.data$.length + l_offset;
+    } else if (whence == 2) {
+        self.pos$ = self.data$.length + l_offset;
     }
```
meredydd: Q: Should we clamp `self.pos$` to 0...length-of-file here?

eah13:
Yeah I think this may need a browser to test. 

Here it is in production:
https://trinket.io/python/0a646726c1

bnmnetp:
Oh Sweet!!  Thanks @albertjan

src/file.js
```diff
@@ -24,22 +24,8 @@ Sk.builtin.file = function (name, mode, buffering) {
     } else if (this.name === "/dev/stderr") {
         this.fileno = 2;
     } else {
-        if (Sk.inBrowser) {  // todo:  Maybe provide a replaceable function for non-import files
```
bnmnetp: Maybe I'm missing some other piece, but I just built this into my skulpt for Runestone and I get the following error trying to run @eah13 's example code.

```
with open("foo.txt", "w") as f:
  print("Writing to foo.txt...\n")
  f.write("With open(...) now works on trinket!\n")
  f.write("yay for idiomatic Python!")
  
with open("foo.txt") as f:
  text = f.read()
  print("Contents of foo.txt:")
  print(text)
```
I added an element with the id="foo.txt"
The error I get is:  

```
ExternalError: File not found: 'foo.txt' on line 1
```

Even worse is that it appears to have broken all file processing that used to work.  For example:  http://interactivepython.org/runestone/static/thinkcspy/Files/Iteratingoverlinesinafile.html  works with a skulpt pre-this-PR but that same page fails with `ExternalError: File not found: 'qbdata.txt' on line 1` using this build.

src/file.js
albertjan: hehe nope: ```
Python 2.7.10 (default, Feb  6 2017, 23:53:20) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
>>> f = open("README.md")
>>> f.seek(100, 2)
>>> f.tell
<built-in method tell of file object at 0x10083e5d0>
>>> f.tell()
6499
>>> import os
>>> os.path.getsize("README.md")
6399
>>> ```

src/file.js
albertjan: Oh yeah, I may have been a bit rigorous there, I figured that was an implementation detail that didn't really belong in `file` and should be implemented in a `Sk.read` implementation. I'm glad to put it back. But I'd rather then specify a default `Sk.read` implementation rather then adding these things to `file`.

albertjan: Let me know what you want to do with this @bnmnetp.

albertjan:
@meredydd I just forgot to add the tests 😄

src/file.js
bnmnetp: OK, Too much multi-tasking for me!

I think I understand now.   Essentially I need to take that code and put it in my own function and configure it as Sk.read  which in my current environment Sk.read would thrown an unimplemented error.

What about the write function?

albertjan: I'm just of to bed let me check tomorrow.

albertjan: Skulpt currently doesn't support `write` for anything other then `stdout` If runestone does you've probably added something to your branch to do that.

eah13: Ah, I wish.  @brianpmarks writes all our good code.  :)

Things like file handling get pretty idiosyncratic, so I doubt our code is useful to anyone else. Either way, here are some snippets for how we do file writing:

https://gist.github.com/eah13/a6a805afbfca15031893409d936b5df1

Reading is similar.  You'll see a lot in there that's specific how we do file tabs but I think Runestone could totally do something like this for your ActiveCode widgets.

bnmnetp: Now I feel like I've really missed something.  Because I thought Sk.read had to be configured to handle imports -- by checking `Sk.builtinFiles`   Did I miss some configuration change to have imports use something else?

bnmnetp: I **can** have my custom read function do both, but it seems risky to combine to different "virtual file systems" into one read.  Which should take precedence?

albertjan: Well in the real world you can't have two files with the same path, neither should you really with skulpt. And since there's no write function, you can't actually create a clash between files as a user. 

It might even prove interesting to be able it `import yadayada` where that is a bit of python on the site, in a textarea.

bnmnetp: OK, agreed.

src/file.js
```diff
@@ -179,29 +179,39 @@ Sk.builtin.file.prototype["readlines"] = new Sk.builtin.func(function (self, siz
...
-Sk.builtin.file.prototype["write"] = new Sk.builtin.func(function (self, str) {
-    if (self.fileno === 1) {
-        Sk.output(Sk.ffi.remapToJs(str));
+Sk.builtin.file.prototype["write"] = new Sk.builtin.func(function write(self, str) {
+    var mode = Sk.ffi.remapToJs(self.mode);
+    if (mode === "w" || mode === "wb" || mode === "a" || mode === "ab") {
+        if (self.fileno === 1) {
+            Sk.output(Sk.ffi.remapToJs(str));
+        } else {
+            goog.asserts.fail();
```
bnmnetp: Rather than fail,  would this be a a good place to check to see if the system has configured `Sk.writeFile` and if so call that?  Of course if Sk.writeFile is not configured then failing is the thing to do.

albertjan: I agree but maybe for a different PR I can't completely grok the implementation details of that one quickly.

bnmnetp: I wasn't asking you to implement a default version of Sk.writeFile only asking to check for its existence and if so call it passing a javascript string.  -- Not a big deal.  I could do it on my branch and PR it also.

albertjan: Oh yeah sure 😄 I can add that

bnmnetp: OK, sorry two parameters!   the trinket implementation accepts the file object and a string.  so internally it would be called `Sk.writeFile(this, str)`

bnmnetp:
I'm not going to hold this one up over my request for sk.writeFile.  That can easily be done in its own PR without any impact on the intent of this PR.   Unless @albertjan wants me to wait I'm going to merge this one later today.

PR #722, opened 2017-09-29
Support relative imports

meredydd:
This PR supports relative imports - if you are in the `foo` module, and do `import bar.baz`, we look for `foo.bar.baz` first. This involved updating a bunch of the boring plumbing code, including:

* Properly identifying packages vs modules (by putting a `__path__` property on the former, and putting the `__package__` property on each module)
* Making `Sk.importModuleInternal_` load modules from the package `__path__` when specified. While I was reworking it, I made it use `Sk.misceval.chain()` and `Sk.misceval.tryCatch()` instead of hand-rolling everything. (Imports were one of the first things to be suspension-ized, and the tools were primitive back then! It's much nicer now.)
* Changing module initialisation. It is now vital that the module's global scope (`$d`) is linked up *before* the module body runs, so I now actually create the scope object in `import.js` and pass it into the compiled JS.

and finally, the actual change that I was going for:

* Making `Sk.__import__` try loading packages relatively first

This PR entirely replaces @albertjan's "currentDir" mechanism from Jan 2016, which appeared to be a hack to emulate relative imports. (It didn't work if you had two modules of the same name, because it didn't cache the modules with their correct absolute names in `sys.modules`, which cased aliasing between local and global modules with the same name.)

This PR does not support the Python 3 syntax for relative imports, but implementing it would now be fairly straightforward if anyone's interested.


albertjan:
We need to do something about those jscs and jshint things. @j-f1 I thought we fixed this 😄

albertjan:
Awesome work again @meredydd, I'm going to set aside some time to look at this properly. I like how you always seem to run into a hack I did some time ago to make it do something like what it's supposed to do. The `currentDir` thing did solve my problem at that point but I was aware at that point that there is much more to `import` than that.

src/import.js
```diff
@@ -458,62 +436,74 @@ Sk.importModuleInternal_ = function (name, dumpJS, modname, suppliedPyBody, canS
...
-        finalcode += "\n" + co.funcname + "(" + namestr + ");";
+        finalcode += "\n" + co.funcname + ";";
 
-        modlocs = goog.global["eval"](finalcode);
+        modscope = goog.global["eval"](finalcode);
 
-        return (function finishLoading(modlocs) {
+        module["$d"] = {
+            "__name__": new Sk.builtin.str(modname),
+            "__doc__": Sk.builtin.none.none$,
```
ebertmi: What happens when `mod.__doc__` has been defined in the module? I remember, that the parser currently just skips the comments, but they might be present in js modules.

albertjan:
I don't think we need py3 tests, since there is no specific py3 functionality.

src/import.js
meredydd: We haven't run the module body yet at this point! This code is providing default values that the module body can overwrite if it needs to. (Previously, we filled these in after eval-ing the code, in case the module didn't fill it in. But now we have an import system that takes account of `__package__` and `__path__`, we need those variables to be set before the module body runs so that `import` statements within the module are resolved correctly.

src/import.js
```diff
@@ -111,72 +111,44 @@ Sk.loadExternalLibrary = function (name) {
...
-                        return tryNextPath();
-                    }
-                };
-                return ns;
-            } else {
-                return {filename: fns[j], code: s};
+    var tryPathAndBreakOnSuccess = function(filename, packagePath) {
+        return Sk.misceval.chain(
+            Sk.misceval.tryCatch(function() {
+                return Sk.read(filename);
```
albertjan: Does this mean, Sk.read can now return a suspension?

src/import.js
```diff
@@ -339,88 +310,95 @@ Sk.importModuleInternal_ = function (name, dumpJS, modname, suppliedPyBody, canS
...
+                result = Sk.onBeforeImport(name);
+                if (result === false) {
+                    throw new Sk.builtin.ImportError("Importing " + name + " is not allowed");
+                } else if (typeof result === "string") {
+                    throw new Sk.builtin.ImportError(result);
+                }
+            }
+
+            // check first for an externally loaded library
+            external = Sk.loadExternalLibrary(name);
```
albertjan: should this not be handled by `Sk.read`? Not in the scope of this PR just wondering if you agree.

meredydd: Yes, I would like that. We should have a chat about external code, because one of these days I'm implementing AOT compilation for Anvil...

src/import.js
meredydd: Yes. It always could, since about 2014 (it's why the first suspension-ised version of `import` was so tangled). Only now the code is nice enough that you can notice :D

acbart:
This is beautiful - plugged it in to my Skulpt instance and it fixes an error we have with a complicated package. I hope it gets merged in soon!

albertjan:
this has been in production at trinket for a while too. @bnmnetp what do you feel about this? I think it's an important step forward.

bnmnetp:
Yes

And would love your thoughts on #798

albertjan:
@bnmnetp shall we merge this one?

PR #724, opened 2017-10-05
remove calls to jscs and jshint in travis.yml

albertjan:
At some point in the past `jscs` and `jshint` weren't called with the correct args anymore, and because we don't globally install jscs and jshint anymore, the script broke. This PR addresses these things and fixes the collected lint over time.

### DIFF
--- a/skulpt.py
+++ b/skulpt.py
@@ -239,17 +239,12 @@ def test(debug_mode=False, p3=False):
             jshintcmd = "jshint " + folders
             jscscmd = "jscs " + folders + " --reporter=inline"
 
-        print jshintcmd
-        print jscscmd
         ret2 = os.system(jshintcmd)
         print "Running JSCS"
         ret3 = os.system(jscscmd)
-        print ret3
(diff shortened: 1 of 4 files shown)


skulpt.py
```diff
@@ -1312,13 +1316,13 @@ def main():
         f.write(getInternalCodeAsJson() + ";")
 
     if cmd == "test":
-        test()
+        exit(bool(test()))
```
meredydd: Do these get the success/failure thing the right way round? I thought `exit()` took an integer, and 0 was success and 1 was failure...

albertjan: `test()` returns non zero for something bad

`bool(0) => False => 0` which is OK
`bool(>0) => True => 1` which is not OK

meredydd: Approved, then!

albertjan: The problem I was having is that jscs returns `512` on error and `exit(512)` results in a `0` return code :trollface:

meredydd: [Head->Desk]

PR #725, opened 2017-10-09
Processing ectomy

albertjan:
removes old processing and adds global javascript properties 😁


bnmnetp:
Is there a replacement processing forthcoming or are your just trying to slim down the code base?

albertjan:
Drat! this was supposed to stay my secret project for a while 😄. To be continued! I promise!

bnmnetp:
So you are saying this is Deferred ? 😄

dsblank:
Look forward to seeing the secret project! Processing with Python is the cat's meow!

PR #726, opened 2017-10-10
Support suspending in property getters and setters

meredydd:
`@property` is great, but it currently breaks if the getter or setter function suspends. This PR fixes that:

* It compiles python builtins with `canSuspend` set to true, so that the compiled code from `property.py` supports suspensions.
* It fixes some `canSuspend` plumbing to make suspending descriptor setters work
* (This branch is on top of #708, which already makes suspending descriptor getters work)


albertjan:
@skulpt/committers Lets get these merged! 😄

ebertmi:
I just reviewed those changes - get this merged in combination with the #708. I am mainly interested in the property decorator improvements :)

Issue #666, opened 2017-02-03
Switch to webpack


j-f1:
[webpack](https://webpack.js.org) is a new module builder for JavaScript. 

[Comparison with other module builders](https://webpack.js.org/guides/why-webpack/#comparison)
If we decide to switch, here is a checklist:

- [ ] add webpack to the list of dependencies
- [ ] replace all of the `goog.*` functions with native JS
- [ ] replace `Sk.foo.bar.baz` with ES6 imports (webpack will transpile these automatically)
- [ ] make a backwards-compatibility module that creates an `Sk` object with all of the appropriate attributes

<sup>I’m happy to do this work myself, I just don’t want to waste effort creating something nobody here wants.</sup>

waywaaard:
First of all I like the idea. But there are several problems that come to my mind:

- The skulpt code is pretty tight coupled with google closure, e.g. see `/gen/*` or make a search for all `goog.*` occurences
- Our inheritance code relies currently on the google closure behavior, though that should be not a problem to switch out
- Our checking also relies on that
- Currently, we have order problems, so sometimes we rely on the existance of functions at later points
- And I am not sure if we introduce a performance decrease with webpack. Closure compiler is pretty good, and the comparison table lacks google closure

I guess for a project of this size, typescript would be more suited. Yet, a proper review and testing process (which is sometimes quite difficult in skulpt [internals]) catches most of those.

j-f1:
* We can switch to Node/npm for most of the `goog.*` occurrences, and `src/str.js` could use its own regexes.
  <details><summary>Details</summary><ul>
  <li>`goog.isArrayLike` → `isArrayLike` [from Lodash](https://lodash.com/docs#isArrayLike) [npm](https://www.npmjs.com/package/lodash.isarraylike)
  <li>`goog.global` → `global` (set to `window` by webpack automatically)
  <li>`goog.string.*` → regexes (they’re only used in `src/str.js`)
  <li>`goog.string.trimLeft` → [`trim-left`](https://www.npmjs.com/package/trim-left)
  </ul></details>
* Inheritance could be refactored with ES6 classes.
* All assert logic can be replaced with the Node standard library
  <details><summary>Details</summary><ul>
  <li> add `import assert from 'assert'` to the top of all files that use `assert`
  <li>`goog.asserts.*` → `assert.*`
  <li>`goog.asserts.assert(...)` → `assert(...)`
[...]

PR #733, opened 2017-11-04
creates a suspension from a promise

albertjan:
I finally broke and wrote this. This turns a promise into a suspension.

@meredydd I fully expect you to say I'm doing it all wrong .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ sudo: false
 python:
 - '2.6'
 install:
-- nvm install node
-- nvm use node
+- nvm install 'lts/*'
+- nvm use 'lts/*'
 - npm install npm@latest -g
 - npm install
 - pip install "GitPython>=0.2.0-beta1"
--- a/src/lib/time.js
+++ b/src/lib/time.js
@@ -59,12 +59,12 @@ var $builtinmodule = function (name) {
     mod.sleep = new Sk.builtin.func(function(delay) {
         Sk.builtin.pyCheckArgs("sleep", arguments, 1, 1);
         Sk.builtin.pyCheckType("delay", "float", Sk.builtin.checkNumber(delay));
-        var susp = new Sk.misceval.Suspension();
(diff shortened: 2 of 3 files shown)


albertjan:
Well no I haven’t I was interrupted by dinner :P. What do you mean with renamed the variable?

meredydd:
We call it `newSuspension`. You call it `susp` (and `suspension`, apparently ;) ). Other than that, that's the same code we use :)

albertjan:
There it builds and I've even scrubbed the history of my faux pas.

meredydd:
Does Closure understand the `let` statement and compile it to vanilla IE-compatible Javascript?

(I agree that it would be good to be able to use modern features like `let` in Skulpt, but to do that we'll need tooling that allows us to support those older browsers...)

albertjan:
The version of closure or v8 we use don't support `let` it doesn't compile. And yes, I love all the new javascript niceties, and I'd love to use them in skulpt too. This was copied from a side project, which uses babel and rollup to do the compiling.

meredydd:
Minor change, merging.

bnmnetp:
How closely related to #666 is this issue?  Is there a version of closure that is compatible with let and other es6 goodness??

albertjan:
I don't know @bnmnetp but it might be worth looking into 😄

bnmnetp:
The closure compiler wiki says:

ECMAScript 6 is now officially supported as an input language for the Closure Compiler. You cannot use it as an output language yet; for now you need to transpile down to ES5 or ES3. If you find that the Closure Compiler does not support your favorite ES6 feature, you can file an issue, but please be aware that supporting the entire ES6 specification is a non-goal for the Closure Compiler.

They have a nice list of features that are supported on the input:  https://github.com/google/closure-compiler/wiki/ECMAScript6

albertjan:
OHhhh this is really nice!

meredydd:
YES. Propose we start doing ES6 right away.

PR #735, opened 2017-11-17
Prevent time.js exploding if no timezone available

meredydd:
It now defaults to UTC if there's no timezone in the browser's default string representation, which will be an unpleasant surprise but beats an ExternalError and is probably a sensible default.


albertjan:
@skulpt/committers trivial change merging tomorrow if no objections

ebertmi:
do it :dancer:

PR #736, opened 2017-11-30
Function descriptor

albertjan:
This makes it possible to do:

```
>>> def test(self):
...      print self

>>> test.__get__(5)()
5
```

or (what I wanted to do):

```
class ClassOrInstanceMethod(object):
    def __init__(self, wrapped):
        self.wrapped = wrapped
    def __get__(self, instance, owner):
        if instance is None:
            instance = owner
        return self.wrapped.__get__(instance, owner)

class demo(object):
    @ClassOrInstanceMethod
    def foo(self):
        # self will be the class if this is called on the class
        print(self)

demo.foo()
demo().foo()
```

I commented out `CodeType = type(_f.func_code)` `func_code` has never had a proper type. So this never made any sense what so ever. 

@meredydd I hope I haven't disturbed any of the attribute changes you did.

It also clears up so annoyances I've had with the browser run template for a file.


src/object.js
```diff
@@ -334,6 +334,14 @@ Sk.builtin.object.prototype["__ge__"] = function (self, other) {
     return self.ob$ge(other);
 };
 
+Sk.builtin.object.prototype.__get__ = function __get__(self, instance, owner) {
+    Sk.builtin.pyCheckArgs("__get__", arguments, 1, 2, false, true);
+
+    var l_owner = !owner ? instance.ob$type : owner;
+
+    return self.tp$descr_get(instance, l_owner);
```
meredydd: In CPython, `object` does not automatically have a `__get__` method.

I appreciate we'd prefer not to rewrite a generic definition of `__get__` for every type that implements `tp$descr_get()`, but adding a method to `object` that shouldn't be there isn't the answer.

Perhaps we should have a function that uses the `dunderToSkulpt` mapping (from `type.js`) in reverse - ie, given a native implementation of a slot (like `tp$descr_get`), it generates a Python-callable version of that function (in this case `__get__`)? Something like `makeDunderFunction(Sk.builtin.func.prototype, "__get__")`?

support/run_template.html
```diff
@@ -14,7 +14,8 @@ <h1>In Browser Testing Page</h1>
 <script type="text/javascript">
 function outf(text) {
    var mypre = document.getElementById(Sk.pre);
-   mypre.innerHTML = mypre.innerHTML + text;
```
meredydd: I'm guessing this HTML-munging was meant to be in a separate PR :)

src/object.js
albertjan: Ohh right yeah of course there isn't, I'll have to rethink this 😁.

support/run_template.html
albertjan: Yeah I guess.

albertjan:
I need to fiddle with it a bit

albertjan:
🎉 Some unit tests to be added.

albertjan:
@meredydd @bnmnetp this is done 😄 I think.

bnmnetp:
Of course now that I've accepted this I found this:

```

**Comment by albertjan** [comment]
🤦‍♂️

**Comment by albertjan** [comment]
I've got the downstream merge from trinket almost done. Apparently that contains a fix for this 😄 See below:

```JavaScript debugger enabled
Python 2.6(ish) (skulpt, Thu Mar 14 2019 16:11:12 GMT+0000 (GMT))
[v8: 3.25.30] on a system
>>> def foo():
...     print('hello')
...     print(foo.bar)
... 
>>> foo.bar = 10
10
>>> foo()
hello
10
```

bnmnetp:
Should I keep this merge or revert it?  I'm happy to do whatever will best facilitate the merge back from trinket!

albertjan:
Please leave it :) if we merge #827 it should be fixed

PR #739, opened 2017-12-15
Re.compile mapped to existing re functions

### LINKED ISSUE #705: spoofing re.compile

albertjan:
This includes:

re.compile
RegexObject and the methods I could reuse code for: .findall(), .search(), .split(), and .match().
re.purge() got a stub so we can run code that uses it but we don't cache so no need to purge
A bunch of CPython tests. These led me to do a few random things like __str__ and __repr__
As I culled stuff from the CPython file I left failing tests behind a if include_failing statement if they looked like things we'd get to the next time someone looks at re. Seemed like a good thing to do but I can scrap this if the consensus is that it's messy.

Made some style changes to the file here and there, trying to follow our guidelines. Hoisted vars, double quotes, semicolon hygiene, etc. LMK if anything seems amiss. I'm not a Javascript native.

I left plenty of stuff out of scope in the interests of moving quickly. For instance, you'll see todos for pos or endpos arguments. I also couldn't figure out how to get the .flags and .pattern attributes set up when the RegexObject gets instantiated. I've never had to do either of these in Skulpt before so any pointers or a link to an example in the codebase would be welcome.

Someone, someday will implement re.sub. I'm not that one and/or this is not the day.

Fixes #705 

Re does: #709


albertjan:
This is all @eah13's work I just reapplied the changes and fix one bug @bnmnetp found.

bnmnetp:
That works for me now too!

Thanks!

src/lib/re.js
```diff
@@ -220,89 +218,227 @@ var $builtinmodule = function (name) {
...
+
+        $loc.__repr__ = _repr;
+
+        // Given a string, start, and end position, return sliced string
+        _slice = function(string, pos, endpos) {
+            // Per docs, ^ should match index after newlines.
+            // this matches the first
+            var str = Sk.ffi.remapToJs(string);
+            var start = pos == undefined ? 0 : Sk.ffi.remapToJs(pos);
+            var end = endpos == undefined ? str.length : Sk.ffi.remapToJs(endpos);
```
albertjan: @eah13 these two `undefined` checks were added ti fix the bug @bnmnetp found

eah13: Ah yes.  Thanks!

PR #743, opened 2018-01-02
fix the skulpt.py bug

acekingke:
Signed-off-by: Ke Yuchang <keyuchang@kaka888.com>

### DIFF
--- a/skulpt.py
+++ b/skulpt.py
@@ -175,7 +175,7 @@ def is64bit():
     return sys.maxsize > 2**32
 
 if sys.platform == "win32":
-    winbase = ".\\support\\d8\\x32"
+    winbase = ".\\support\\d8\\x32_"
     if not os.path.exists(winbase):
         winbase = ".\\support\\d8"
     os.environ["D8_PATH"] = winbase
--- a/skulpt.py
+++ b/skulpt.py
@@ -175,9 +175,7 @@ def is64bit():
     return sys.maxsize > 2**32
 
 if sys.platform == "win32":
-    winbase = ".\\support\\d8\\x32_"
-    if not os.path.exists(winbase):
-        winbase = ".\\support\\d8"
+    winbase = ".\\support\\d8"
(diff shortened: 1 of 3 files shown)


skulpt.py
```diff
@@ -175,7 +175,7 @@ def is64bit():
     return sys.maxsize > 2**32
 
 if sys.platform == "win32":
-    winbase = ".\\support\\d8\\x32"
+    winbase = ".\\support\\d8\\x32_"
```
albertjan: I see why you did this but, I think you should change this to `".\\support\\d8.exe"` and remove line 179 and 180. I don't see the need of this 32 bit check.

src/pgen/parser/tokenize.py
```diff
@@ -31,13 +31,13 @@
...
 from codecs import BOM_UTF8, lookup
-from lib2to3.pgen2.token import *
+#from lib2to3.pgen2.token import *
 
 import token
 __all__ = [x for x in dir(token) if x[0] != '_'] + ["tokenize",
            "generate_tokens", "untokenize"]
 del token
-
+from token  import *
```
albertjan: Remove the extra space though

albertjan:
Hi @acekingke can you revert the change in the package.json, and just call `.\skulpt.py` on windows  directly, we need to figure out a better way to do this on both windows and mac. 

@j-f1 do you know how we could solve this in a cross platform way? Maybe install skulpt in `./node_modules/.bin` (but I don't know how)

j-f1:
You can do `python skulpt.py`, which will look for `skulpt.py` in the current directory.

albertjan:
@j-f1 yeah 😄 ... but then we have the whole python version thing that screws us up.

bnmnetp:
@albertjan Is this still relevant or should we close it?

bnmnetp:
Thanks, this is superseded by the ongoing work with node.

PR #748, opened 2018-01-11
Processing  - stroke alpha added

pinkfoot:
Stroke alpha added as per fill alpha


dsblank:
Looks good to me. Thank you for reporting, and for fixing!

albertjan:
@skulpt/committers merge at will 😄

PR #750, opened 2018-01-28
Fix 'import *'

meredydd:
`from module import *` should only import public names (ones that don't start with `_`).

I'd classify this as a small change that can be approved by one reviewer.


albertjan:
Does the `__name__` erasing thing not happen anymore?

meredydd:
No, because `__name__` starts with an underscore, so it won't be imported. (The bug that caused me to find this issue was that I was getting weird import path behaviour because `import *` was picking up the `__path__` from another package, causing subsequent relative imports to look in the wrong place.)

albertjan:
Merge away, but we should mention this'll need to be revisted at some point to take `__all__` into account, right?

Issue #751, opened 2018-02-02
Java is required to build Skupt


tahteche:
Java is required to build Skulpt but that is not indicated in the "Building Skulpt" section of README.md

When building Skulpt on Ubuntu 16.04 I got an error that Java was not found. Had to install it before the build succeeded.

PR #753, opened 2018-02-02
Doc: Adds Java to the list of requirements in "Building Skulpt" section.

tahteche:
Fixes: #751

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ benefit.
 Building Skulpt is straightforward:
 
 1. Clone the repository from GitHub, ideally using your own fork if you're planning on making any contributions
-2. Install node.js and Python 2 (required to run the build scripts)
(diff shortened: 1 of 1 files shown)


rixner:
This is an obvious fix.  Thanks!

j-f1:
Have you considered using [`closuer-compiler-js`](https://github.com/google/closure-compiler-js)? You can `npm install --save-dev` it, and scripts run via `npm run` will have access to it as the `google-closure-compiler-js` command. Java is not required.

rixner:
@j-f1  I think that is part of a larger discussion about modifying our overall build process...  I don't know if there is critical mass to have that discussion and actually follow through with overhauling things?

In my opinion, this pull request should go through regardless, as it improves the accuracy of the description of the current process.

PR #759, opened 2018-02-08
spruce up jseval

albertjan:
@prescod have a play with this see if this works for you.

`jseval("5")` will return `5`

and

`jseval("window.alert('test')")` also still works and returns `None`

and 

`jseval("throw new Error('testing!');")` also still "works"

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -862,7 +862,16 @@ Sk.builtin.raw_input = function (prompt) {
 Sk.builtin.input = Sk.builtin.raw_input;
 
 Sk.builtin.jseval = function jseval (evalcode) {
-    goog.global["eval"](Sk.ffi.remapToJs(evalcode));
+    var result = goog.global["eval"](Sk.ffi.remapToJs(evalcode));
+    try {
+        return Sk.ffi.remapToPy(result);
+    } catch (err) {
+        if (err.constructor === goog.asserts.AssertionError) {
+            return Sk.builtin.none.none$;
+        }
+
+        throw err;
+    }
 };
 
 Sk.builtin.jsmillis = function jsmillis () {


src/builtin.js
```diff
@@ -862,7 +862,16 @@ Sk.builtin.raw_input = function (prompt) {
 Sk.builtin.input = Sk.builtin.raw_input;
 
 Sk.builtin.jseval = function jseval (evalcode) {
-    goog.global["eval"](Sk.ffi.remapToJs(evalcode));
+    var result = goog.global["eval"](Sk.ffi.remapToJs(evalcode));
+    try {
+        return Sk.ffi.remapToPy(result);
+    } catch (err) {
+        if (err.constructor === goog.asserts.AssertionError) {
```
j-f1: Ho come this is special-cased?

albertjan:
If you return something that Sk.ffi.remapToPy can’t remap it throws an AssertionError.

prescod:
It seems to work for me, but I guess my deeper question is why are we adding things to the global namespace at all? What's wrong with "from skulpt import jseval" or something like that?

Or perhaps at least call it __jseval__ or something?

Cramming stuff into the main namespace seems a bit unPythonic to me, TBH.

prescod:
Maybe it's a separate issue that I'm complaining about now, though. JSEval definitely works better than it did. Thanks.

albertjan:
@prescod why it is where it is predates my involvement, but changing it now constitutes a breaking change and it doesn't bother me much it's there 😄

albertjan:
@skulpt/committers do we like this? Or are we going to do a separate module so we can `import js`, although I don't know when we'll get to that.

meredydd:
I don't like it in the global namespace either, but given that it is there, making it less likely to explode is an easy accept. Approved.

Issue #762, opened 2018-02-12
Request to add codetheblocks.com to list of "Skulpt in the Wild" projects


chhabrakadabra:
This request is based on your comment on [your homepage](skulpt.org):

> If you have a project that uses skulpt and would like me to link to it here, let me know on our github page.

I'm hoping you wouldn't mind adding my project [codetheblocks.com](https://codetheblocks.com) to this list.

Hope you like it :)

albertjan:
Looks cool, open a PR and I'll merge it the site is in the `doc/` directory

PR #763, opened 2018-02-13
Added "Code the blocks" to the list of projects using Skulpt.

chhabrakadabra:
Fixes: #762 

And thank you for agreeing to add this in 😄


chhabrakadabra:
@albertjan would you be so kind as to merge this in 😄

albertjan:
it'll take a while before this ends up on the site though 😄

chhabrakadabra:
Sounds good. Thank you.

PR #768, opened 2018-02-22
don't copy the array when we make an iterator

### LINKED ISSUE #749: Regression: variables in "for" loops are handled incorrectly when modified during the loop
In the latest version of Skulpt ([a472e2e](https://github.com/skulpt/skulpt/tree/a472e2e170302c123db35e366cbd917b25a55998)), the following code results in the value of `counter` being set to `4`, whereas it should only end up being `2` because the iterated-over list is being modified during the iteration.

```
L = ['a', 'b', 'c', 'd']
counter = 0
for c in L:
    L.pop()
    counter += 1
print counter
```

Note that this works correctly on www.skulpt.org, which runs an earlier version of Skulpt (not sure which one), so I think it's a regression in the library. I did a little digging and it looks like it was working correctly in v0.10.0 (b431c9c) but is broken in c140150, so it's probably due to some commit between those two boundaries.

albertjan:
fixes #749


meredydd:
If I'm reading the linked issue correctly, does this mean we're switching from Python 2 to Python 2 behaviour here?

Also, just to check - when we modify the list underneath its iterator (eg adding elements, deleting the element the iterator points to), does it error out nicely or throw something gnarly in JS?

albertjan:
Wha? 

```➜  ~ python3
Python 3.4.3 (default, Aug  1 2017, 12:26:02) 
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.38)] on darwin
>>> L = [1,2,3]
>>> for _ in L:
...      L.pop()
...      print("t")
... 
3
t
2
t
>>> 
➜  ~ python
Python 2.7.13 (default, Aug  3 2017, 20:23:46) 
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.38)] on darwin
>>> L = [1,2,3]
>>> for _ in L:
...     L.pop()
...     print "t"
... 
3
t
2
t
>>> 
```

albertjan:
I don't see how we could f-up the underlying list in such a way that we would get a runtime error, unless you do it form javascript. I might just be being dense, and I'll have a look at this tomorrow morning to see if I can see any ovbs. mistakes I made.

albertjan:
Well this happened: 😄 more to come

```
➜  skulpt-fix git:(fix/list-iterator-modification-bug) ✗ ./m repl
JavaScript debugger enabled
Python 2.6(ish) (skulpt, Thu Feb 22 2018 18:05:40 GMT+0000 (GMT))
[v8: 3.25.30] on a system
>>> L = [1,2,3]
>>> 
>>> for _ in L:
...     L.extend(['a'])
... 
>>> L
[1, 2, 3, 'a', 'a', 'a']
>>>

➜  skulpt-fix git:(fix/list-iterator-modification-bug) ✗ python3 
Python 3.4.3 (default, Aug  1 2017, 12:26:02) 
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.38)] on darwin
>>> L = [1,2,3]
>>> for _ in L:
...     L.extend([4])
... 
... hangs ...
```

albertjan:
Well this happened: 😄 more to come

```
➜  skulpt-fix git:(fix/list-iterator-modification-bug) ✗ ./m repl
JavaScript debugger enabled
Python 2.6(ish) (skulpt, Thu Feb 22 2018 18:05:40 GMT+0000 (GMT))
[v8: 3.25.30] on a system
>>> L = [1,2,3]
>>> 
>>> for _ in L:
...     L.extend(['a'])
... 
>>> L
[1, 2, 3, 'a', 'a', 'a']
>>>

➜  skulpt-fix git:(fix/list-iterator-modification-bug) ✗ python3 
Python 3.4.3 (default, Aug  1 2017, 12:26:02) 
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.38)] on darwin
>>> L = [1,2,3]
>>> for _ in L:
...     L.extend([4])
... 
... hangs ...
```

prescod:
Here are three different test cases.

Python 2 likes all three.

Python 3 likes 2 of 3.

Your PR likes 2 of 3. But not the same 2 of 3 that Python 3 likes.

My patch (in the other issue) matches Python 3 I think.

<pre>
def test_sinkstate_list():
        # This used to fail
        a = range(5)
        b = iter(a)
        # print "list of b", list(b)
        # print next(b)

        assert list(b)==range(5)
        a.extend(range(5, 10))
        assert list(b) == []

def test_change_iterator():
        L = [1,2,3,4]
        i = iter(L)
        L.pop()
        assert len(list(i))==3

def test_lengthening():
        L = [1,2,3,4]
        i = iter(L)
        L.append("a")
        assert len(list(i)) == 5

test_sinkstate_list()
test_lengthening()
test_change_iterator()
print("PASSED")
</pre>

rixner:
At this point, I think we should use the solution suggested by @prescod in #749 and wrap it in in a ```if (Sk.__future__.python3)``` and leave the current behavior for Python 2 alone.  (@prescod if you still care all this time later and want to make such a pull request, I'd be happy to approve it.)

PR #778, opened 2018-04-30
added pep8 style names to image manipulation functions

schellenberg:
When creating a resource for teaching Python (using the Runestone framework), everything I created used pep8_style_variable_names, except for the built in image manipulation method calls (such as getRed). The only thing that has been added in this image.js file are new function names that follow pep8 styling, such as get_red. 

There may be a more elegant way to do this. I am no JS expert, so I just hacked together something that works. At first, I attempted something like:

``$loc.get_red = $loc.getRed``

but was getting errors. I'm assuming there is a way to alias the function in a reasonably elegant way (like I was trying to do above), but I'm not sure what it is...

So, instead, this file simply has a second function defined with the same code body (though with an updated ``Sk.builtin.pyCheckArgs("get_red", arguments, 1, 1);`` type call).

### DIFF
--- a/src/lib/image.js
+++ b/src/lib/image.js
@@ -109,6 +109,19 @@ $builtinmodule = function (name) {
             }
         });
 
+        // adding pep8 compliant alias to function
+        $loc.set_delay = new Sk.builtin.func(function (self, delay, interval) {
+            var i;
+            Sk.builtin.pyCheckArgs("set_delay", arguments, 2, 3);
+            self.delay = Sk.ffi.remapToJs(delay);
+            i = Sk.builtin.asnum$(interval);
+            if (!i) {
+                self.updateInterval = 1;
+            } else {
+                self.updateInterval = i;
+            }
+        });
+
 
         //get a one-dimensional array of pixel objects - Zhu
         $loc.getPixels = new Sk.builtin.func(function (self) {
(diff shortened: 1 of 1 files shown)


albertjan:
hey @schellenberg thanks for your contribution 😄 I'm surprised the `$loc.get_red = $loc.getRed` didn't work. that would be a neater way to solve this. 

@bnmnetp why did you choose to use `camelCase` here instead of `snake_case`?

bnmnetp:
@albertjan -- it was a long time ago.  In those days I was more thinking of Python as a beginning with a transition to Java or C++.  In that context camelCase made sense to me.  My thinking and the acceptance of Python have evolved and I would try much harder to follow PEP8.

schellenberg:
@bnmnetp That looks like a much cleaner method.

bnmnetp:
Great @schellenberg, if you get that cleaned up I will be happy to sign off on this one.  I appreciate the improvements.  Its a direction I'm going to be moving in for all of my writing.

schellenberg:
Ok @bnmnetp , the snake_case and camelCase functions now both alias the same private method.

bnmnetp:
@alberjan, I used let in my example, so I feel somewhat responsible.  Umm, let support has been around for many versions in Firefox, chrome and all versions of Edge, for Safari it goes all the way back to 10.0.

There is much that is good in the new ES6 and I could see us making use of it.  But we should all be on the same page.  Maybe moving to the new style modules would be a way to shake the bonds of google closure compiler...

j-f1:
#719 :)

schellenberg:
@albertjan @bnmnetp switched private methods from using ``let`` to ``var``

bnmnetp:
Thanks @schellenberg

PR #789, opened 2018-07-17
Exec for Skulpt

acbart:
So, here's some code I've written that seems to work from my limited testing. Perhaps you could comment on my approach? The core idea is to temporarily enable retainGlobals, swap out the global dictionary with the user provided one, exec the given code, swap the global dictionary back, and then copy over any updated variables into the user provided one.

* I cannot name the function `exec`, because that's a (unimplemented) statement in Skulpt's parser. So for now, I'm creating a thing called `execf`. I don't think I'll have the energy to fight the whole tokenizer and make it a function like it's supposed to be.
* I would like a better way to match up the user provided variables with the pseudo-globals dictionary that gets created before the import, but I wasn't sure if that was possible. It's strange to me that the Sk.globals dictionary seems to be a JS object instead of a Python Dictionary.
* I'm also concerned that I've failed to account for something that would allow scopes to escape.
* Finally, the code is hackish. If this is the right approach, you'll want me to clean this up and provide some docs. For now, I just want to know if this rings true for you.

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -1329,6 +1329,45 @@ Sk.builtin.delattr = function delattr (obj, attr) {
     throw new Sk.builtin.TypeError("can't set attributes of built-in/extension type '" + obj.tp$name + "'");
 };
 
+var extractDict = function(obj) {
+    ret = {};
+    for (iter = obj.tp$iter(), k = iter.tp$iternext();
+         k !== undefined;
+         k = iter.tp$iternext()) {
+        v = obj.mp$subscript(k);
+        if (v === undefined) {
+            v = null;
+        }
+        kAsJs = Sk.ffi.remapToJs(k);
+        // todo; assert that this is a reasonble lhs?
+        ret[kAsJs] = v;
+    }
+    return ret;
+}
+Sk.builtin.execf = function execf(python_code, new_globals) {
+    Sk.builtin.pyCheckArgs("execf", arguments, 1, 2);
+    var backupRG = Sk.retainGlobals;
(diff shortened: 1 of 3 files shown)


acbart:
So, I accidentally missed adding some "var"s onto the variable declarations, but a bigger issue is definitely present here. Somehow, this process messes up existing imports (possibly because of integrating #722). If I import some libraries (particularly complex ones with their own imports), exec some code, and then try to run imports again later (in a separate run session, no less), something gets clobbered and I can't import stuff anymore. So I'm missing some element here (possibly `Sk.sysmodules`?)

albertjan:
Oh man that is some side effect you have there.

I can totally see the issue you have with getting into the compiler. Although I also think it won't be so hard. What will be harder is people will rely on `locals` and `globals` to function and be correct when using `exec` and they aren't at the moment. 

I think it makes sense to wait until #722 is merged before continuing work on this. If you get this to work I'll do the compilery bits. (how hard can it be :P)

albertjan:
@acbart I think this is a great start to `exec` but there are some limitation as you explained, I don't think we can add this until we've solved those. Have you fixed these on your branch?

acbart:
I don't want to claim "fixed" too strongly, but my implementation has certainly evolved a lot more. I can try and get those changes together into a PR (which also come with a very rudimentary implementation of `compile`). I suspect others will want to poke some holes in my solution - it's worked pretty much fine for the past year, but that doesn't mean that it's to be trusted.

s-cork:
closing via #1212

PR #801, opened 2018-10-11
Ebertmi pr log2

albertjan:
Rebased version of #521


albertjan:
Well this has some unexpected failures... I'll look into them.

s-cork:
I think we should close this as the maths functions have since been added in #1004

PR #827, opened 2019-03-14
Downstream merge from trinket :tada:

albertjan:
# What is in this PR:

## Breaking changes:

- external module support is removed. this didn't belong in skulpt and can easily be moved to a `Sk.read` implementation
- `Sk.onBeforeImport` and `Sk.onAfterImport` are now suspendible if this is the case a lot of the this causes `Sk.read` to be suspended so you need to update any uses of those to handle that see `src/lib/collections.js` for example
- `Sk.config.python3` is moved to a hash containing flags for each python3 specific feature in skulpt this hash is called `__future__` there are 2 defaults (`Sk.python2` and `Sk.python3`) in `env.js` you can use them like this: `Sk.configure({ __future__: Sk.python3})`
- `co_numargs` is no longer necessary because `co_varnames` should contain all variable names of a function.

## Features and other changes (in order of files changed)

- use `setImmediate` for no-op suspensions this is faster then `setTimeout(0)`
- `skulpt.py` now takes a -d arg to disable the tests (handy if you're developing libs)
- `map` handles suspensions
- builtin's are now real python functions so they can take varargs en kwargs
- `Sk.configure()` gained `killableWhile` and `killableFor`if you don't want to lock up the browser when someone write an infinite loop
- `kwargs` `varargs` handling greatly improved
- javascript functions are automatically wrapped at runtime into skulpt functions.
- `Sk.configure()` gained `fileopen` and `filewrite` for obvious reasons
- `file` also gained the capability to read/write files in the dom set `Sk.inBrowser = true` and it will look for an element with the filename as id.
- a barebones implementation of the `array` module
- image module makes use of the `Sk.imageProxy`
- image module gained `getData`
- a `signals` module was added for handling keyboard interups. setup required we should add docs.
- turtle module gains the much awaited background image and image turtle shape, setup required we should add docs.

That's all folks! Please say thanks to @trinketapp @brianpmarks, @eah13 and @ebertmi  🙏 ❤️


bnmnetp:
Thanks @albertjan !

I've already looked at a lot of this, and I will test it on my stuff tomorrow.  I really appreciate you getting this done so quickly.

src/file.js
```diff
@@ -207,13 +229,25 @@ Sk.builtin.file.prototype["truncate"] = new Sk.builtin.func(function truncate(se
...
-            Sk.output(Sk.ffi.remapToJs(str));
+        if (Sk.filewrite) {
+            if (self.closed) {
+                throw new Sk.builtin.ValueError("I/O operation on closed file");
+            }
+
+            if (self.fileno === 1) {
+                Sk.output(Sk.ffi.remapToJs(str));
+            } else {
+                Sk.filewrite(self, str);
```
bnmnetp: Since Sk.filewrite is configurable should we not remap str to JS instead of making the configured  external function handle it??

albertjan: Yeah I guess that makes sense.

src/builtin.js
```diff
@@ -929,33 +942,41 @@ Sk.builtin.map = function map (fun, seq) {
...
             }
             retval.push(item);
-        } else {
-            if (!(item instanceof Array)) {
-                // If there was only one iterable, convert to Javascript
-                // Array for call to apply.
-                item = [item];
-            }
-            retval.push(Sk.misceval.apply(fun, undefined, undefined, undefined, item));
+            return loopDeLoop(iter);
```
meredydd: We could hit a stack overflow here, if the JS engine doesn't support tail calls

src/builtin.js
```diff
@@ -929,33 +942,41 @@ Sk.builtin.map = function map (fun, seq) {
...
     }
+
     if (!Sk.builtin.checkIterable(seq)) {
         throw new Sk.builtin.TypeError("'" + Sk.abstr.typeName(seq) + "' object is not iterable");
     }
 
-    retval = [];
+    iter = Sk.abstr.iter(seq);
+
+    return (function loopDeLoop(i) {
```
meredydd: This loop construct is subject to a stack overflow on long sequences, because it keeps calling itself. I would suggest using `Sk.misceval.iterFor()`, which abstracts out the annoying "loop over a potentially-suspending iterator without stack overflows" logic

src/builtin.js
```diff
@@ -929,33 +942,41 @@ Sk.builtin.map = function map (fun, seq) {
...
+            // If there was only one iterable, convert to Javascript
+            // Array for call to apply.
+            item = [item];
         }
-    }
 
-    return new Sk.builtin.list(retval);
+        return Sk.misceval.chain(Sk.misceval.applyOrSuspend(fun, undefined, undefined, undefined, item), function (result) {
+            retval.push(result);
+            return loopDeLoop(iter);
```
meredydd: This will call into a new stack frame for each element, which will overflow on long sequences.

`Sk.misceval.chain()` isn't like JS promises - if the first `applyOrSuspend()` does *not* suspend, it will call the function straight-out. (And it's not a tail call, so there's no way V8 can save us!) This is a substantial performance boost, which  is why `Sk.misceval.chain()` works like this, but it does mean we need annoying logic to do "loop on this Python iterator but also suspend if it suspends, then resume the loop". Thankfully, this logic is already abstracted into `iterFor()`, so we should use that instead.

src/compile.js
```diff
@@ -2011,8 +2048,8 @@ Compiler.prototype.cclass = function (s) {
     scopename = this.enterScope(s.name, s, s.lineno);
     entryBlock = this.newBlock("class entry");
 
-    this.u.prefixCode = "var " + scopename + "=(function $" + s.name.v + "$class_outer($globals,$locals,$cell){var $gbl=$globals,$loc=$locals;";
-    this.u.switchCode += "(function $" + s.name.v + "$_closure(){";
+    this.u.prefixCode = "var " + scopename + "=(function $" + s.name.v + "$class_outer($globals,$locals,$cell){var $gbl=$globals,$loc=$locals;$free=$globals;";
+    this.u.switchCode += "(function $" + s.name.v + "$_closure($cell){";
```
meredydd: <3 for this fix
(TODO I've not fully reviewed this chunk but this looks very much like the fix to a bug I've seen intermittently for a while and would like to see chased down...)

src/object.js
```diff
@@ -345,6 +351,14 @@ Sk.builtin.object.prototype["__ge__"] = function (self, other) {
     return self.ob$ge(other);
 };
 
+Sk.builtin.object.prototype.__get__ = function __get__(self, instance, owner) {
+    Sk.builtin.pyCheckArgs("__get__", arguments, 1, 2, false, true);
+
+    var l_owner = !owner ? instance.ob$type : owner;
+
+    return self.tp$descr_get(instance, l_owner);
```
meredydd: And what's the default impl of `tp$descr_get()`? (I might need to just read this more closely...)

meredydd:
I believe mine is the second +1 review, so I'm going to hit the Merge button.

src/compile.js
albertjan: I think this was to fix nested closure's I think ... It's been a while.

src/object.js
albertjan: Good question ... I don't think this should be here to be honest. Because object doesn't have __get__ in python2 right?

meredydd: Just checked...no, they don't! (Having `__get__` would turn _every_ object into a descriptor, which is probably not what you want!)

src/compile.js
meredydd: Yeah; that's the bug I spent too long banging my head against :)

PR #831, opened 2019-03-15
Move builtin.map() to use iterFor() (fixes #828)

### LINKED ISSUE #828: `map` is a stack overflow waiting to happen
As I mentioned in my review of #827, our implementation of `map()` is vulnerable to a stack overflow on long sequences, and needs to use `iterFor()`.

I'm going to make a hostage to fortune and volunteer to fix it (as the original author of `iterFor()`, which I acknowledge can be a bit intimidating).

meredydd:


albertjan:
Thanks! Looks good to me! I have to have a look at what `iterFor` exactly does. 😬

meredydd:
`iterFor` is a bit of a Swiss army knife - it's a combination `for` loop ("execute this thing for every element in this iterator, with an option to `break`) and `reduce` (the return value from each iteration is fed into the next one).

In any case, this is a minor change. Merging now.

PR #832, opened 2019-03-17
'import *' should ignore names beginning with underscores

meredydd:
Looks like we somehow regressed on the issue originally addressed by #750. `import *` should not import any name that begins with an underscore. Right now we have a special-case exception for `__name__`, but as it begins with an underscore, this is correctly covered by the more general rule.

### DIFF
--- a/src/import.js
+++ b/src/import.js
@@ -641,13 +641,11 @@ Sk.importStar = function (module, loc, global) {
     // from the global scope, globals and locals can be the same.  So the loop below
     // could accidentally overwrite __name__, erasing __main__.
     var i;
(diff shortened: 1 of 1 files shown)


meredydd:
(Note: I submit this without much enthusiasm, as we actually have to make this behaviour optional for Anvil at the moment, for backwards-compatibility reasons, But I'd rather have Skulpt be correct and carry Anvil-specific weirdness in a patch!)

meredydd:
Ack. Abandon this. I had a wrong/stale checkout. Current behaviour is correct.

PR #833, opened 2019-03-18
Support explicit relative imports

meredydd:
Clean up `__import__`, and support explicit relative imports (eg from .. import foo).


albertjan:
Yes I think this makes sense. Although `__import__` magic is dark, and we're still not not completely the same as cpython 😄. But this takes us closer so I say merge.

Issue #414, opened 2015-05-20
Accepting PRs and reviewing code


bnmnetp:
I'm moving this Here from #410 to get more people involved in the conversation.

bnmnetp:
I would really prefer to have a system where we had at least two people looking at PRs and concluding that yes, they are good enough to be accepted.  Given that we have two yes votes and no real dissenting votes then any of a few people could accept the PR. 

A lot of projects have a group of core committers.  I think there is a reasonably obvious subset of people who have contributed to skulpt over time that would form a nice group of committers.  This group has some decent variation in background and skills, and has demonstrated to me over time that they know what they are doing.  I will refrain from naming names at the moment and instead ask for volunteers to step forward.  

If other people have experience on other projects where things are run successfully with a different model I would be happy to hear about that too.

bnmnetp:
Just drawing certain people's attention to this issue:  @albertjan @eah13 @waywaaard @rixner @meredydd @bzwheeler @LeszekSwirski

albertjan:
Seems like a sensible direction. :smile: And I would like to *"volunteer"*. Just to make things clear the second person looking at the PR would then accept it if it's up to standards? 

Are we all still in agreement on the contributing [guidelines](https://github.com/skulpt/skulpt/blob/master/CONTRIBUTING.md)? Might be nice to add a section about what a *good* PR looks like.

albertjan:
Interesting read: http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests/

bnmnetp:
I'm not sure it matters if its the second person that accepts the PR, as long as we have comments from at least two people to the effect that yes, this one is good enough, and nobody is strongly saying, I don't think this should be accepted.

bnmnetp:
Nice article @albertjan.  I experienced some of those things first hand when I made a PR to another repo last week.  I felt like I was running the gauntlet between two guys who could accept or reject the PR.  At times getting conflicting comments from the two of them.

I especially like the "keep it small" approach.  #387 is a great case study.

* It is not small, it is huge 
* it addresses many issues **and** adds a new feature
* In large projects like this I would like to see us create a feature branch, so that we can accept small PR's towards the larger project.
* once all of the work is done, then the feature branch can be merged with master.

Now, the challenge for us in this is that many of these projects arise spontaneously. Someone gets annoyed enough to actually start on the project and so they work on it in private until some progress is made and then create a PR.  I'm not sure how to change that.

[...]

PR #835, opened 2019-03-23
Fix dotted imports

meredydd:
Looks like #833 broke imports deeper into heirarchies, and our test suite didn't catch it. Oops. It broke Anvil pretty hard, though, and will probably other people using current master too, so let's get this merged and restore my honour ;)

This PR adds tests that would have noticed this bug, and then fixes it.


albertjan:
✅ honor restored

rixner:
While I don't think it matters for this pull request, but have we given up on 2 people approving a pull request before we accept it?  (To be clear, I was about to approve this one and am not quibbling with its acceptance.)

bnmnetp:
I think we said something like for relatively minor changes if a committer and one other reviewer agreed that it was OK to merge with just one review.

I worry a little bit that the two review policy was the start of the great slowdown in progress on the master branch.  It is easier to keep your own fork moving forward if PRs are languishing on master.  I don't think that the master should be a dumping ground, but I do think we make better progress if changes get onto master quicker for others to take advantage of.  We may just need to get to a place where we are tagging actual solid release points.

meredydd:
I agree with @bnmnetp that the "two-reviewer rule" appears to have slowed us down a lot. (Also that this particular fix didn't need two reviewers even under our existing rules.)

I think for most (small-to-medium-sized) changes authored by a maintainer, (especially with tests that exercise the functionality), one review from another maintainer should suffice - ie the author can count as one of the two reviewers.

But perhaps that's just selfish - it would unblock us personally, but we might still end up with third-party PRs languishing forever.

bnmnetp:
This all started with #414 -- which is still open... Maybe we should continue this discussion there to decide if we want to streamline things a bit now that we have some momentum again!

Issue #834, opened 2019-03-20
String method .format() does not properly repr certain values


acbart:
The following code:

```
print("{}".format([1,2,3]))
```

Produces the following output:

```
[object Object],[object Object],[object Object]
```

Instead of the expected:

```
[1,2,3]
```

PR #838, opened 2019-03-26
Fix string formatting

rixner:
This addresses #834 but I'm not sure it's the best solution.  The string formatting code is fast and loose with Python vs. Javascript types, making it a bit messy.  I think I did manage to fix the reported bug (and a few related bugs) without breaking anything else, though.  But I'm not happy with the back and forth conversions that were needed.

This would benefit from others running more code through it before accepting it, though, to make sure that I did not introduce any regressions that are not captured by the somewhat limited test suite.


meredydd:
LGTM! Merging.

PR #839, opened 2019-03-26
Convert callsim to not mess with arguments.

rixner:
This introduces callsimArray and callsimOrSuspendArray which takes a function to call and an array of arguments.  The array can then be passed on directly to apply/applyOrSuspend without needing to slice the arguments.  This helps the optimizer and eliminates the unnecessary array construction.

All calls to callsim and callsimOrSuspend have been replaced with calls to the Array versions and these functions have been dropped.

### DIFF
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -180,7 +180,7 @@ Sk.abstr.binary_op_ = function (v, w, opname) {
             if (wop.call) {
                 ret = wop.call(w, v);
             } else {
-                ret = Sk.misceval.callsim(wop, w, v);
+                ret = Sk.misceval.callsimArray(wop, [w, v]);
             }
             if (ret !== undefined && ret !== Sk.builtin.NotImplemented.NotImplemented$) {
                 return ret;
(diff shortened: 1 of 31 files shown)


rixner:
Rebased and fixed conflicts after @albertjan approved.  Nothing changed.  I'm not sure why git thought there were conflicts as there were changes on distinct lines from the different branches, so I just had to pick and choose about 6 lines from each.

rixner:
@bnmnetp no leaving them does not cause trouble, so I can easily put the old ones back.  (I put the removal in it's own commit at the end, so I can just back that one out trivially.)  My motivation for taking them out was just so that no one adds new calls to them going forward.  It's relatively easy to change things to make it work with the new versions, and would be a one time fix.

I'll do whatever people think is best for the project.

bnmnetp:
I am fine with eliminating it. Those who write their own modules are more than capable of making the change.  having it break is the most efficient way of notifying them.  😁

meredydd:
Gnurk! I blinked and all my code broke :-P

I'll make a PR to reintroduce `callsim` and friends, and perhaps that can be the venue to discuss our policy on changes like this!

meredydd:
(PS: I want to make it very clear that the actual meat of @rixner's speed increase gets two thumbs up from me!)

PR #840, opened 2019-03-30
Reintroduce callsim and callsimOrSuspend, to avoid gratuitously breaking users

meredydd:
This is a straight revert of the last commit from #839, as it broke my code (and it's a pretty small compatibility shim to carry, as was discussed in the original PR). I'm happy to have it accepted or rejected, depending on how much we decide we care about API stability.

The real question is what our attitude should be towards API-breaking changes like this. For starters, is any of the Skulpt "user API" intended to be stable? Should we keep a changelog when we make a breaking change?

I have personal opinions (yes and yes, respectively), but I see three viable policies:

1. Skulpt's API is unstable. We deliberately commit breaking changes at will, and if you use Skulpt you'd better review every PR yourself (or have 100% test coverage), else your app _will_ break.

2. The Skulpt API changes rapidly, and anyone who uses it should be prepared to make updates whenever they update Skulpt. We keep a changelog of breaking changes you might need to adapt to - you can check them when you update, so you'll know what to grep for and fix in your code. (We'll miss things by mistake, of course, so you'll still need to test, but the changelog will help you catch most of them.)

3. The Skulpt API is stable-ish, and we avoid making breaking changes unless it's necessary for some other goal we have. We keep a changelog of these breaking changes so you'll know what you need to adapt. (We'll miss things by mistake, of course, so you'll still need to test, but we try to keep the rate of these things down.)

I'd argue that we should avoid (1). It was very easy for me, as a currently-active committer, to miss the breaking PR -- it came and went in half a week, and I didn't get any notifications from Github. Silent users (of which we may have a few) don't stand much chance, and nobody has 100% test coverage.

Basically, #839 implemented policy (1); this PR implements policy (3), but I am willing to go with (2) if that's what others prefer.

(Tagging all @skulpt/committers for the policy question)

### DIFF
--- a/src/misceval.js
+++ b/src/misceval.js
@@ -814,6 +814,16 @@ goog.exportSymbol("Sk.misceval.callOrSuspend", Sk.misceval.callOrSuspend);
 
 /**
  * @param {Object} func the thing to call
+ * @param {...*} args stuff to pass it
+ */
+Sk.misceval.callsim = function (func, args) {
+    args = Array.prototype.slice.call(arguments, 1);
+    return Sk.misceval.apply(func, undefined, undefined, undefined, args);
+};
+goog.exportSymbol("Sk.misceval.callsim", Sk.misceval.callsim);
+
+/**
+ * @param {Object} func the thing to call
  * @param {Array=} args an array of arguments to pass to the func
  *
  * Does the same thing as callsim without expensive call to Array.slice.
@@ -835,6 +845,17 @@ Sk.misceval.callsimAsync = function (suspensionHandlers, func, args) {
     return Sk.misceval.applyAsync(suspensionHandlers, func, undefined, undefined, undefined, args);
 };
 goog.exportSymbol("Sk.misceval.callsimAsync", Sk.misceval.callsimAsync);
+
+
+/**
+ * @param {Object} func the thing to call
+ * @param {...*} args stuff to pass it
+ */
+Sk.misceval.callsimOrSuspend = function (func, args) {
+    args = Array.prototype.slice.call(arguments, 1);
+    return Sk.misceval.applyOrSuspend(func, undefined, undefined, undefined, args);
+};
+goog.exportSymbol("Sk.misceval.callsimOrSuspend", Sk.misceval.callsimOrSuspend);
 
 /**
  * @param {Object} func the thing to call


bnmnetp:
My original instinct was to maintain backward compatibility.  Maybe it was Friday afternoon impatience, or the fact that @albertjan and @rixner seemed to be in agreement on changing the API that caused me to merge.  As I pointed out in the discussion it did break one of my modules, but the fix was really easy.

I think the reality is that skulpt is stable-ish, but not necessarily by policy.  It just hasn’t changed that much in the last several years.  Our API isn’t really documented so I don’t feel like we have made promises to other developers and users, other than dig into the code and figure it out yourselves.

If #2 helps us make better progress toward a python3 future then I think we should go with 2 along with a goal of providing better documentation for internal developers as well as for integrators.  Somewhere there is an old issue that I think I wrote proposing a new API for integrators that never gained traction.

rixner:
@meredydd my problem with your position is that there is no documentation of which functions are sacred and which are not.

Your (and @albertjan 's) pull requests routinely break my code.  I fix it and move on.  If we are going to have a policy that some particular "API" is fixed, that API has to be clearly documented somewhere so that (a) we all agree on what is inviolable, (b) we know not to change it, and (c) there is grounds for discussion for a change if/when it is warranted.

I will argue that this particular change was trivial.  I put it in it's own commit so I could back it out if others disagreed (as you did, which is fine).  But, furthermore, you could have fixed it in your code with very little effort.  You could have put those two functions back in your code if you don't care about the performance impact.  Or you could have used grep and fixed your other code pretty quickly.

I wanted that change in there for the following reason: I do not want anyone introducing new calls to those functions within the core of Skulpt.  What others do in their own forks and libraries is their business.  But, if those remain in, then others can create a pull request with calls to them, I "blink", and then I get a performance regression that I need to fix.  I think what happens in the core and what happens outside are different, and unfortunately, those functions are used in both cases.

So, my position is as follows: I think we should either document the unchangeable "API" or we should not be using some reference to it as grounds for complaint when functions change.  (Obviously, that doesn't mean that without documentation any old change should get accepted, the approvers would still need to agree it is okay.)

For example was the external module code that @albertjan removed in #827 part of the fixed API?  I would argue that it was.  (Note that I have no problem with it's removal, but am making a point that we have no documentation for what is and is not part of the "API" that others can rely on and this is clearly something that others could have been relying on.)  So, how was that any different?

rixner:
I want to be really clear that I am not upset in any way that this was accepted.  I was willing to do this in the original pull request.  I just want to be clear that I think we need to either document an API or accept that everything in Skulpt is a moving target.  I am **not** complaining about anything that has happened up to this point.

rixner:
@meredydd quite a while ago, when I did serious profiling, the attribute infrastructure was a major time consumer.  I therefore reworked it a bit and disabled some features.  It seems like a lot of the pull requests interact with that code, and break what I did.  I update things, include what I want and leave out what I don't, each time.  I haven't thought about whether I can decouple the things that would be accepted back into the project from those that wouldn't and whether that would ease the situation for me.

I think the situation is a little more complex, as I think there are 3 (or more?) types of users of Skulpt:

1) The person who is just trying to use Skulpt as-is within their application.  They largely interact via the configuration settings.  Even there, though, it seems like there are recurring questions from people about how to do that.  And when suspensions were introduced, it got even harder.  I doubt, for instance, that I handle "pausing" execution in the manner that you intended, but there is no real documentation or examples about how one is supposed to do that.  This is one undocumented API.

2) The person who is trying to add their own custom libraries.  They interact with this mythical "user API", which quite frankly is as large as every closure exported function without limitation.  While presumably most libraries only use a subset of those functions, it would seem that which ones would completely depend on what they are trying to do in those libraries. There have been people in the past, I believe, who have really intricately tied their libraries with the core of Skulpt, for either performance or functionality reasons. This is another undocumented API.

3) The person who modifies the internals of Skulpt for their own purposes.  Such as Trinket or myself (and you?).  This potentially even includes people adding functionality within the core itself.  There really is no API, and as my recent work on formatting.js reminded me, we aren't consistent in the core - instead of using ```Sk.ffi.remapToJs``` that code just inappropriately accesses ```.v``` everywhere.  There is arguably a missing "API", or at least set of rules about what object boundaries are, here.  We just sort of have a loosely agreed upon set of principles.

While we probably agree that group 3 is on their own, even that decision limits what they might give back to the project, though.  Especially if we limit what can be changed.

The question in my mind, then, is if others agree that groups 1 and 2 are distinct, if they agree we should define a stable API for either or both of those groups, and how we might even go about doing that...

rixner:
(And I will add that one concern I have is that the more rules/restrictions we place on pull requests, the fewer we will get - from others, from us, etc.)

bnmnetp:
Good summary @rixner 

For 1.  I would refer you all to the very old #478 which describes a simpler version of what we have today.   The externalLibrary interface was documented on the skulpt.org page.  But I took it down now since its gone.  I'd love to try to bring back a better mechanism for that as I've had to incorporate several custom 3rd party modules into my private build of skulpt-stdlib.  I have my own implementation of a fileReader that allows Runestone users to read "files" that are stored in our backend database.  

But-- speaking of performance killers -- I had to exclude files that ended in .js or .py because of the way import systematically tries all the different combinations of .py and .js files when it wants to import a module, and if one wasn't found in the local stdlib it also tried to get it from the DB.  I think I've nearly lost the thread here, but a simple way of registering modules to be retrieved from a known URL would be a great goal.

For 2.  http://www.skulpt.org/ProgMan was an attempt to make use of jsdoc to document/capture the API that would be needed by people writing libraries.

I think that having and sticking to a well defined API on those two levels would be helpful in getting more people helping to write and supplement the standard library.

For 3, I agree that there does not seem to be any documented stable API.  Even doing more to enforce documenting existing internal definitions could go a long way there.  But I 100% agree with @rixner 's point that the more rules and restrictions we add the fewer PRs we are likely to get.

I'm pretty sure we all agreed to eliminate the use of `.v` a few years ago.  What we haven't done is a sort of big old cleanup where we have a PR that cleans up all the existing code.  I don't think it would be hard to produce such a PR, but I know it would likely cause conflicts with many existing PR's.

bnmnetp:
Also FWIW, I merged the big reunification PR because I thought it was a decent tradeoff to get us back on track, even though I would have preferred to keep the externalModules interface.  @albertjan kindly offered to sketch out a strategy for how we might replace it.

But, I felt that we were better off finding a new better way to reimplement externalModules than to have a PR like that just sit there, like too many others have done over the last couple of years.  Maybe a new separate issue just to design such a thing would be a good idea.

Issue #842, opened 2019-04-02
File reading bug


bnmnetp:
An instructor reported this bug to me today which I have confirmed:

```
F=open("file.txt")
header=F.readline()
for line in F.readlines():
   print(line)
F.close()

print("="*60)
F=open("file.txt")
header=F.readline()
for line in F:
   print(line)
```
Cpython prints 2 and 3 in both cases
```
2

3

============================================================
2

3
```
But skulpt gives the following output, resetting the file pointer on the iteration
```
2

3

============================================================
1

2

3
```

PR #843, opened 2019-04-02
fix for file iterator to start and current line of file

bnmnetp:
Fix for #842 

This seems like a pretty straight forward fix.  Set the initial line in the iterator to the current line rather than 0.


bnmnetp:
Hmm, the tests pass when run locally but travis fails to find the file..  Any suggestions on how to better test this?

bnmnetp:
Duh 🤕 I had an extra copy of the file that was fooling me.

PR #844, opened 2019-04-12
Update the "Skulpt in the Wild" section

ludwiktrammer:
Remove broken links, add Code Shrew


ludwiktrammer:
@albertjan It seems that a new, much nicer, skult.org has launched. Unfortunately it features content from before this pull request - so the "Skulpt in the Wild" section now includes multiple dead links and the mention of Code Shrew is gone.

bnmnetp:
Sorry about that @ludwiktrammer.  I made those changes and pushed them again.  the link to Code Shrew is restored.

If you would like to make any further updates to the new site, you can do so against the gh-pages branch of skulpt.

Brad

ludwiktrammer:
Thank you!

PR #850, opened 2019-04-19
Bug fix for determining if 0 args

### LINKED ISSUE #849: Turtle crashes on bare call to `pencolor()`

bnmnetp:
This was originally to fix #849 

It seems that in the current system arguments.length is not 0 if no arguments are passed.  When pencolor is called with no arguments, arguments.length is 5 

This seems a more reliable way of dealing with it and should fix the call with no args across the board.

### DIFF
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -849,7 +849,7 @@ function generateTurtleModule(_target) {
         };
 
         proto.$speed = function(speed) {
-            if (arguments.length) {
+            if (speed !== undefined) {
                 this._speed          = Math.max(0, Math.min(1000, speed));
                 this._computed_speed = Math.max(0, speed * 2 - 1);
                 return this.addUpdate(undefined, false, {speed:this._computed_speed});
@@ -861,7 +861,7 @@ function generateTurtleModule(_target) {
         proto.$speed.co_varnames = ["speed"];
 
         proto.$pencolor = function(r,g,b,a) {
-            if (arguments.length) {
+            if (r !== undefined) {
                 this._color = createColor(r,g,b,a);
                 return this.addUpdate(undefined, this._shown, {color : this._color});
             }
@@ -873,7 +873,7 @@ function generateTurtleModule(_target) {
         proto.$pencolor.returnType = Types.COLOR;
 
         proto.$fillcolor = function(r,g,b,a) {
-            if (arguments.length) {
+            if (r !== undefined) {
                 this._fill = createColor(r,g,b,a);
                 return this.addUpdate(undefined, this._shown, {fill : this._fill});
             }
@@ -885,7 +885,7 @@ function generateTurtleModule(_target) {
         proto.$fillcolor.returnType = Types.COLOR;
 
         proto.$color = function(color, fill, b, a) {
-            if (arguments.length) {
+            if (color !== undefined) {
                 if (arguments.length === 1 || arguments.length >= 3) {
                     this._color = createColor(color, fill, b, a);
                     this._fill  = this._color;
@@ -1019,7 +1019,7 @@ function generateTurtleModule(_target) {
         proto.$write.minArgs     = 1;
 
         proto.$pensize = proto.$width = function(size) {
-            if (arguments.length) {
+            if (size !== undefined) {
                 this._size = size;
                 return this.addUpdate(undefined, this._shown, {size : size});
             }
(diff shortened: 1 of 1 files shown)


rixner:
I don't generally use turtle, so I'm going to ask rather than try to figure it out for myself.  What are the 5 arguments?

bnmnetp:
I presume self and then r, g, b, a

All arguments have the value undefined ... 

Some calling mechanism must be setting them to undefined so that it is defeating the check of arguments.length.

rixner:
I don't necessarily see a problem with this solution.  However, it seems like it would be better to figure out the underlying cause and solve that.

bnmnetp:
I'm not sure if the underlying problem is a bug or a feature.  @albertjan @meredydd Any thoughts on this??

albertjan:
What is in `arguments` when `arguments.length` is `> 0`?

bnmnetp:
@albertjan for a call to `t.pencolor()` the arguments end up as `[undefined, undefined, undefined, undefined, undefined]`

albertjan:
Ah ok, let me check, it's probably the right thing to do! I'll get back to you tomorrow.

albertjan:
@bnmnetp ah yes this is due to the new way of handling kwargs I overhauled it because the original implementation wasn't exactly doing the same thing as python. See this commit https://github.com/trinketapp/skulpt/commit/08e9abebed3eb6255243bc68a5a0d55149900990

bnmnetp:
OK, I'm going to take @rixner 's comments as approval since there is not some other underlying problem.  And this is pretty simple anyway.

rixner:
Is this going to affect things beyond turtle?  If so, I would like more justification for incorporating the kwargs change without fixing everything else.  It seems as if this may mean that we can no longer effectively check how many arguments are passed to a function.  Which is bad.  Really bad.

albertjan:
@rixner this is due to the implementation in turtle. For example the function @bnmnetp was talking about has 4 optional arguments with `None` as default. In this case skulpt does the correct thing passing 4 `None`'s one for each argument. Python does the same thing: 
```
>>> def func(a=None, b=None, c=None):
...     frame = inspect.currentframe()
...     _, _, _, values = inspect.getargvalues(frame)
...     print(values)
... 
>>> 
>>> func()
{'frame': <frame object at 0x10bff3048>, 'c': None, 'b': None, 'a': None}
>>> 
```

albertjan:
@rixner Compare this code with codeskulptor:

```
def test(a=None, b=None):
    print (a, b);
    
test(10, a=1)
```

on code sculptor this outputs `(1, None)` 

with the args fix on master it throws see https://trinket.io/python/08588306fb

```
JavaScript debugger enabled
Python 2.6(ish) (skulpt, Wed Apr 24 2019 17:10:17 GMT+0100 (BST))
[v8: 3.25.30] on a system
>>> def test(a=None, b=None):
...     print (a,b)
... 
>>> test(1, a=10)
TypeError: test() got multiple values for keyword argument 'a' on line 1
1>: test(1, a=10)
>>> 
```

albertjan:
This is because the original implementation didn't fill the array with the default parameters, before assigning the kwargs to the fitting parameters.

rixner:
@albertjan I wasn't quibbling with the value of the fix.  Rather I am/was worried about the potentially unintended consequences of the fix...

Is the issue then that ```undefined``` is being used to represent Python ```None```, instead of using ```Sk.builtin.none.none$```?

I feel like if those arguments were supposed to be ```None```, they should actually be Python ```None```, not ```undefined```, which theoretically means the absence of a passed argument.

albertjan:
@rixner it is using `Sk.builtin.none.none$` but these get automatically translated to `undefined` here: https://github.com/skulpt/skulpt/blob/master/src/lib/turtle.js#L2148 

This change in the arg handling is a breaking change though, `co_varnames` has to include all the variable names and `$defaults` should contain the defaults for the arguments. These are right justified, so: 
```
func.co_varnames = [ "a",  "b",  "c" ];
func.$defaults = [ Sk.builtin.none.none, $Sk.builtin.none.none$ ];
```

this is the same way python stores the defaults in a function object.

albertjan:
This is for a function that looks like this in python:

```
def func(a, b=None, c=None):
    pass
```

rixner:
Okay, so this ```undefined``` problem is confined to turtle, then?  In all other instances, functions will properly receive ```Sk.builtin.none.none$```?  In that case, I have no objections about anything.  (Not that it matters, since this has already been pulled. :) )  Thanks for explaining it @albertjan !

PR #857, opened 2019-04-26
Feat/es6 modules

albertjan:
A what I think failed attempt to move skulpt to a more modern javascript. I ran into so many recursive dependencies. And there are no ways or dictating load order in es6/node. 

Lessons should be learnt from this, and we should try again. But with more of a plan. 

@rixner @bnmnetp @meredydd @eah13 @j-f1 

see #666


rixner:
Presumably this requires refactoring a bunch of code.  I assume that files like misceval and abstract are issues as they aggregate a bunch of utility functions.  I stared at the svg file for a while, but it's just too big to really see the cycles.  Do you have any suggestions about where to start?  And are you suggesting we start from here, or we start over?

albertjan:
@rixner I'm unsure how to proceed. Everything depends on everything. Worst offenders are definitely abstract and misceval. But I mean there's just no hierarchy at all in the object model, there is no rhyme or reason to the dependencies. And I don't believe this is our fault it grows this way and it works, because we currently have a clear distinction between runtime and parse time, but with the new class inheritance you have to have compiled the base before you can inherit from it and this happens as it's compiling. 

I think we need something like compile time macro's in babel to do common transformations like jsfunc to pyfunc and we should have a value of boolean that's a javascript constant that compares to the python runtime constant like a bool would. In cpython builtin functions aren't real functions either but a c object that looks and acts the same as a real python function.

Because you can't change the load order in node or web pack and they are not the same we will have to solve it in a way that doesn't rely on the implementation. 

We might have to split up the shell of the object and some of the implementation, this way object can be constructed to be the base of other objects before we patch the prototype to add functions that depend on things further down the dependency tree.

I hope I make sense. 😄

albertjan:
I think continuing with this will be quite difficult 😄 but if we're going to attempt this again we should definitely learn from this. It contains some valuable lessons.

rixner:
@albertjan I think I understand what you are saying.  I am inclined to try myself, too, just to see some of the problems first hand.  Of course, I'm not sure I have time to "waste" on that.  Looking at what you did, it's a little bit difficult to take away any concrete lessons.  (Just because there is so much to look at.)

Regardless, as time goes on, I have become increasingly annoyed I can't use node and webpack.  I have no end of trouble with d8, and can't even get the debugger to work properly on my machines anymore.  So, while everything works the way we have it, I really think using more modern infrastructure would be worth it.

It sounds like the biggest issue is we can't really do this incrementally?  It might be more palatable if we could figure out a way to slowly work towards this...

rixner:
@albertjan I just spent the last 2 hours staring at what you did.  I think I get at least some of the complexity.  Having said that, I don't really have any good ideas on how to proceed further.  Regardless, doing this is going to completely change the code base and cause a lot of disruption.  I'm still in favor, but you're right, we need some principled strategy if it's ever going to work.

It's going to be tough to keep other changes in sync with something like this, though, as it will likely take long enough that other development will continue on the current code base and then be difficult to port over to the new one.  So, I'm not sure how we manage that process either.

Not to mention the difficulty of everyone getting their custom code rewritten to work with all of this, as well.

Sigh.  Maybe it's just a pipe dream.

albertjan:
This was an attempt at doing this all the way, it might be worth to try to do the least amount of changes to make it work in node and (webpack and babel later). That would at least give us debugger and modern js, and we'd loose the d8 dependency. We could slowly break out pieces into separate files and modules. We'd be bad citizens of node, but if it gives us something I'm not so worried about that.

With bad citizens I mean that `require('skulpt')` would not return something but inject `Sk` into the global scope.

src/types/long.js
```diff
@@ -0,0 +1,860 @@
...
+import { isIndex, asIndex } from '../misceval';
+import { numtype } from './numtype';
+import { str } from './str';
+import { int_, str2number } from './int';
+import { float_ } from './float';
+import { tuple } from './tuple';
+import { true$, false$ } from '../constants';
+import { bool } from './bool'
+
+import biginteger from 'big-integer';
```
albertjan: @rixner this is the package I was talking about.

rixner: @albertjan I'm glad we're on the same page.  I suppose we'd need to discuss which package makes the most sense, but that's a future discussion...

albertjan: Yeah I picked one that was actively maintained had a lot of downloads and was a polyfill for `BigInt` I think it's actually a future version of what we use :smile:

PR #859, opened 2019-05-07
Normalize get/set attribute infrastructure to take Python attribute names

rixner:
In this pull request, all of the get/set attribute functions are converted to take Python strings for the attribute names.  To make it more clear when a variable holds a Python string or a JS string, all variables holding attribute names are called ```pyName``` or ```jsName``` as appropriate.  Python strings for attributes are created once in ```constants.js``` so that they can be used directly.

There are two reasons for this change:

1. The attribute infrastructure can sometimes be relatively deep, and functions were converting names back and forth between Javascript and Python.  This pull request normalizes things so that you always pass down a Python string everywhere.  Previously, the internal functions took Javascript strings and the Python-facing functions took Python strings.  But, these functions have to call each other, and therefore convert their attribute names.
2. The ```Sk.builtin.str``` function is relatively heavyweight.  So, it's expensive to convert a Javascript string into a Python string.  The reverse is very cheap - just grab the ```.v``` attribute.  Given that in many/most instances the Python string object existed somewhere, it was a complete waste to reconstruct it elsewhere.  I added ```Sk.builtin.str.$jsstr()``` to extract the Javascript string from the Python string, rather than directly accessing ```.v```.  Using ```Sk.ffi.remapToJs``` was a waste, as you know you have a Python string, so the type checks in there are unnecessary.

When I originally did this, it yielded a substantial performance benefit for me by eliminating an enormous number of calls to ```Sk.builtin.str```.  I did not rerun any performance tests here.  I feel that whether it still has gains or not, it is still the right thing to do and it makes reading, understanding, and extending the attribute infrastructure much easier (in my opinion) regardless.

For external libraries that explicitly get/set attributes or provide their own get/set attribute functions, this is potentially a breaking change.  You would now have to accept and pass Python string objects as the attribute names.  It is not possible to provide backward compatibility, nor do I think it makes any sense to do so.  We either agree to normalize things this way or we don't.  Leaving some infrastructure the old way is messier than just leaving it all alone, in my opinion.

The one thing that might be an issue is that the test coverage of the libraries is quite light compared to the core.  So, it is possible that there is infrastructure in a library that is not getting tested that does operate on attributes in an incompatible way.  While I have been using this successfully, I could not just cherry pick some commits out of my repository.  The attribute infrastructure changes a lot (it seems like nearly every pull request by @meredydd or @albertjan touches it in some way), so I have repaired things continuously for a long time.  Meaning, I may have missed some modifications that are not explicitly tested by our test suite.  And I do not use all of the modules in ```lib``` (most notably turtle), so those could have issues, too, if they have insufficient test coverage.

### DIFF
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -479,7 +479,7 @@ Sk.abstr.sequenceContains = function (seq, ob, canSuspend) {
      *  Look for special method and call it, we have to distinguish between built-ins and
      *  python objects
      */
-    special = Sk.abstr.lookupSpecial(seq, "__contains__");
+    special = Sk.abstr.lookupSpecial(seq, Sk.builtin.str.$contains);
     if (special != null) {
         // method on builtin, provide this arg
         return Sk.misceval.isTrue(Sk.misceval.callsimArray(special, [seq, ob]));
@@ -676,7 +676,7 @@ Sk.abstr.objectFormat = function (obj, format_spec) {
     var result; // PyObject
 
     // Find the (unbound!) __format__ method (a borrowed reference)
-    meth = Sk.abstr.lookupSpecial(obj, "__format__");
+    meth = Sk.abstr.lookupSpecial(obj, Sk.builtin.str.$format);
     if (meth == null) {
         throw new Sk.builtin.TypeError("Type " + Sk.abstr.typeName(obj) + " doesn't define __format__");
     }
@@ -796,21 +796,22 @@ Sk.abstr.objectSetItem = function (o, key, v, canSuspend) {
 goog.exportSymbol("Sk.abstr.objectSetItem", Sk.abstr.objectSetItem);
 
 
-Sk.abstr.gattr = function (obj, nameJS, canSuspend) {
+Sk.abstr.gattr = function (obj, pyName, canSuspend) {
     var ret, f;
     var objname = Sk.abstr.typeName(obj);
+    var jsName = pyName.$jsstr();
 
     if (obj === null) {
-        throw new Sk.builtin.AttributeError("'" + objname + "' object has no attribute '" + nameJS + "'");
+        throw new Sk.builtin.AttributeError("'" + objname + "' object has no attribute '" + jsName + "'");
     }
 
     if (obj.tp$getattr !== undefined) {
-        ret = obj.tp$getattr(nameJS, canSuspend);
+        ret = obj.tp$getattr(pyName, canSuspend);
     }
 
     ret = Sk.misceval.chain(ret, function(r) {
         if (r === undefined) {
-            throw new Sk.builtin.AttributeError("'" + objname + "' object has no attribute '" + nameJS + "'");
+            throw new Sk.builtin.AttributeError("'" + objname + "' object has no attribute '" + jsName + "'");
         }
         return r;
     });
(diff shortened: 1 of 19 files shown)


meredydd:
I love the idea of this; the disparity between `tp$getattr()` and `__getattr__` has annoyed me for a while. I also greatly approve of using `pyXyz` and `jsXyz` to flag what type of string we're playing with here.

It's a big PR, though, and I'm just back from PyCon so can't review today. I am hereby committing to review it by the end of the week, though.

albertjan:
I think I've grokked this it looks good to me!

albertjan:
There are definitely some in the wild `__getattr__` implementations that will break because of this. But it's a price we'll have to pay. We should mark this as a breaking change.

albertjan:
Added a new label "breaking change" it might be a good idea to start adding them to PR's where we change the api. 😄

rixner:
Rebased.

eah13:
RE breaking changes: let's make sure to [SemVer](https://semver.org/) this.[1]  LOTS of merged PRs recently, so seems like well past time to bump 0.12.0, and I propose we do that as of this PR.  (There hasn't been an official release in about 2 years...no wonder people are getting out of date stuff out of skulpt-dist.  If they're like me they look on the releases page to get files from a project.  So many of us build from our own forks that it's easy to forget this.)

The two year changelog is now huge but a quick summary of PRs that may break modules (labeled as such, as @albertjan suggests) would be enough. Few changes break otherwise working user code, though, so many Skulpt consumers won't need to worry.

While the dev momentum keeps up, "patch" releases (0.0.Z) could happen every month or two, with manageable changelogs.

Maybe worth doing a retroactive 0.11.Z as of sometime before the past 1-2 months' spate of potentially breaking work too.  So people can fall back easily if there are problems.

Sorry to add this comment to a PR, but I am just reentering skulpt land.  I'm still (kinda) on parental leave but will try to get caught up with all the (awesome) stuff happening in the repo best I can.  If versioning/releases is a pain but there's consensus on the rules I can also look at taking some of that on.  But ideally all committers would keep an eye on breaking changes as 0.Y.0 candidates (and tag them) and 0.0.Z versions just happen on a schedule. 

This is not intended to hold up merging this PR, just to suggest an official release happen now or soon.

- [1] "Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable." - So at the current state of Skulpt MINOR 0.Y.0 version bumps would indicate potential backwards incompatibility with third party js modules.

meredydd:
Congratulations, @eah13!

I agree re releases; I'm planning to get my teeth into this review this weekend. (Was going to be on the train this evening, but then I left my laptop at home... :-/)

albertjan:
It's a bit of a loose classification whether something is a breaking change, we should let this hold us back. We don't have a clearly defined public/private interface, so some things may be considered non breaking by us because it's affecting a private interface. But this private interface may be used by someone out there (esp. us mayor users :smile:).

bnmnetp:
Merging this now....  We can always fix anything that other @skulpt/committers may find.

alexwenbj:
Hi,@rixner ,When I used this version skulpt with [pygal](https://github.com/trinketapp/pygal.js/blob/master/__init__.js) ,it always raised an exception in the render() function:
`TypeError: a.$jsstr is not a function. (In 'a.$jsstr()', 'a.$jsstr' is undefined) `
For example,the code below:
```
import pygal
date_chart = pygal.Line()
date_chart.x_labels = list(range(2010,2017))
date_chart.title = "populations"
date_chart.add("p1", [7869,7899,7920,7939,7960,7976,7999])
date_chart.add("p2", [6865,6899,6920,6939,6860,6676,6399])
date_chart.render() #<-Here raise an exception
```
It seems caused by this PR change.Can you give me a hint for this?Thank you.

alexwenbj:
It seems most of the third-party libs should refactor,for example,the numpy.random lib,when I `import numpy.random` will raised the same exception.Please give us a workaround to fix them.Thank you very much. @rixner @ebertmi

rixner:
The "breaking change" that was made in this pull request is that setting/getting attributes now require you to pass Python strings for the attribute name, instead of Javascript strings.  So, ```GenericGetAttr```, ```Sk.abstr.gattr```, ```tp$getattr```, etc. (and the equivalent functions to set attributes) all need to take Python strings instead of Javascript strings.

alexwenbj:
Got it,Thank you. @rixner

Issue #729, opened 2017-10-27
Long addition error


rixner:
There seems to be some sort of error in the long implementation:

```python
# Expected: 10821521012635269
# Skulpt: 10821521012635268
a = 8000000000000000
b = 2821521012635269
print a + b
```

I haven't had time to look into the issue, but wanted to report it so that it doesn't get lost.  I haven't seen this in other cases and I'm not sure if there's anything special about one or both of those particular numbers.

albertjan:
I think I know what's going on here. 😄 

It's not a skulpt problem it's a javascript problem:

```
> (8000000000000000 + 2821521012635269) - 2821521012635269
< 7999999999999999
```

j-f1:
Maybe something like [BigInteger.js](https://github.com/peterolson/BigInteger.js) would help here by letting us store large integers that JS doesn’t like.

albertjan:
It would I think, but this is too good of a problem not to figure out 😄

albertjan:
I'm pretty sure this is the problem:
```
> (2**53 + 1) === 2**53
< true
```

albertjan:
We just have to think of a performant way to solve this. The idea is that if the result of the addition is larger then `2**53` we have to pass it on to biginteger.js

rixner:
Nice catch @albertjan .  I agree, it looks like this is a question of how to modify the ```Sk.builtin.int_.prototype.nb$*``` functions to check for overflow without killing performance...

rixner:
I have students who continually trip over this.  Since there is no longer a long type in Python 3 and BigInt is coming to Javascript, does it make sense to rewrite our int class to use BigInt (https://github.com/GoogleChromeLabs/jsbi until it is native everywhere, at which point we should be able to convert to use it relatively easily) and toss the long type?

I guess one issue is that I suspect that using this always for ints will result in slowdowns for small integer math.  And it's clear that switching back and forth appropriately is non-trivial or one of us would have fixed this bug long ago.

rixner:
Yeah, it seemingly kills small integer performance: https://peterolson.github.io/BigInteger.js/benchmark/

That page doesn't show what the native BigInt does, though.

So, we would still need to fix the way we detect when we should be using big integers and when we should not be.

rixner:
Never mind, the fix isn't that bad (#860).  I still think that in the long run we should move towards eliminating long and using ```BigInt```.

albertjan:
[...]

PR #860, opened 2019-05-09
Check for integer overflow on add/subtract and fall back to long math.

rixner:
Add some simple checks to determine if add/subtract result overflows integers and use long add/subtract when it does.  This fix was so simple, I'm not sure why we didn't do it long ago...

I ran a simple performance test, and using d8, I see about a 2% overhead to doing these checks on a long running while loop that does nothing but add/subtract numbers that don't overflow.

Fixes: #729


albertjan:
@rixner feel free to merge this I deem this a small change 😄

PR #774, opened 2018-03-31
Slight formatting changes in README

harsh183:

PR #467, opened 2015-06-01
Update using.html

eah13:
A more polished draft of the Using Skulpt page.  As discussed in #454 it's split into three sections of increasing complexity.

The elements I'd love comments on that will be used on other pages if they're generally seen as successful:
- the floating nav bar that appears when you hit the first section
- the callout panel at the top

This is by no means a final draft. Merging this PR will not deploy to production, just take another step towards the final layout/content. I've not done much editing inside the three sections and expect to give those more love once the site in general is laid out.  So, in other words, layout and organization comments are more helpful than wordsmithing at this particular juncture. :)

**Update: I've temporarily deployed this PR to http://elliotthauser.com/skulpt/ for ease of review**

---

Since this is the first PR for the website, here are some quick instructions on building the site. ~~Let me know if anything's amiss- this is from memory.~~ Update: added gem and jekyll install instructions:

Make sure you have rubygems installed (`sudo apt-get install rubygems-integration` on Ubuntu 14 - or follow instructions [here](https://rubygems.org/pages/download)).  Then:

```
gem install github-pages
```

This is a meta-bundle that gets you everything you need for your setup to work like github's prod including jekyll and its various plugins.

Next, if you don't have a copy of the gh-pages branch you'll need to create an orphan branch to pull this code to:

```
git checkout --orphan pages
git rm -rf .    
git pull git@github.com:trinketapp/skulpt.git pages
```

If you already have gh-pages you can `git checkout -b pages` from that branch. 

With the code safely on your system you're now ready to

```
npm install
bundle install
bundle exec jekyll serve --watch
```

If these commands doesn't work for you, [see the pre-reqs](https://github.com/skulpt/skulpt/tree/gh-pages#skulpt-project-homepage)

This will build the site in the gitignored `_site` directory.  This jekyll build, by the way, is essentially what Github does when we commit a change to the gh-pages branch. That's why we don't need to include the built static site in the repo.

If you'll be making SCSS changes you'll want to open a new terminal tab and 

```
grunt
```

More on hacking in the [README](https://github.com/skulpt/skulpt/tree/gh-pages#skulpt-project-homepage)

[...]

PR #861, opened 2019-05-10
Experiment to replace d8 with node.

rixner:
This is an attempt to do the least modifications possible to replace d8 with node.  It is not using node in the manner in which node is intended to be used.  Instead, ```skulpt.py``` is modified to just call node everywhere we used to call d8.  This means there are temporary .js files being created by just concatenating everything together.  The concatenated file with minimal boilerplate is then run though node.

Anyone who does actual development using node is likely to cringe at this.  You have been warned.  It's ugly, but it mostly works.  I did not modify the internals of skulpt, except minimally in ```src/env.js```.  All the other modifications are in the infrastructure.

One issue is that I used the google closure library from npm for the test runs, and that seems to behave slightly differently than the ancient version we use (it adds ```base``` as a property of our classes).

I do not bundle anything from npm into the distribution output.  That is basically built the same way as before.

Effectively, what this accomplishes is makes it so that no one has to run d8, and can just install node on their local system to do development.  This represents the opposite end of the spectrum from what @albertjan was trying to accomplish with #857.

I only ran this on a linux system, I did no testing on any other system.  And I didn't try to run anything beyond the tests.  The question is whether people think this is a worthwhile short term step or a waste of time.  Regardless, it can sit here for people to learn from.  Maybe someone else can make a better second attempt...


package.json
```diff
@@ -41,8 +41,12 @@
   "license": "MIT",
   "private": true,
   "devDependencies": {
-    "jshint": "~2.5.2",
+    "fs": "0.0.1-security",
```
j-f1: It shouldn’t be necessary to manually install `fs`, since it’s included by default with Node. I believe this package only exists to prevent a bad actor from taking it over.

rixner: Yeah, that was a mistake.  Sort of beside the point of the experiment, though. :smile:

skulpt.py
```diff
@@ -225,7 +201,28 @@ def test(debug_mode=False, p3=False):
     ret4 = 0
     if not p3:
         buildNamedTestsFile()
-        ret1 = os.system("{0} {1} {2} -- {3}".format(jsengine, ' '.join(getFileList(FILE_TYPE_TEST)), ' '.join(TestFiles), debugon))
+
+        # Initial "requires"
```
albertjan: can we store the require.js file instead of building it everytime?

albertjan:
I like it! I just ran it it's so much faster! haha I do have two failing tests somehow. I think this removes a lot of cognitive overload with the strange v8 version we were running. If the others concur I'm overmuch in favour of pulling this in. 

We should specify the minimum node version we require I think see: https://docs.npmjs.com/files/package.json#engines

albertjan:
The `dir` of an object gained a `base` field: 
```
-----
GOT:
-----
['__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__repr__', '__setattr__', '__str__', 'a', 'b', 'base', 'c']
['__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__repr__', '__setattr__', '__str__', 'a', 'b', 'base', 'c', 'd']
['a', 'b', 'c', 'd']

-----
WANTED:
-----
['__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__repr__', '__setattr__', '__str__', 'a', 'b', 'c']
['__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__repr__', '__setattr__', '__str__', 'a', 'b', 'c', 'd']
['a', 'b', 'c', 'd']
```

albertjan:
The second failure is `most_common` is sorted differently on node. `t901.py`

rixner:
The first problem is an issue with the closure library, as I said: "One issue is that I used the google closure library from npm for the test runs, and that seems to behave slightly differently than the ancient version we use (it adds base as a property of our classes)."

I don't know if we modified our closure library (the issue doesn't occur when I use it) or if something changed between the version we use and the current one that gets installed by npm.

skulpt.py
rixner: Yes.  If we're serious about doing this, I can clean it up a bit.

albertjan:
Sorry @rixner I completely missed that bit. 😄

albertjan:
I've dug up the inherits function from the good library. It's simple enough we can change it and make our own. 

```
/**
 * Inherit the prototype methods from one constructor into another.
 *
 * Usage:
 * <pre>
 * function ParentClass(a, b) { }
 * ParentClass.prototype.foo = function(a) { };
 *
 * function ChildClass(a, b, c) {
 *   ChildClass.base(this, 'constructor', a, b);
 * }
 * goog.inherits(ChildClass, ParentClass);
 *
 * var child = new ChildClass('a', 'b', 'see');
 * child.foo(); // This works.
 * </pre>
 *
 * @param {!Function} childCtor Child class.
 * @param {!Function} parentCtor Parent class.
 * @suppress {strictMissingProperties} superClass_ and base is not defined on
 *    Function.
 */
goog.inherits = function(childCtor, parentCtor) {
  /** @constructor */
  function tempCtor() {}
  tempCtor.prototype = parentCtor.prototype;
  childCtor.superClass_ = parentCtor.prototype;
  childCtor.prototype = new tempCtor();
  /** @override */
  childCtor.prototype.constructor = childCtor;

  /**
   * Calls superclass constructor/method.
   *
   * This function is only available if you use goog.inherits to
   * express inheritance relationships between classes.
   *
   * NOTE: This is a replacement for goog.base and for superClass_
   * property defined in childCtor.
   *
   * @param {!Object} me Should always be "this".
   * @param {string} methodName The method name to call. Calling
   *     superclass constructor can be done with the special string
   *     'constructor'.
   * @param {...*} var_args The arguments to pass to superclass
   *     method/constructor.
   * @return {*} The return value of the superclass method/constructor.
   */
  childCtor.base = function(me, methodName, var_args) {
    // Copying using loop to avoid deop due to passing arguments object to
    // function. This is faster in many JS engines as of late 2014.
    var args = new Array(arguments.length - 2);
    for (var i = 2; i < arguments.length; i++) {
      args[i - 2] = arguments[i];
    }
    return parentCtor.prototype[methodName].apply(me, args);
  };
};
```

rixner:
@albertjan what OS and node version are you using?  I have no idea how to tell what this actually requires.  I'm using 64-bit linux and it "works" (modulo the outstanding test failures) on both node 8.10 and 10.15.2.

```t901.py``` does not fail for me...

**Comment by albertjan** [comment]
I'm on Mac with node 11.14.0 it works for me too except for that extra failure. 

That snippet of closure code just explains what that `base` function does and how it's put there.

**Comment by rixner** [comment]
From the closure library we're using:

```
/**
 * Inherit the prototype methods from one constructor into another.
 *
 * Usage:
 * <pre>
 * function ParentClass(a, b) { }
 * ParentClass.prototype.foo = function(a) { }
 *
 * function ChildClass(a, b, c) {
 *   goog.base(this, a, b);
 * }
 * goog.inherits(ChildClass, ParentClass);
 *
 * var child = new ChildClass('a', 'b', 'see');
 * child.foo(); // This works.
 * </pre>
 *
 * In addition, a superclass' implementation of a method can be invoked as
 * follows:
 *
 * <pre>
 * ChildClass.prototype.foo = function(a) {
 *   ChildClass.superClass_.foo.call(this, a);
 *   // Other code here.
 * };
 * </pre>
 *
 * @param {Function} childCtor Child class.
 * @param {Function} parentCtor Parent class.
 */
goog.inherits = function(childCtor, parentCtor) {
  /** @constructor */
  function tempCtor() {};
  tempCtor.prototype = parentCtor.prototype;
  childCtor.superClass_ = parentCtor.prototype;
  childCtor.prototype = new tempCtor();
  /** @override */
  childCtor.prototype.constructor = childCtor;
};
```
Do we even need that base function defined?  I don't see anywhere we are ever calling ```goog.base``` in our classes.  Are we?  I think we call the parent constructor as follows:

````
            // Call up through the chain in case there's a built-in object
            // whose constructor we need to initialise
            if (klass.prototype.tp$base !== undefined) {
                if (klass.prototype.tp$base.sk$klass) {
                    klass.prototype.tp$base.call(this, args, kws);
                } else {
                    // Call super constructor if subclass of a builtin
                    args_copy = args.slice();
                    args_copy.unshift(klass, this);
                    Sk.abstr.superConstructor.apply(undefined, args_copy);
                }
            }
```

We're not really fully creating/using classes the way that the closure-library intends.  We can probably just inline the closure code we want into ```Sk.abstr.setUpInheritance```...

albertjan:
We’re not as far as I know I agree we can just inline that bit.


rixner:
Before we can seriously consider merging this, I think we need someone with a Windows setup to confirm that it also works there.  And it doesn't seem like it's working on Travis...

As for t901, it passes for me when I run ```python skulpt.py test``` (all tests pass).  But, I get different results for that file when I run ```python skulpt.py run test/run/t901.py``` than when I run ```python skulpt.py test```.  The answers are all correct, the dictionaries just print out in different orders and ```most_common``` has slightly different output, as @albertjan noted.  I haven't figured out why running the same file the two different ways yields two different results.

albertjan:
One test case with a different order is not something I feel should stop the merging of this. :smile: we should look into using a stable sorting mechanism in the `Counter`class.

I've also been looking at the `goog.exportSymbol` function which handily adds the exported thing to the global scope. This means that we wouldn't have to do that trick to cat all files into one big one. We could make file that `require`'s them all. This way all exported symbols end up in the global scope and this giving us the required bits of skulpt.

We might how ever not have correctly exported the entire public API. I don't know if this should be part of this PR but it is an interesting avenue to consider I think.

I don't own a windows machine anymore. But I have a VM on my other laptop that I can boot up. I'll see if I have time to have a look at the windows env this week.

rixner:
@albertjan I'll point out that no one else has chimed in in favor of moving in this direction.  It may only be you and I who think this is even remotely a good idea. :smile:

I'll look into requiring everything instead of cating the files together.

bnmnetp:
I have been watching and staying out of the way. I think this is a great step forward. Incremental improvements.

albertjan:
about `t901.py` the output was already forces and significantly different from python2.7 see: 
Screenshot 2019-05-14 at 07 45 31

albertjan:
@rixner I can't believe any one is against moving to node. d8 has been a pain in our backside for such a long time. :smile:

rixner:
@albertjan I don't believe the ordering of dictionary keys when iterating is not specified by the language.  I don't think we've ever strived to match Python in that.   (In fact, in CodeSkulptor, I intentionally scramble them, so students don't rely on some perceived order.)

My concern is less that they are "out of order" and more that they are in a *different* order when I run ```test``` vs. ```run```.  Why is the same program operating differently in those two scenarios?

And while I also believe everyone would be in favor of moving to node, that doesn't mean they are in favor of this hackish way of taking step 1. :smile: (@bnmnetp thanks for speaking up!)

rixner:
@albertjan also, do you know what's going on with Travis?  Does this "work" there?

albertjan:
@rixner I think you have to promote this to a real PR.

albertjan:
@rixner I suspect this is due to some optimisation in node/v8 where it has been sorting skulpt stuff for a long time but the time it runs to t901.py so it has formulated optimised routes and objects it probably chooses a different sorting mechanism because it know more about the object in the test run then it does when you just run that file.

rixner:
@albertjan ```gen/astnodes.js``` doesn't export anything or add anything to the global ```Sk``` object, so "require"ing it doesn't work.  I'm trying to decide what the right solution is.  I feel like all of the nodes should be properties of an "astnode" object that gets exported.  I can then import it into the relevant files.   That moves us closer to a real solution but may trap me in the rathole you started down.  Before I start on that, what do you think?

bnmnetp:
Although I have no code to contribute to this particular PR, I'm really interested in moving forward onto node.  So, my suggestion (and its only a suggestion) is that we:

1. Tag and do a release of our current state.
2. Polish up the PR as it stands.  I don't particularly care if the node purists won't love what we have.  But it seems like a real step forward from where we are.
3. Accept the polished PR
4. Then go diving down rat holes.

bnmnetp:
I like the fact that by getting us onto node we free up many people to be working on improving our 'node world' rather than restricting all the efforts to a single PR.

albertjan:
Ah yeah, I do think that moving those astnodes to an object would be a good idea but that is a lot of work because those are generated from the asdl file. So lets stay with `cat` and this can be a good follow up PR.

rixner:
@albertjan I did it anyway.  It wasn't terrible.  And it gets rid of cat.  I would consider that last commit a WIP.  Please take a look and see what you think.

I haven't checked the impact on the closure compiler, though. However, this may make it such that we can build dist using webpack?  Given that you tried that before, do you think it would work on this?

rixner:
@meredydd can you look at and merge (or comment upon if there are issues) #859.  I would like that merged before this one if at all possible.  This is not ready to merge yet, but I'd like to clear any other pull requests that we are going to before I turn this into a real pull request.  Thanks.

albertjan:
@rixner I can't say much about how successful we'll be with web pack but it's definitely worth a try.

rixner:
@albertjan I was going to first just try to get this to work without it.  I'm stuck on the closure compiler, though.  It doesn't like what I have.  Not sure what the best path forward is.  I feel like this is close, but I don't know if I went one step too far.

All of the testing on the uncompressed code works now.  I could have missed code that is not covered by the tests, but other than that.  So, it's just a matter of getting dist to work.  I can then clean this up and turn it into a real pull request.

rixner:
@bnmnetp I definitely agree we need to do a tag and release before something like this gets merged.  There could easily be latent bugs in what I have done.  We may even want to initially put it on a branch while people test it...

src/ast.js
```diff
@@ -1,4 +1,4 @@
-const astnodes = require('../gen/astnodes.js');
+var astnodes = require('../gen/astnodes.js');
```
albertjan: this require doesn't exist in a browser. I think it would be best if astnodes were exported with `goog.exportSymbol` and loaded into the global scope like the rest. It might be best if we rename them to Sk.astnodes.

rixner: I think that would fix my current problem.  This makes it such that I need a bundler, but the closure library doesn't work with them.  Let me do that.

bnmnetp:
This is slightly out of place here but we can get everyone involved in the discussion anyhow.

Here's what we have for open PR's at the moment:

* #861 - this one
* #859 - waiting for @meredydd to weigh in and then we can likely [accept]
* #857 Feat/es6 modules - an experiment similar to this one that we can [close??]
* #801 log2 - opened Oct 11 2018 -  seems  simple enough but tests are failing and its out of date. [keep]
* #789 Exec for skulpt - seems handy but is stale and needs polishing by @acbart [keep]
* #774 - README changes [merge]
* #768 don't copy the array when making an iterator [keep/merge]
* #743 fix the skulpt.py bug [close]
* #609 support ahead of time compilation -- No idea ??
* #478 proposal for new API - nothing to accept yet but is related to defining an external API that we would support
* #467 Update using.html Very dated, can probably [close]
* #157 [close]

Generally, I think we need to just get some consensus on #609, #768, #859 from @rixner, @albertjan and @meredydd 

I'll be happy to press the button to close those that I've marked for closing.

src/ast.js
rixner: This gets me further, and I think it is the right thing to do for this first step.  But now I have a closure compiler error on a line that I didn't touch.  Investigating, but unsure of what happened.  I committed and pushed anyway.

rixner:
I am ambivalent about #609 (as long as it doesn't impact those who don't use it), I think #768 still needs some resolution about which behavior we want, and I am for #859 (obviously).

I agree with closing the proposed requests.  Except perhaps #857.  We're going to want to revisit that at some point (which I suppose we can just as easily do if it's closed).

src/ast.js
rixner: Never mind I figured it out.

albertjan:
- #861 - this one [aj: merge when ready/approved]
 - #859 - waiting for @meredydd to weigh in and then we can likely [accept] [aj: merge when approved]
 - #857 Feat/es6 modules - an experiment similar to this one that we can [close??] [aj: close] it's a failed project that we can learn from in the future
 - #801 log2 - opened Oct 11 2018 - seems simple enough but tests are failing and its out of date. [keep] [aj: keep]
 - #789 Exec for skulpt - seems handy but is stale and needs polishing by @acbart [keep] [aj: keep]
 - #774 - README changes [merge] [aj: merge]
 - #768 don't copy the array when making an iterator [keep/merge] [aj: needs work]
 - #743 fix the skulpt.py bug [close] [aj: close]
 - #609 support ahead of time compilation -- No idea ?? [aj: merge if it doesn't break anything]
 - #478 proposal for new API - nothing to accept yet but is related to defining an external API that we would support [aj: I'm not in favour of this one]
 - #467 Update using.html Very dated, can probably [close] [aj: close]
 - #157 [close] [aj: close]

albertjan:
there's definitely something up with the time, I'm getting this failure:

```
test/unit/test_time.py
Fail: Expected Jan 01 1970 02:01:01 to equal Jan 01 1970 01:01:01
Tests failed in  TimeTestCase.test_strftime  
Ran 5 tests, passed: 4 failed: 1
```

Where did we get strftime from before?

rixner:
The only remaining serious issue I face is ```strftime```.  If I hack up ```support/time-helpers/strftime-min.js``` to place ```strftime``` in the global scope, then ```run```, ```test```, ```rununits```, and ```dist``` work for me with node.  Hacking it up in that way isn't really viable, though.

That file was used then and now for the minimified dist file.

Now, during testing I am using the ```strftime``` package installed by ```npm```.  I'm not sure where ```strftime``` was coming from before.  Was it a part of ```d8```?

If I (or someone else) can solve this ```strftime``` issue, I can clean this up and mark it ready for review.

rixner:
(I suspect ```strptime``` has similar problems, but I don't know that there are any actual tests of it.)

I believe the only place ```strftime``` and ```strptime``` are used are in ```src/lib/time.js```.  Perhaps a working implementation should just be dumped directly into that file?

albertjan:
I think `support/time-helpers/strftime-min.js` used to get loaded into d8.

albertjan:
There is this hack in node to eval into the global scope so you could do something like this for those `ExtLib` files: 

```
(1, eval)(fs.readSync('support/time-helpers/strftime-min.js', 'utf8));
```

For the usual minified version nothing should have changed. It should still concat those ExtLib files to the output.

rixner:
@albertjan you're right ```getFileList``` added them.

Those files are already in the minified file, so I was trying to not to bring them in again (my hack just changed the way they installed themselves).

When I test it, though, the time module doesn't see them when it gets loaded.  Which brings up a broader issue (for the future) if we actually start packaging the code into modules that external modules will no longer be able to rely on anything being in the global scope other than ```Sk```.

rixner:
With my latest commit (b02c705) things now work for me.  In particular, ```run```, ```test```, ```rununits``` and ```dist```.  Others should test to make sure it works for them too.  I am not seeing the hour off issue that @albertjan is.

I don't believe there are any tests using ```strptime```, and I have not imported an implementation of it when testing uncompressed.  There doesn't seem to be a nice node package like there is for ```strftime```.

The question is what other commands in ```skulpt.py``` do people want to see working before this is a legitimate pull request?  There are a bunch of commands that I never use in there.  If no one else uses them either, I'd be in favor of deleting them rather than updating them.

I should probably at least fix ```repl```.  I think that will be easy.

rixner:
I'm not sure why the repl doesn't work...

rixner:
The issue with the repl is that ```repr``` no longer works.  Somehow calling ```repr``` imports ```src/lib/repr.py``` now.  I need to track that down.

rixner:
I believe ```repl``` works now, too.

rixner:
At this point, I think I'm done, unless someone tells me there's something that doesn't work for them.

albertjan:
Nice work @rixner I'm checking things now. Let's promote this to a real PR.

albertjan:
I still have this failing test t901.py and it now fails in such a way that the test don't fail because we're not exiting with a non-zero exit code. Fixing the test is easy I suggest just updating the output file and forcing it. The other error however is something we need to look into, I think.

```
FAILED: (test/run/t901.py)
-----
import collections

print collections.Counter()
print collections.Counter('gallahad')
c =  collections.Counter({'red': 4, 'blue': 2})
print c
print c['green']
print c
c['green'] = 0
print c

print
print
x = collections.Counter('hello world!')
print x
for i in 'hello universe!':
    x[i] += 1
print x

l = list(x.elements())
l.sort()
print l

print x.most_common(2)
print x.most_common()

a = collections.Counter({1:6, 2:4, 3:3})
print a
a.subtract({1:5, 2:-2, 4:7})
print a
a.subtract([1, 1])
print a
a.subtract(collections.Counter({1:-8, 3:2}))
print a

a.update({1:5, 2:-2, 4:7})
print a
a.update([1, 1])
print a
a.update(collections.Counter({1:-8, 3:2}))
print a


-----
GOT:
-----
Counter()
Counter({'a': 3, 'd': 1, 'g': 1, 'h': 1, 'l': 2})
Counter({'red': 4, 'blue': 2})
0
Counter({'red': 4, 'blue': 2})
Counter({'red': 4, 'blue': 2, 'green': 0})


Counter({'d': 1, 'e': 1, 'h': 1, 'l': 3, 'o': 2, 'r': 1, 'w': 1, ' ': 1, '!': 1})
Counter({'d': 1, 'e': 4, 'h': 2, 'i': 1, 'l': 5, 'n': 1, 'o': 3, 'r': 2, 's': 1, 'u': 1, 'v': 1, 'w': 1, ' ': 2, '!': 2})
[' ', ' ', '!', '!', 'd', 'e', 'e', 'e', 'e', 'h', 'h', 'i', 'l', 'l', 'l', 'l', 'l', 'n', 'o', 'o', 'o', 'r', 'r', 's', 'u', 'v', 'w']
[('l', 5), ('e', 4)]
[('l', 5), ('e', 4), ('o', 3), ('h', 2), ('r', 2), (' ', 2), ('!', 2), ('d', 1), ('i', 1), ('n', 1), ('s', 1), ('u', 1), ('v', 1), ('w', 1)]
Counter({1: 6, 2: 4, 3: 3})
Counter({1: 1, 2: 6, 3: 3, 4: -7})
Counter({1: -1, 2: 6, 3: 3, 4: -7})
Counter({1: 7, 2: 6, 3: 1, 4: -7})
Counter({1: 12, 2: 4, 3: 1, 4: 0})
Counter({1: 14, 2: 4, 3: 1, 4: 0})
Counter({1: 6, 2: 4, 3: 3, 4: 0})

-----
WANTED:
-----
Counter()
Counter({'a': 3, 'd': 1, 'g': 1, 'h': 1, 'l': 2})
Counter({'red': 4, 'blue': 2})
0
Counter({'red': 4, 'blue': 2})
Counter({'red': 4, 'blue': 2, 'green': 0})


Counter({'d': 1, 'e': 1, 'h': 1, 'l': 3, 'o': 2, 'r': 1, 'w': 1, ' ': 1, '!': 1})
Counter({'d': 1, 'e': 4, 'h': 2, 'i': 1, 'l': 5, 'n': 1, 'o': 3, 'r': 2, 's': 1, 'u': 1, 'v': 1, 'w': 1, ' ': 2, '!': 2})
[' ', ' ', '!', '!', 'd', 'e', 'e', 'e', 'e', 'h', 'h', 'i', 'l', 'l', 'l', 'l', 'l', 'n', 'o', 'o', 'o', 'r', 'r', 's', 'u', 'v', 'w']
[('l', 5), ('e', 4)]
[('l', 5), ('e', 4), ('o', 3), ('r', 2), ('!', 2), ('h', 2), (' ', 2), ('n', 1), ('s', 1), ('u', 1), ('v', 1), ('w', 1), ('i', 1), ('d', 1)]
Counter({1: 6, 2: 4, 3: 3})
Counter({1: 1, 2: 6, 3: 3, 4: -7})
Counter({1: -1, 2: 6, 3: 3, 4: -7})
Counter({1: 7, 2: 6, 3: 1, 4: -7})
Counter({1: 12, 2: 4, 3: 1, 4: 0})
Counter({1: 14, 2: 4, 3: 1, 4: 0})
Counter({1: 6, 2: 4, 3: 3, 4: 0})

-----
DIFF:
-----
len got: 909

len wanted: 909

firstdiff at: 563 got: 104 (h', 2), ('r', 2), (' ', 2), ('!', 2), ('d', 1), ('i', 1), ('n', 1), ('s', 1), ('u', 1), ('v', 1), ('w', 1)]
Counter({1: 6, 2: 4, 3: 3})
Counter({1: 1, 2: 6, 3: 3, 4: -7})
Counter({1: -1, 2: 6, 3: 3, 4: -7})
Counter({1: 7, 2: 6, 3: 1, 4: -7})
Counter({1: 12, 2: 4, 3: 1, 4: 0})
Counter({1: 14, 2: 4, 3: 1, 4: 0})
Counter({1: 6, 2: 4, 3: 3, 4: 0})
) expect: 114 (r', 2), ('!', 2), ('h', 2), (' ', 2), ('n', 1), ('s', 1), ('u', 1), ('v', 1), ('w', 1), ('i', 1), ('d', 1)]
Counter({1: 6, 2: 4, 3: 3})
Counter({1: 1, 2: 6, 3: 3, 4: -7})
Counter({1: -1, 2: 6, 3: 3, 4: -7})
Counter({1: 7, 2: 6, 3: 1, 4: -7})
Counter({1: 12, 2: 4, 3: 1, 4: 0})
Counter({1: 14, 2: 4, 3: 1, 4: 0})
Counter({1: 6, 2: 4, 3: 3, 4: 0})
)
-----
JS:
-----
(node:384) UnhandledPromiseRejectionWarning: ReferenceError: js_beautify is not defined
    at compareResult (/Users/albertjan/skulpt-ref/test/test.js:276:37)
(node:384) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 44)
(node:384) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

albertjan:
Awesome @rixner it works. Lets put the new output of t901.py in the .real and .real.force files and I'm happy to see this merged 😄

rixner:
@albertjan I fixed the ```UnhandledPromiseRejectionWarning```.  However, the ```t901``` failure is problematic.  It works for me.  So, if I modify the expected output to work for you, then it will fail for me.  And we don't know what other people will get.  Presumably it is because you are using a different version of node than I am?

I think there are two options:

1. Add a ```t901.py.real.alt``` file with the alternate output.  The only other test that has that is ```t502```.  If there are only two variations that we see, then I guess that will work?
2. Convert this to a unit test and do a much more sophisticated check for correctness.

I lean towards (1) because it is easy.  I'm going to do that speculatively based upon your comment above, but can back out of it if other disagree.  We should all confirm, though, that the output you are getting is actually correct!

rixner:
Has anyone running windows tried this yet to confirm that I didn't break windows?  I am running Linux and @albertjan is on a Mac and it works for us across 3 different node versions (8.10, 10.15.2, and 11.14).

rixner:
All, I updated this to a real pull request so that we can see if it works on Travis.  I still don't think it should be merged yet, though.  I think we should do the housekeeping that @bnmnetp is working on to clear all pending pull requests first.

bnmnetp:
I have cleaned up the PRs that we were going to close.

Not sure if any of the remaining open PRs really need to be merged before tagging and then accepting this one...


In my opinion, I think we ought to tag what we have as 1.0.0 and move on from there.  We've been using skulpt in production for a long time and I don't think that 0.12.0 really represents that very well.  We don't even have published criteria for what it 1.0 would be.

albertjan:
@vasiljevic are you still around? You fixed the windows devenv last time can you have a look if this works for you?

rixner:
@bnmnetp I was originally hoping #859 would go before this pull request, but I guess it ultimately won't matter that much...

As for "1.0.0", I'd really like to see #859 in before we leave the "0.X" series, so that the new internals are codified into the first stable release.  But, I agree, it's about time that we move to "1.X".

rixner:
@albertjan if no one is around to test this on Windows, perhaps we hope for the best and if it breaks, someone will speak up?  I hate to do that, but if none of us uses Windows, I don't know what else to do...

acbart:
Hi, I have a windows machine. The only problem is that it's the last week of classes here at UD and I'm totally swamped. I'll try to grab a little time to give it a go. Can anyone give me the specific command to run and what you want me to report back?

albertjan:
@acbart `./skulpt.py dist`

acbart:
[skulpt_windows.log](https://github.com/skulpt/skulpt/files/3187528/skulpt_windows.log)

Okay, I downloaded a fresh repo, checked out the pull request.

    > npm install

Had to edit package.json to point directly to Python2.7 since it's not on my path.

    > npm run build

Looks like all the tests passed, although it did produce an error afterwards when it tried to do something with the bundled file (looks like a bad path).

```
module.js:471
    throw err;
    ^

Error: Cannot find module '../../distskulpt.min.js'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (C:\Users\acbart\Projects\temp\skulpt-v1-windows\support\tmp\dist.js:2:1)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! skulpt@1.0.0 build: `C:/Python27/python.exe skulpt.py dist`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the skulpt@1.0.0 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\acbart\AppData\Roaming\npm-cache\_logs\2019-05-16T14_29_56_534Z-debug.log
```

But from the attached logs, seems like it's fine otherwise.

albertjan:
Hey @acbart that's brilliant can you send me the contents of `support/tmp/dist.js`.

rixner:
@acbart, thanks!  It seems like ```os.path.join``` is not doing the right thing, maybe?

acbart:
Yeah, looks like it ended up with a back slash instead of a forward slash. Always a pain on windows :)

```

require("../../dist\skulpt.min.js");
require("../../test/test.js");
```

rixner:
I see.  I need to fix something things. Thanks, @acbart!

albertjan:
`os.path.join` concats paths according to the os so it's "correct" to use `\` but node helpfully normalises paths to use `/`. so for those require paths we can just use `/`

rixner:
@albertjan do you know what ```shutil.copy``` is going to want?  We need to either always use ```os.path.join``` or always use ```/```, the problem here is that I mixed and matched.  I'm inclined to fix it so that we always use ```os.path.join```.  Although there are other places in ```skulpt.py``` that use ```/``` that I don't think I touched.

rixner:
(At least I think the problem is that I used both in the same path...)

albertjan:
@rixner you're right it shuts.copy is not going to like the wrong slash on windows. I think you'll have to do something special here: https://github.com/rixner/skulpt/blob/node/skulpt.py#L774

vasiljevic:
@albertjan I've got the same error as @acbart . Sorry I'm late :)

albertjan:
@vasiljevic no worries thanks for checking though 😄

rixner:
@acbart, @acvasiljevic can one of you try again?  I hope I fixed it...  @albertjan you may as well try again too, to make sure I didn't break anything for Macs.

vasiljevic:
@rixner, looks like build works now:
[build-output.txt](https://github.com/skulpt/skulpt/files/3189071/build-output.txt)

rixner:
@vasiljevic great, thanks!

albertjan:
works for Mac too 😄

albertjan:
@skulpt/committers I think we should merge this. It takes us a significant step closer to a modern javascript development setup.

rixner:
@albertjan updated things to use a current version of the closure compiler from npm, instead of carrying around an ancient version inside the repository.

bnmnetp:
@albertjan , @rixner 

As I understand it there is some desire/plan to get #859 merged, and then tag before we merge this one.  I'm happy to do that right now as I've pulled #859 onto my branch, and tried out a bunch of libraries and it appears to work just fine.  I know @meredydd had expressed interest in reviewing this one but now that I've done my testing I'm not sure we need to wait for that.  But we could give him the weekend and if he looks at it so much the better.  I can merge this after I get his approval or Monday morning whichever comes first.

What do you think of that plan?

rixner:
@bnmnetp I'm fine with your proposed plan.  Or you could even merge both now.  @meredydd can always submit another pull request to modify/update things from #859 if he wants.  I doubt he's going to be against the whole thing, as he clearly was supportive of the idea.

Regardless, anyone can keep making pull requests against the ```node``` branch in rixner/skulpt if they are eager to work on node things before this gets merged.

My main request is that #859 be merged before we move to revision "1.0".  As long as that happens, we could merge things in either order and I'll be fine with it.  I just have a preference for merging #859 first in case the switch to node has any unforeseen side effects (which we will obviously fix as they are found).

bnmnetp:
Oooh, And now that I merged #859 we have conflicts...

rixner:
@bnmnetp thank you for merging #859 first!  I'll resolve these conflicts this evening.  They are somewhat trivial, but I want to be careful.

rixner:
The merge required a few minor changes to make the new closure compiler happy.  @bnmnetp you might want to try it one more time before merging, but it works for me and it passed Travis, so that just may be me being extra cautious...

bnmnetp:
Yep, I built a couple of books with this version and everything still works as expected.

Thanks @rixner!  Thanks @albertjan !

PR #865, opened 2019-05-15
support turtle color mode 0.00~1.00

alexwenbj:
Add supporting for turtle color float value 0.00 - 1.00

### DIFF
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -2036,10 +2036,18 @@ function generateTurtleModule(_target) {
         }
 
         if (color.constructor === Array && color.length) {
+            // Check float number
+            var float_reg = /^\d+(\.\d+)?$/;
             for(i = 0; i < 3; i++) {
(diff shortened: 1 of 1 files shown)


bnmnetp:
Thanks for this.  

For the sake of completeness, how about adding the [colormode](https://docs.python.org/2.7/library/turtle.html#turtle.colormode) method as well?

alexwenbj:
> Thanks for this.
> 
> For the sake of completeness, how about adding the [colormode](https://docs.python.org/2.7/library/turtle.html#turtle.colormode) method as well?

OK,I will have a try.
But,I checked the python 2.x and 3.x document,I found that the fillcolor/color/pencolor functions have not the 4th argument "alpha",it is not the same as in the file turtle.js fillcolor/color/pencolor definition.Have i mistaken something?
@bnmnetp

alexwenbj:
Adding the colormode function,this PR is not useful anymore.

PR #866, opened 2019-05-15
Add turtle colormode function

alexwenbj:
Add colormode function to turtle.

### DIFF
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -2036,10 +2036,18 @@ function generateTurtleModule(_target) {
         }
 
         if (color.constructor === Array && color.length) {
+            // Check float number
+            var float_reg = /^\d+(\.\d+)?$/;
             for(i = 0; i < 3; i++) {
-                color[i] = (typeof color[i] === "number") ?
-                    Math.max(0, Math.min(255, parseInt(color[i]))) :
-                    0;
+                if(typeof color[i] === "number") {
+                    if(color[i] <= 1 && float_reg.test(color[i])){// If 0.00-1.00
+                        color[i] = Math.max(0, Math.min(255, parseInt(255 * color[i])));
+                    }else{
+                        color[i] = Math.max(0, Math.min(255, parseInt(color[i])));
+                    }
+                }else {
+                    color[i] = 0;
+                }
             }
             if (typeof color[i] === "number") {
                 color[3] = Math.max(0, Math.min(1, color[i]));
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -864,7 +864,7 @@ function generateTurtleModule(_target) {
 
         proto.$pencolor = function(r,g,b,a) {
             if (r !== undefined) {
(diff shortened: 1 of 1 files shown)


src/lib/turtle.js
```diff
@@ -1053,6 +1055,21 @@ function generateTurtleModule(_target) {
...
+                } else {
+                    this._colorMode = 1.0;
+                }   
+                return this.addUpdate(undefined, this._shown, {colorMode : this._colorMode});         
+            }
+
+            return this._colorMode===255?Sk.builtin.int_(255):Sk.builtin.float_(1.0);
+        }
+        proto.$colormode.minArgs     = 0;
+        proto.$colormode.co_varnames = ["mode"];
```
bzwheeler: It is preferred to use the `returnType` decorator to supply information on how values should be cast to Sk types as opposed to returning Sk types directly.

```js
proto.$colormode.returnType = function(value) {
    return value === 255 ? Sk.builtin.int_(255) : Sk.builtin.float_(1.0);
};
```
This keeps the turtle implementation as "pure" javascript so we don't have to do remember to do any type juggling if we call other methods from within this file.

src/lib/turtle.js
```diff
@@ -2028,18 +2046,35 @@ function generateTurtleModule(_target) {
         return result;
     }
 
-    function createColor(color, g, b, a) {
+    function createColor(turtle,color, g, b, a) {
```
bzwheeler: All we actually need from the turtle object here is the color mode setting so it would be better to just accept that value rather than accepting a reference to the entire turtle object.

src/lib/turtle.js
```diff
@@ -2028,18 +2046,35 @@ function generateTurtleModule(_target) {
...
+                        color[i] = Math.max(0, Math.min(255, parseInt(color[i])));
+                    } else { 
+                        throw new Sk.builtin.ValueError("bad color sequence");
+                    }
+                }
+            } else {//In python,if the colormode not equals 255,it should be 1.0
+                for(i = 0; i < 3; i++) {
+                    if(typeof color[i] === "number") {
+                        if(color[i] <= 1){
+                            color[i] = Math.max(0, Math.min(255, parseInt(255 * color[i])));
```
bzwheeler: I'd consider removing the `if/else` block and just replacing `color[i] = Math.max(0, Math.min(255, parseInt(255 * color[i])));` with `color[i] = Math.max(0, Math.min(255, parseInt(colorMode * color[i])));`.

alexwenbj: Yeah!Completed your two suggests directly in this branch.

alexwenbj: Because here should raise an error,so i didn't change it as you suggested.

alexwenbj: Didn't change for your suggestion about replace if/else.

albertjan:
:tada:

bnmnetp:
Thanks @alexwenbj !

alexwenbj:
@bnmnetp My pleasure.

Issue #867, opened 2019-05-16
Thoughts on the evolution of the code base


rixner:
While working on #861, I came to two conclusions:

1. The closure library is probably not pulling it's weight.
2. I think ```skulpt.py``` should be rewritten in Javascript.

First, at this point, from the closure library, we are using ```goog.exportSymbol```, ```goog.asserts.assert```, ```goog.asserts.fail```, and a few various ```goog.string``` functions.  Presumably, we could write our own versions of these few functions (or use lighter weight packages that do the same thing).  Instead, we're adding some 4000 lines of Javascript into our distribution from a very old version of the closure library.  While the closure compiler still probably has some value, I really don't think the closure library does for us.  Furthermore, the closure library does not appear to be modularized in a modern way, so if we start rewriting the core of Skulpt, it will take some effort to pull in just the pieces of the closure library we want.

Second, I think it will be easier overall to use node to manage and run things if the control code is in Javascript instead of Python.  This is likely something we could do incrementally, moving pieces of the infrastructure to ```skulpt.js``` until ```skulpt.py``` is obsolete.

Just my thoughts.  Figured I'd record them here for posterity.

rixner:
I have successfully removed all use of the closure library from the code.  I will make a pull request after #861 is merged.

rixner:
I now also am able to build Skulpt with webpack.  And I can run the closure compiler via webpack.  I also have it run jshint during the build.  I cannot get jscs to work correctly with the current version of webpack, but I think that's largely because jscs has been subsumed by eslint.  Eslint does run, but with the automatically converted rc file I made it throws tons of errors.  I think it would be good for us to convert to using eslint, but we need to agree on an rc file (preferably one that enforces our existing style without adding new restrictions, in my opinion).

In converting to webpack, it also obsoletes a lot of things in skulpt.py, so that's getting smaller and smaller.  I do have a few questions for people:

1. What's with the STANDARD_NAMING if statements?  Do we really need to carry the non-standard names, or can I just drop them entirely?
2. I favor always building either a compressed or uncompressed skulpt.min.js/skulpt.js file before running or testing anything.  That ensures we are testing an actual build with everything defined properly, rather than just importing the entry point and hoping that's the same.  Are people okay with that?
[...]

Issue #868, opened 2019-05-20
turtle drawing will block in strange situation


alexwenbj:
Please visit the link [here](https://trinket.io/library/trinkets/a29bcba579) ,It will block when the i is 126.And the drawing is stop,but the turtle head is still turn around slowly.
(Mac OS + Chrome 74.0.3729.157)
It is OK when run in Safari.

bnmnetp:
I think you are just hitting an execution time limit.  Instead of speed(0) use tracer(32,0) to speed it up.

alexwenbj:
> I think you are just hitting an execution time limit. Instead of speed(0) use tracer(32,0) to speed it up.

But it is OK in Python.Moreover,It is OK in Safari.

alexwenbj:
If I add 3 lines code:
```
from turtle import *
speed(0)
colors = ["red","yellow","blue","black"] #Add line 1
for i in range(360):
    for j in range(4):
        color(colors[j%4])  #Add line 2
        forward(100)
        left(90)
        print("j=",j)  #Add line 3
    left(1)
    # print(i)
```
It is also OK.

bnmnetp:
Python has no execution limit.  I'm not sure about trinket, but on my site I add an execution limit to keep beginners from locking up their page when they make an infinite loop.

But, I'm not sure what to make of your last example or why that would change the behavior in any significant way.

alexwenbj:
I think it is not because of execution time limit.If you look at the drawing,you can find the forward is stop but the turn left is still executing,so I guess it is because the turn left and forward action is not synchronously after execute some times.The print() and change color function, slow down the whole actions, then the forward and turn left is synchronously.

alexwenbj:
By the way,I looked inside the print.js code and found print function is supported the end argument,but when I run the code below:
```
for i in range(10):
  print(i,end=",")
```
It failed,do you know why?thank you. @bnmnetp

bnmnetp:
I'm not sure what the default mode is for trinket.  But you need to be in python3 compatible mode for end to work.  Your example works fine for me.

It does seem like there is something funny going on with the "anonymous" turtle.  If I change your original example to explicitly use a turtle object, it works fine:

```
import turtle
t = turtle.Turtle()
t.speed(0)
for i in range(360):
    for j in range(4):
      t.forward(100)
      t.left(90)
    t.left(1)
    print(i)
```

This would not be the first time that the "anonymous" turtle had a bug in it.

[...]

PR #869, opened 2019-05-20
Remove closure library

rixner:
As I wrote in #867, I don't think the using the closure library is worth it.  We use very few functions from it and it is not designed in a way that is friendly to modern Javascript modules.

This pull request includes replacements to the functions we actually do use (found in ```src/assert.js``` and ```src/util.js```).

The one noticeable difference is that an assertion failure now throws and ```Error``` instead of an ```AssertionError```.  It doesn't seem like the machinery to subclass ```Error``` in an ES5 compatible way that preserves the stack trace is worth it.  Especially since assertions are turned off in the distribution anyway, so this is only a development-time error.


rixner:
FYI, I just tried a quick experiment, and I think this is going to make it possible to use webpack to build the dist file instead of having to have a separate file list in ```skulpt.py```.  I used ```main.js``` as an entry point, and it just built the file correctly and it runs (I didn't minify it).  I didn't run the closure compiler on it, but I think that is possible through webpack (maybe?), so we can hopefully create the distribution that way.  (The impediment before was the non-node-friendliness of the closure library.)

rixner:
@skulpt/committers does anyone have time to take a look at this?  Thanks!

bnmnetp:
I've built and tested on all of my tricky examples...With the exception of finding a case in image processing that exercises #875 / #868 in Chrome everything works fine.

I have no issues with removing the closure library in general either.  So unless we get an objection from someone else I can merge.

rixner:
@s-cork you are correct, this is wrong.  Assertion failures just throw ```Error``` now.  Apparently, that's not tested, so it slipped through...

PR #873, opened 2019-05-22
Fix/decode escape

### LINKED ISSUE #871: Some character make result wrong in format.
When I use print to output some Chinese. I already set  `__future__=Sk.python3`
e.g
```
print("这是{}".format("中文"))
>>> 这是中文
```
But when I append some character, for example:
`\n`
```
print("这是\n{}".format("中文"))
>>>è¿æ¯
>>>中文
```
This caused an coding error.

albertjan:
A unicode string with an escape in it didn't get correctly decoded, this addresses this.

fixes #871

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -1726,7 +1726,7 @@ function parsestr (c, s) {
                 ret += c;
             }
         }
-        return ret;
+        return decodeUtf8(ret);
     };
 
     // console.log("parsestr", s);
--- a/test/unit/test_unicode.py
+++ b/test/unit/test_unicode.py
@@ -1,0 +1,15 @@
+import unittest
+
+class TestUnicode(unittest.TestCase):
+    def test_normal_string_should_be_parsed_correctly(self):
+        unicode_string = u"这是"
+        self.assertIn(u"是", unicode_string)
+
+    def test_string_with_escape_should_be_parsed_correctly(self):
+        unicode_string = u"这是\n这这这"
+        self.assertIn(u"是", unicode_string)
+
+
+
+if __name__ == '__main__':
+    unittest.main()


albertjan:
@Dustyposa please have a look and see if this fixes it.

Dustyposa:
@albertjan  Ok,I will have a look. The dist directory is newest file? Do I need to compile it, and get the newest dist files?

albertjan:
You'll need to compile my branch but I can give you the files.

albertjan:
ugh test failures... 😄

Dustyposa:
emm, I think the problem is [formatting.js](https://github.com/skulpt/skulpt/blob/master/src/formatting.js)

eg1:
```python
print("我\n你")
>>> 我
>>> 你
```
The result is right.
eg2:
```python
print("我\n你{}".format("我"))
>>> æ
>>> ä½ 我
```
The result is wrong.
Expect:
```
我
你我
```
eg3:
```python
print("我\n你{}".format("wwwwww"))
>>> 我
>>> 你wwwwww
```
This result is right.

I doubt when Sk handle format function, cause this problem.
And I guess the formatting encoding is determined by args make this problem.

albertjan:
hmmm ok we're seeing different behaviour now 😄

Dustyposa:
I want to change the encode way. But I found decodeUTF8 function in ast.js , use it is't convenient.
😊

Lifan1998:
@Dustyposa @albertjan I also encountered this problem.

eg1:
```
print("我\n你")
>>> æ
>>> ä½ 
```
eg2:
```
print("我\n你{}".format("我"))
>>> æ
>>> ä½ 我
```
eg3:
```
print("我\n你{}".format("wwwwww"))
>>> æ
>>> ä½ wwwwww
```
eg4:
```
print("我\n你{}".format("我\n"))
>>> æ
>>> ä½ æ
```
I modify ast.js:
```
1728  -    return ret
1728  +   return decodeUtf8(ret);
```
This result is All right.

albertjan:
@Lifan1998 really?? do all the tests still pass if you do that?

Lifan1998:
@albertjan  No. There are two test methods that didn't pass, but it works fine and I plan to release it.

Dustyposa:
Oh, has this function `decodeUT8`? 
I didn't find this function before. 
Now still need py27 to compile Sk? or py3?

meredydd:
I think this is superseded by PR #983 (the test case passes on that branch).

@albertjan, are you comfortable that the test cases in #983 cover this bug and will prevent regressions, or do you want to import some of this test code?

PR #886, opened 2019-05-25
add resizemode,shapesize function to turtle

alexwenbj:
And change the colormode var name to cmode, identical to Python.


alexwenbj:
After reading the python turtle.py source code,I found my implementation is not very rigorously identical to the native python,So I close this PR temporarily. Maybe I will re-open it after when it is fully qualified.Sorry for that.

albertjan:
That's perfectly fine @alexwenbj we appreciate your contributions 😄

Issue #875, opened 2019-05-22
Long loop breaks suspensions


cgibin:
Hello everyone.
I am facing weird problem using suspensions. I have this problem with python->js code generator, but reduced it to make it more clear.

This code represents suspensions mehanic in a nutshell.

```javascript
(function() {
  var _module = function() {
    const _scope = function() {
      var a = 0,
        block = 0;
      var suspend = function() {
        var context = {};
        context.a = a;
        context.block = block;
        context.resume = function() {
          _scope.context = context;
          return _scope();
        };
        _scope.context = context;
        return context;
      };
      var awake = function() {
        context = _scope.context;
        delete _scope.context;
        //_scope.context = undefined;
        a = context.a;
        block = context.block;
      };
      if (_scope.context !== undefined) {
        awake();
      }
      while (true) {
        switch (block) {
          case 0:
            block = 1;
            return suspend();
          case 1:
            block = 0;
            if (a++ > 10000) {
              return;
            }
            continue;
        }
      }
    };
    return _scope();
  };

  var runMod = function() {
    var susp = _module();
    var nSteps = 0;
    while (susp) {
      susp = susp.resume();
      if (!(++nSteps % 1000)) {
        console.log(nSteps, susp);
      }
      if (nSteps > 100000) {
        console.log("aborted");
        break;
      }
    }
    console.log("complete", nSteps);
  };

  runMod();
})();
```
This code should iterate 10000 times and return
Instead, it runs infinitely in Chrome(I added sane check, and it stops after abnormal amount of steps), but works fine in Firefox
If I reduce iterations to 100 it works in both browsers(both up to date)
When it hangs, variable "a" stuck at same value without increment(about 1000 on my cpu).
I found that problem in this line of code "delete _scope.context", replacing it with "_scope.context = undefined" fixes this bug
The problem exists in code generated by skulpt, if you need I can provide generated code

The code is here https://jsfiddle.net/daborkpw/

Can you confirm this problem and restore faith in my sanity?

bnmnetp:
@meredydd @albertjan @rixner I have a feeling this is related to #868 which I can still reproduce in the very latest builds but forgot to reopen.  It also is reproducible in older builds so I don't think it has anything to do with the move to node.

cgibin:
Yes, #868 also related to this, I somehow missed that. I also originaly started to investigate this bug when one of our kids tried to draw circle by 1 degree steps. And it drew only partialy

albertjan:
[...]

PR #888, opened 2019-05-26
Workaround chrome<75 optimizer delete bug

cgibin:
Addresses issues #875 and #868
Already fixed in Chrome > 74


albertjan:
Thanks @cgibin!

Issue #870, opened 2019-05-22
wrong output with class variable name


ANDEMEN:
```python
class Dog:
  name = 'xxx'
  nickname = 'xxx'
  
print(Dog().name) # output: d
print(Dog().nickname) # output: xxx
```

why the output of the variable in class called name is always d ?

albertjan:
You're right that's a bug.

```
➜  skulpt-ref git:(fix/repl-whitespace) ✗ ./m repl
Python 2.6(ish) (skulpt, Wed May 22 2019 09:04:06 GMT+0100 (British Summer Time))
[node: v11.14.0] on a system
Don't type "help", "copyright", "credits" or "license" unless you've assigned something to them
>>> class Test():
...     name = 'xxx'   
... 
>>> Test.name
klass
>>> 
>>> Test().name
klass
>>> class Test2:
...     name = 'xxx'
... 
>>> Test2.name
klass
>>> Test2().name    
klass
>>>
```

How has no-one ever tried to use a name field on a class. 😄 

This has to do with the fact that a function in javascript has a name property. And apparently we prioritise looking up stuff on the javascript class above stuff that lives in the class dict.  I'm unsure if we should change this priority or if we should make a special case for `name`.

PR #889, opened 2019-05-27
Use webpack to build Skulpt

rixner:
As discussed in #867, this pull request changes the build process to use Webpack.

I think this is an important improvement as it eliminates the duplicate file list used to build skulpt (in skulpt.py and main.js).  ```main.js``` is now the definitive entry point and is used by Webpack to build the final distribution file.  This also means that others can start making incremental progress on improving the code base.  We can create better modules that do not put everything in the global ```Sk``` namespace, for example and instead just get imported where they are needed.  We can therefore move incrementally towards what @albertjan tried to do in one step with #857.

This does mean that one needs to use ```npm``` instead of ```skulpt.py``` to run and build Skulpt now.

I made this a "draft" pull request because I don't think it is going to work properly with Travis yet.  I have no idea how that is set up, so I don't know how to fix it.  Presumably, we would want to run ```npm run dist``` and make sure that passes.  The other issue with that, though, is that it currently does not pass.  Some of the Python 3 unit tests currently fail in master.  This is unrelated to ```node``` or ```webpack``` and should otherwise be fixed.  We previously did not run the Python 3 tests on accepting pull requests.  I think we should, and have therefore added them to the dist build.  There is also a lightweight ```devbuild``` that does not run the tests, the linters, or the closure compiler, which one would use during development, and it can be rebuild automatically when you edit source files using ```watch```.

There probably should be a developer readme to explain how to use the new commands.  I will add that, if people think this is the right direction in which to go.


albertjan:
I just ran the first few things through this I like it it works well. I love that the (dev)build, it's really nice to have a command that builds something I can drop into a project with skulpt, and have the stdlib there too without having to run through and wait for the tests to run. 😄 

I do have a few test fails but you have them too I assume.

bnmnetp:
Cool! 💯 

I just did a build and it worked great.  I also did testopt and I don't have any failures... `Passed: 432 Failed: 0`

bnmnetp:
We should definitely get a bit more developer doc done and get it ready to merge

rixner:
@albertjan what errors are you seeing?  I see two failures in Python 3: one each in test_class.py and test_defaultdict.py.  As I said, these predate the migration to node.  I believe my student has a fix for both, so we should be making a pull request in a few days for those.

I want to consolidate some commands to make things a little bit simpler (and use flags to control python version, debug, etc.), then I'll write some documentation (and ```npm run help```) and convert to a real pull request.

rixner:
Travis tried to do npm run build-min, which I don't have.  I could add it, or we could change Travis to run build or devbuild.

bnmnetp:
@rixner - the travis stuff is in `.travis.yml` line 13 you can change the build command.

albertjan:
@bnmnetp @rixner and in diet-update.sh, which also contains a bunch of other stuff around updating the website and skulpt-dist.

rixner:
Thanks @bnmnetp and @albertjan !  I think I updated them both correctly.  It now fails, as expected, on the Python 3 tests.  However, it fails 3 tests, whereas I only see two failing tests.  It fails on test_set.py which I don't see locally.

rixner:
I can change it to ```npm run dist```, that makes sense.  Should it be ```build``` or ```dist``` in ```dist-update.sh```, though?

albertjan:
@rixner yeah because we've already proven the tests pass 😄

.travis.yml
```diff
@@ -1,7 +1,7 @@
-language: python
+language: node_js
 sudo: false
-python:
-- '2.7'
+node_js:
+- '10'
 install:
 - nvm install 'lts/*'
```
albertjan: these 3 lines can go now 😄 nvm is a version manager for node. and `npm install npm@latest -g` was because of a  bug in npm years ago.

rixner: To be clear, you mean these 3:

- nvm install 'lts/*'
- nvm use 'lts/*'
- npm install npm@latest -g

Also, they are redundant anyway, I think, since I switched the language to node_js and told it which version to use...

albertjan: yeah exactly 😄

albertjan:
```
UNCAUGHT EXCEPTION: ExternalError: TypeError: Cannot assign to read only property 'name' of function 'function(a,b){void 0!==e.prototype.tp$base&&(e.prototype.tp$base.sk$klass?e.prototype.tp$base.call(this,a,b):(a...<omitted>...)}' on line 1187
```

I'm getting this error too we should add `name` to the reserved words. that also fix #870

bnmnetp:
Just FYI, since you are all chatting on this PR.  I got control of skulpt.org this morning, and rather than go through the hassle of setting it up to point to google app engine, and then change it to point to github.  I just set it up to point at github.  So anything in dist-update.sh script that has to do with deploying to GAE can go.

I'll make a separate PR that removes the code for GAE, and I can also handle updating dist-update.sh then as well.

albertjan:
@bnmnetp Nice one 😄

rixner:
Why has this just started happening now, and why does it not happen everywhere?  As I said, this doesn't happen for me locally...  It doesn't happen when I run ```npm run btest``` either.  Regardless, should that be a separate pull request, or should I do it here?

albertjan:
> Why has this just started happening now, and why does it not happen everywhere? As I said, this doesn't happen for me locally... It doesn't happen when I run `npm run btest` either. Regardless, should that be a separate pull request, or should I do it here?

I don't know it doesn't seem to be entirely true either see: 
```
➜  skulpt-ref git:(rixner-webpack) ✗ node   
> var x = function test() {}
undefined
> x.name
'test'
> x.name = 'test2'
'test2'
> x.name
'test'
> 
```

albertjan:
I think it only breaks in the minified version 

```
npm run dist
npm run run3 test/unit3/test_set.py <-- breaks
npm run brun3 test/unit3/test_set.py <-- works
npm run run3 test/unit3/test_set.py <-- works
```

rixner:
@albertjan I don't think adding "name" to the reserved names worked.  The failure is now "worse" on Travis:

test/unit3/test_set.py
Uncaught Error in:  TestSet
Uncaught Error in:  TestSetSubclass
Uncaught Error in:  TestSetSubclassWithKeywordArgs
Uncaught Error in:  TestBasicOpsEmpty
Uncaught Error in:  TestBasicOpsSingleton
Uncaught Error in:  TestBasicOpsTuple
Uncaught Error in:  TestBasicOpsTriple
Uncaught Error in:  TestBasicOpsString

And even worse, my infrastructure doesn't seem to count that as a failure.  I don't know if that's a flaw in something I've done or something in our unit test module.  With the "fix", I'm not seeing the error from before locally:

test/unit3/test_set.py

UNCAUGHT EXCEPTION: ExternalError: TypeError: Cannot assign to read only property 'name' of function 'function(a,b){void 0!==e.prototype.tp$base&&(e.prototype.tp$base.sk$klass?e.prototype.tp$base.call(this,a,b):(a...<omitted>...)}' on line 1187

albertjan:
Hmmm very odd sorry. I'll try and investigate some more.

rixner:
@albertjan I didn't even test it as it seemed so obviously correct, so it's on me, too!  Regardless, I think that was necessary anyway.

albertjan:
Right well this was a bit of a rabbit-hole. First this `"Uncaught Error in: black"` message is from `unittest/__init__.py`, this should be changed to raise the exception I think, it means it fails in the unittest code.

And then there's a problem with multiple inheritance in the python3 mode I've reduced the example to this:

```
class A:
    def __init__(self):
        self.a = []

    def run(self):
        for name in dir(self):
            # self doesn't have an a variable here. 
            # this could be because we didn't call __init__
            # I'm still investigating.
            self.a.append(getattr(self,name))

        print(self.a)
 

class B:
    def feep(self):
        self.c = "test"


class C(B,A):
    def foop(self):
        self.a.append(self.c)


c = C()

c.run()
```

albertjan:
Yeah it doesn't call `__init__` on the second object you inherit from.

rixner:
I believe #870 is now fixed in this branch.

rixner:
@albertjan when you get a chance, can you look at the ```.eslintrc.json``` file and see if you believe it has correctly captured all of our style rules?  If so, I can drop jshint.  I believe you were the one who initially added the style rules, right?  I suppose that does not necessarily need to happen as a part of this pull request, though.  Keeping jshint is only slowing down the build, not hurting anything else.

rixner:
@albertjan, in regards to the multiple inheritance failure:

1. I don't know whether the unittest code should print that message or throw an exception, but I now parse that message and convert it into a test failure, so the tests will no longer pass if that happens again.

2. Did this ever work, or was that test always failing and none of us ever noticed?  I guess it doesn't matter, but I'm wondering if I need to try to put more effort into "battle hardening" the testing framework so that errors actually always cause the tests to fail.

rixner:
FYI: CONTRIBUTING.md has been changed to document the available ```npm``` commands, and ```npm run help``` also prints them all out with the same description of what they do.

rixner:
With regards to "battle hardening" the unit tests, I would like to propose that we write our unit tests so as to never print anything out.  Then, if the test runner detects any output other than something like "Ran 12 tests, passed: 12 failed: 0", it can declare it a failure.  That would make it so that we don't have to predict all possible internal failure cases.  Does that make sense to people or is that overly draconian?

albertjan:
>  I would like to propose that we write our unit tests so as to never print anything out. 

That seems like a very brittle way, we should be able to rely on test failures in the code. This uncaught exception thing should not have happened. I think we should not be catching these errors. I don't mind if none of the tests pass if there's an error in the unittest code, or some mayor error in skulpt.

albertjan:
About the inheritance problem, when you're in python 3 mode `object` is automatically added to the bases for classes. And then in type look up it walks the `mro` object if finds `__init__` in the base of the object and calls that instead of calling the `__init__` of the object. 

What does python3 do in this situation?

rixner:
@albertjan, the unittest code did catch the exception.  It then printed that error...  You want to reraise the exception?  I don't know what the Python unittest code does in this situation.  I have never seen the real Python unittest cause an exception itself.

We are already using text processing to count the number of tests that pass and fail, which is equally brittle...

albertjan:
> We are already using text processing to count the number of tests that pass and fail, which is equally brittle... 

I agree, we should make a test runner class that runs all the tests in a package. Like python does. :smile:. But that should be part of another project. I think that if there's a bug in the unittest package in python, it'd just throw an exception, and not swallow it.

rixner:
@albertjan, I'm good with leaving improving the unittesting framework as a project for another day. :)   It really has nothing to do with node/webpack/etc.

At this point, then, I think there are two options:

1. Fix the failing tests so that this pull request can pass.
2. Comment out the failing tests for this pull request and then fix them in the future.

Personally, I favor option 1, simply because they need to get fixed sometime, so why not now.  I believe we should have a fix the other two shortly.  I don't know if you have a handle on the fix for the multiple inheritance problem or just a diagnosis...

bnmnetp:
If only Brad Miller could remember why he wrote that code the way he did all those four years ago... 😄

rixner:
@bnmnetp don't use git blame too much, I suspect we'll all find things we don't remember, don't understand, and/or aren't proud of. :smile:

rixner:
At this point, I think the only outstanding issue with this pull request is the multiple inheritance issue causing the failures in test/unit3/test_set.py.  Once we decide what to do with that, this is ready to be reviewed and merged.

rixner:
Failing tests have been temporarily disabled.  I believe fixing them should still be a high priority, even if they are being decoupled from this pull request.

@skulpt/committers please review!

support/run/run_template.ejs
```diff
@@ -97,9 +108,13 @@ <h3>Test Code</h3>
 <div id="runbrowser">
 <form>
 <textarea edit_id="eta_5" id="runbrowser_code" cols="72" rows="20" style="font-family: 'Lucida Console', Monaco, monospace;">
-%(code)s
+<%- code %>
 </textarea>
-<button onclick="runit('runbrowser')" type="button">Run</button>
+<br>
+<input type="checkbox" id="debug">Debug</input>
```
albertjan: I love this 😄

support/time-helpers/strptime.js
```diff
@@ -0,0 +1,483 @@
+/* jshint -W067 */
```
albertjan: couldn't we find this on npm?

bnmnetp:
So I was trying this out on my Fork.  I have the following really simple module:

```
var $builtinmodule = function (name) {
    var mod = {};

    mod.__file__ = "/src/lib/vega_caller.js";

    mod.__package__ = Sk.builtin.none.none$;

    mod.render_graph = new Sk.builtin.func(function (vegaspec) {
        Sk.builtin.pyCheckArgs("render_graph", arguments, 1, 1);
        // set actions to false to remove links to editor etc.
        let divid = '#' + Sk.divid + '_graphics';
        try {
            vegaEmbed(divid, Sk.ffi.remapToJs(vegaspec), {actions: true});
    } catch(e) {
        throw new Sk.builtin.Exception("Error in graph specification unable to render the graph\n" \
+ e.message);
    }
    });

    return mod;
}
```

But the new build gives me this:

```
> skulpt@1.0.0 postbuild /Users/bradleymiller/Runestone/skulpt
> node support/build/wrapmodules.js builtin

/Users/bradleymiller/Runestone/skulpt/support/build/wrapmodules.js:44
                                throw new Error("Error minimizing " + fullname + " (uglify js error code: " + result.error + ")");
                                ^

Error: Error minimizing src/lib/altair/vega_caller.js (uglify js error code: SyntaxError: Unexpected token: name «divid», expected: punc «;»)
    at files.forEach (/Users/bradleymiller/Runestone/skulpt/support/build/wrapmodules.js:44:39)
    at Array.forEach (<anonymous>)
    at dirs.forEach (/Users/bradleymiller/Runestone/skulpt/support/build/wrapmodules.js:29:15)
    at Array.forEach (<anonymous>)
    at processDirectories (/Users/bradleymiller/Runestone/skulpt/support/build/wrapmodules.js:26:10)
    at files.forEach (/Users/bradleymiller/Runestone/skulpt/support/build/wrapmodules.js:36:21)
    at Array.forEach (<anonymous>)
    at dirs.forEach (/Users/bradleymiller/Runestone/skulpt/support/build/wrapmodules.js:29:15)
    at Array.forEach (<anonymous>)
    at processDirectories (/Users/bradleymiller/Runestone/skulpt/support/build/wrapmodules.js:26:10)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! skulpt@1.0.0 postbuild: `node support/build/wrapmodules.js builtin`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the skulpt@1.0.0 postbuild script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/bradleymiller/.npm/_logs/2019-06-04T21_56_43_778Z-debug.log
```

I'm not seeing the problem...

support/time-helpers/strptime.js
rixner: I could not find an acceptable strptime on npm.  If you can find one, I'd be happy to use it!

rixner:
@bnmnetp , does the error go away if you use "var" instead of "let"?

rixner:
@bnmnetp, I switched over to ```babel-minify``` which should accept more modern javascript in libraries.  However, it doesn't like you continuing the line with ```\```.  That's completely unnecessary in that expression, so...

FYI, all, when we run the tests, it does not use ```skulpt-stdlib.js``` (nor has it ever).  To actually confirm that the libraries are packaged correctly, I've been using ```npm run btest``` which does load and use ```skulpt-stdlib.js``` in the browser.

I believe the library js code should be minified, as it's just a gratuitous waste of bandwidth/storage not to.  But, that does mean you need to make sure there are no "gotchas" like @bnmnetp just found.  I also exclude things (like ```webgl```, for example) which are completely unsupported in my ecosystem to further save bandwidth/storage.  You can now optionally supply a ```libexcludes.json``` file at the top level to exclude whatever libraries you don't want.

bnmnetp:
Thanks  @rixner -- The \ was an artifact of copy/paste (sorry) it wasn't in the code.  With `babel-minify` Everything builds and checks out in the browser.  I think support for modern Javascript will be a good thing going forward.

I'm going to merge.

haoranyu:
@albertjan Can you please consider updating the package released at https://www.npmjs.com/package/skulpt .

This webpack version will be much popular in the modern community.

Issue #891, opened 2019-05-30
A strange bug in turtle


alexwenbj:
The turtle functions arguments length always one more than the actual arguments length.For example,if I print out the forward(100) argument:
`console.log(arguments)`
it will print out:
`Arguments(2) [100, undefined, callee: (...), Symbol(Symbol.iterator): ƒ]`
If I print out the color("red","yellow") arguments,it will print out:
`Arguments(5) ["red", "yellow", undefined, undefined, undefined, callee: (...), Symbol(Symbol.iterator): ƒ]`

First,the arguments is one more than actual.Second,the undefined value should not be there.

bnmnetp:
@alexwenbj Check out the discussion for PR #850 It should shed some light on this for you.

alexwenbj:
OK,thank you.@bnmnetp

PR #892, opened 2019-05-30
fix the color function for arguments length

alexwenbj:
See: #891  and #850

### DIFF
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -888,7 +888,7 @@ function generateTurtleModule(_target) {
 
         proto.$color = function(color, fill, b, a) {
             if (color !== undefined) {
-                if (arguments.length === 1 || arguments.length >= 3) {
+                if (fill === undefined || b !== undefined) {
                     this._color = createColor(this._colorMode, color, fill, b, a);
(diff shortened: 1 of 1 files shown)


bnmnetp:
@alexwenbj what problem are you trying to fix here?

We should be able to call color (too) many different ways
```
color() --> returns the current colors
color(string) set color and fillcolor
color(string,string) set color and fillcolor
color((tuple), (tuple))
color((tuple))
color(num,num,num) where num is either an int or a float and works depending on cmode
```
I think I have them all. It would be good to document what you are fixing and probably put something into the code to make it clear what is trying to handle which cases.

alexwenbj:
You can try color("red","yellow"),because the `arguments.length` is 5,so the condition `arguments.length>=3` is met,and it will call` createColor(this._colorMode, color, fill, b, a)`,it is not correct here.

alexwenbj:
By the way,why the color()/fillcolor()/pencolor() function have 4 arguments?In Python,they have at most 3 arguments.

albertjan:
@bnmnetp this is the same problem you fixed before. I think this is good to merge. 😄

PR #893, opened 2019-05-30
Fixes to failing Python 3 tests

rixner:
This pull request fixes the Python 3 failures in ```test_defaultdict.py``` and ```test_class.py``` that were discovered in #889.


rixner:
Need to redo this.

Issue #896, opened 2019-06-05
ECMAScript Version


rixner:
The closure compiler is a transpiler, so we can tell it what version of ECMAScript we want it to output.  I think the reasonable options are ECMASCRIPT 5 or ECMASCRIPT 2015.  If we all agree, I can make a trivial pull request to explicitly set the output language.  This will then set the stage for us to use more modern Javascript in the code base without breaking our browser compatibility.  (Note that given that we are already using the closure compiler, there is no need for us to add babel to the build process in order to allow us to write more modern Javascript.)

I will point out, though, that the libraries do **not** go through the closure compiler.  So, they get placed in ```skulpt-stdlib.js``` as-is (though they now get minified).  It likely behooves us to figure out a strategy to transpile the libraries to the same target ECMAScript version as the core of Skulpt....

bnmnetp:
I would certainly cast my vote for ES6.

The libraries are going to be even more tricky because they are a mixture of python and js.  We should figure that out separately from whether we move to ES6 as it will definitely not be trivial.

LeszekSwirski:
+1 to more modern code, particularly since then a Skulpt embedder can choose to have a different compilation target than the default (if e.g. they don't care about IE11) and get better/smaller code as a result.

albertjan:
@bnmnetp the in language should be as high as it will go if you ask me. 

The out language, I'm ok with 2015, but we should be able to do a devbuild that builds to es5 for ie11. Because I know for a fact @trinketapp will want to keep supporting ie11 for a while. I don't know about you @bnmnetp what's your % of users on ie11?

albertjan:
@LeszekSwirski we'll have to start using those new features too right? 😄 Because I think that the codebase is mostly es3 with maybe a few accidental es5 features.

LeszekSwirski:
Well yes :) But imagine the cleanup you'd be able to do with rest parameters alone!

albertjan:
![](https://i.imgflip.com/32q2w9.jpg)

bnmnetp:
@albertjan IE accounts for only 1.3% of the traffic to runestone.

rixner:
Based on this conversation, it seems like everyone is good with ECMAScript 2015 as the default output language?  The question, then, is whether people should just manually change that on their own (by editing ```webpack.config.js```) if they need a different build?  Or if we should provide some sort of flag command to either select among a few options or allow one to specify an arbitrary output language (that the closure compiler supports)?

[...]

PR #897, opened 2019-06-05
Select the output language of the closure compiler

rixner:
A few small housekeeping things from #889, like create the ```tmp``` directory if it doesn't exist and use the production build to test in the web browser.

Mainly, however, sets the output language of the closure compiler to ECMAScript 2015 by default, but lets you select other variants if you prefer, as discussed in #896.  Speak up if you believe the default should be ECMAScript 5.

If people don't like this approach, I'm happy to go back to the drawing board.


support/run/test_template.ejs
```diff
@@ -2,7 +2,7 @@
 <html>
 <head>
     <title></title>
-    <script src="skulpt.js" type="text/javascript"></script>
+    <script src="skulpt.min.js" type="text/javascript"></script>
```
albertjan: is this change on purpose? I thought we used the dev build for browser debugging?

rixner: I needed to test that changing the output language produces a distribution file that still functions in a web browser.  As I thought about it, I started to think that we should be testing on the "real" distribution file under normal circumstances, as well.  The "brun" command still uses the dev build, as that seems more appropriate.  But, what does it matter if all the tests pass on the dev build?  They need to pass on the real build...

Ultimately, I think this is debatable and depends on what one is trying to do at the time.  Perhaps there needs to be a switch to allow either somehow?  Or I can switch it back.

rixner:
Alternatively, maybe we should start with defaulting to ECMAScript 5 to avoid disruption.  If everyone finds themselves then always building ECMAScript 2015 (I'd make that a build command - ```build-es15```), then we can update it?

albertjan:
I think 2015 should be fine for the majority of the users. If they're deploying it to a production environment that requires ES5 and they're building it themselves, they're probably able to figure out how to set that flag.

albertjan:
I do think we should push a es2015 and a es5 version to skulpt-dist, so we can still help those people that don't build skulpt themselves.

support/run/test_template.ejs
albertjan: A future update to this might be detecting which (minified or non minified) is available and adding that to the template.

rixner: Easily done, so I did.

rixner:
@bnmnetp I'm good.  I am just reacting to any changes anyone would like.

rixner:
This has sat with no activity for a week, can we either merge it or explain the problem with it?  Thanks!

@albertjan are you asking me to figure out how to push 2 versions to skulpt-dist?  Do you have a suggestion on how to name them?

```skulpt.min.js```  (es2015)
```skulpt.es5.min.js```

Or something else?

albertjan:
That sounds good to me. i’m not too fussed on the name. Let’s merge this asap. 👍

PR #901, opened 2019-06-07
Update turtle.js

### LINKED ISSUE #899: The ondrag() function can't distinguish different Turtle instance
Run the python demo code:
`https://github.com/python/cpython/blob/master/Lib/turtledemo/colormixer.py`,Skulpt `ondrag()` can't distinguish the different turtles.Drag one turtle,the other two will response too.

alexwenbj:
Fixed #899

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -41,7 +41,8 @@ Sk.python2 = {
     bankers_rounding: false,
     python_version: false,
     dunder_next: false,
-    dunder_round: false,
+    dunder_round: false,    
+    dunder_clear: false,
     exceptions: false,
     no_long_type: false,
     ceil_floor_int: false
@@ -62,6 +63,7 @@ Sk.python3 = {
     python_version: true,
     dunder_next: true,
     dunder_round: true,
+    dunder_clear: true,
     exceptions: true,
     no_long_type: true,
     ceil_floor_int: true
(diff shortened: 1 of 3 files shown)


alexwenbj:
I am not very sure it is a perfect solution, but it works for me.Please review it carefully.

albertjan:
I don't think list has a clear function. 

```
Python 2.7.10 (default, Feb 22 2019, 21:17:52) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.37.14)] on darwin
>>> a = []
>>> a.clear
  File "<stdin>", line 1, in <module>
AttributeError: 'list' object has no attribute 'clear'
```

alexwenbj:
@albertjan https://docs.python.org/3/tutorial/datastructures.html

alexwenbj:
list.clear()
Remove all items from the list. Equivalent to del a[:].

rixner:
@alexwenbj  ```list.clear``` should only exist in Python 3.  This is similar to the ```__round__``` method on the numeric types.  You can see how to pull that off by looking in ```src/env.js```.  You'll want to create a ```clear$``` method that implements the Python 3 clear functionality in ```list``` and then setup things in ```Sk.setup_method_mappings``` as appropriate so that it gets installed for Python 3 as ```clear``` and does not get installed for Python 2.

alexwenbj:
@rixner Got it,and modified list.js and env.js.

src/env.js
```diff
@@ -41,7 +41,8 @@ Sk.python2 = {
     bankers_rounding: false,
     python_version: false,
     dunder_next: false,
-    dunder_round: false,
+    dunder_round: false,    
+    dunder_clear: false,
```
albertjan: please call this `list_clear` 😄 dunder methods are all called like this: `__name__`. So this isn't a dunder method. 😄

alexwenbj: OK.

rixner:
The ```list.clear``` part now looks good to me!  I defer to others on the turtle part.

PR #906, opened 2019-06-17
Processing improvements

cgibin:
This pr contains two changes to processing module. Since they are small, I decided to make them in a single pr.

The first one is related to DOM - it removes jq dependency(this is overkill here - used only once) and removes canvas element manipulations. Direct dom changes make modern virtual dom frameworks unhappy and desyncing might lead to weird behavior. Also I think its not processing wrapper responsibility to care is element visible or exists - someone might want to use hidden element for some reasons for example.

The second change is to make default canvas size useful. Currently without size() function call it sets width and height attrs of canvas to 100. It makes little sense, since in most cases this will be different from canvas element dimensions, and that will make all processing output stretched/shrinked accordingly. With this change, output without explicit call to size() renders properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,15 +10,27 @@ Skulpt is a Javascript implementation of Python 2.x. Python that runs in your br
 
 Skulpt is the brainchild of Scott Graham. See [Skulpt.org](http://skulpt.org) for some early demos of skulpt in action.
 
-Brad Miller has been shepherding the development since sometime in 2010/2011.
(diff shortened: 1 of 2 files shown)


alexwenbj:
@cgibin Would you please add a sample code to show how to use processing for this PR?Thank you.

cgibin:
Well, the main reason was to remove the call to size() in case where canvas size might change. So nothing really interesting here. As a playground I've used doc/static/proctest.html file @alexwenbj

alexwenbj:
@cgibin Thank you.

PR #915, opened 2019-06-25
Implement Python 3 @ operator

rixner:
Title says it all.


src/ast.js
```diff
@@ -247,8 +247,21 @@ var operatorMap = {};
     operatorMap[TOK.T_PERCENT] = Sk.astnodes.Mod;
 }());
 
+function py3_operators(py3) {
```
albertjan: I don't think we should be modifying this operatorMap. We could just check in `getOperator` if it's available when we get the operator right?

rixner: Well, you need to indicate somewhere that @ is or is not an arithmetic operator.  Currently, ```operatorMap``` is where that is done.  Yes, we could modify ```getOperator``` instead, but then it would need to have a notion of whether the operator found in ```operatorMap``` is valid or not.  Requiring another structure to indicate that somehow.  In other words, I believe this is exactly the structure that tells you whether or not the operator is "available", though I would have used the word "valid".

By doing it this way, you update the operator map *once* in ```Sk.configure``` based on whether you are using Python2 or Python3, then ```getOperator``` works correctly.  If we move a check into ```getOperator``` it would have to run that check every time an operator is parsed.

If you have a better way, can you be more concrete about what you mean?

meredydd: FWIW I'm in favour of @rixner's approach here, although I'd prefer a more descriptive name like `setupPy3Operators` to make it more clear that it's a one-time init function.

In fact, my only comment is on introducing the `matmul` function in `type.js` -- I imagine we'll pick up more python-3-specific dunder-methods as we go, so we should go ahead and create a `setupPython3DunderMethods` (pick a name as you prefer).

Otherwise, this change looks awesome!

src/ast.js
```diff
@@ -247,8 +247,21 @@ var operatorMap = {};
...
+    if (py3) {
+        operatorMap[TOK.T_AT] = Sk.astnodes.MatMult;
+    } else {
+        if (operatorMap[TOK.T_AT]) {
+            delete operatorMap[TOK.T_AT];
+        }
+    }
+}
+Sk.exportSymbol("Sk.ast.py3_operators", py3_operators);
```
meredydd: Erm...should we really be exporting as `Sk.ast.py3_operators` when the function is not actually set as an attribute on `Sk.ast` object? This seems a bit odd.

meredydd:
Code looks good to me (not tested, but tests look good and I see they pass), with the exception of two comments I made. Once those are done, I'm +1 to merge.

rixner:
Sorry for the slow response - the latest commits should address your concern @meredydd .

src/ast.js
```diff
@@ -247,7 +247,7 @@ var operatorMap = {};
     operatorMap[TOK.T_PERCENT] = Sk.astnodes.Mod;
 }());
 
-function py3_operators(py3) {
+function setupOperators(py3) {
```
meredydd: Shouldn't this be:
```
Sk.ast.setupOperators = function py3_operators(py3) {
...
}
Sk.exportSymbol("Sk.ast.setupOperators", Sk.ast.setupOperators);
```

Everywhere else I can see that we use `Sk.exportSymbol()`, the thing being exported is already set up with its fully-qualified name.

src/type.js
```diff
@@ -57,7 +57,7 @@ Sk.dunderToSkulpt = {
     "__set__": ["tp$descr_set", 3]
 };
 
-function matmuls(py3) {
+function setupDunderMethods(py3) {
```
meredydd: Same comment as above

meredydd:
I'm still puzzled about the `Sk.exportSymbol()` stuff, but apart from that, this is good to go!

rixner:
@meredydd ```exportSymbol``` places the function into the desired global namespace.  The second argument does not need to match the location that you export it to.  We have always done that, though.   However, I think that was largely to make it easier to find things, not because it is necessary in any way.

Now that things are using node and we can move towards being more modular, presumably we are going to want to stop doing this.  Ideally, you don't want to use ```exportSymbol``` at all (or at least sparingly).  You want to actually export the functions that should be callable externally, and then access them via require'ing the module.

If the difference with the rest of the code base offends you, we can easily change the names so they match.  I just want to be clear that there is no functional reason why this is necessary.

meredydd:
Aha, thank you, that does explain things!

Yeah, especially if `exportSymbol()` is going away one day, it's probably cleaner to arrange things such that it _could_ be a no-op. Thanks for explaining!

rixner:
@meredydd should I take that to mean you are good with how it is now?  Or do you want to change the function names to include the ```Sk.```?  It's trivial to do so, so either way is fine with me.

meredydd:
Could you make the change please? Thanks for tolerating my nits :) To reiterate, this is a great piece of work!

Issue #914, opened 2019-06-20
Support type assignments


acbart:
https://github.com/skulpt/skulpt/blob/7264ca798b85faadc75c4c6583631cfacd979df6/src/ast.js#L2125

Skulpt does not yet support type assignments, based on this note.

I believe the relevant code from ast.c: https://github.com/python/cpython/blob/master/Python/Python-ast.c#L2924

albertjan:
Does it not silently parse type annotations and ignore them now?

acbart:
It supports them in function definitions:

```
def x(a: int, b: str) -> str:
    pass
```

but when you do regular assignments with type annotations, it throws an error.

```
a: int = 0
```

acbart:
Is anyone else working on this? If not, then I can probably do the conversion from the C world to JS. But I wouldn't want to work redundantly.

rixner:
@acbart I am not currently working on it, but I would definitely like to see it added!

PR #920, opened 2019-07-01
Annotations now parseable

acbart:
This doesn't add runtime support or anything, but you can now extract the annotated assignments from the parse tree. I don't think things are set up for the other typed components, and I see various things missing related to NON_LOCAL and whatnot. Seems like there's still a bit of work to be done, but still great to see so much progress being done!


acbart:
#914

acbart:
I believe that should handle the requested changes.

PR #922, opened 2019-07-04
Fix: raise of StringIO.py in python 3

Lifan1998:
My project has been using StringIO and is configured as Python 3, but yesterday I compiled the latest skulpt and I found an error.

main.py:

       from StringIO import StringIO
       print('Hello World!')

output:
       
      SyntaxError: Old raise syntax is not available in Python 3 on line 36

[The cause of the error: ](https://portingguide.readthedocs.io/en/latest/exceptions.html)

- Python 2’s raise statement was designed at a time when exceptions weren’t classes, and an exception’s type, value, and traceback components were three separate objects:

      raise ValueError, 'invalid input'
      raise ValueError, 'invalid input', some_traceback
- In Python 3, one single object includes all information about an exception:

      raise ValueError('invalid input')

      e = ValueError('invalid input')
      e.__traceback__ = some_traceback
      raise e

I found the reason and fixed the bug simply and rudely, but I still don't know why it didn't go wrong before.Please fix this bug in a better way.

main.py:

        from StringIO import StringIO
        print('Hello World!')

output: 
  
        Hello World!


albertjan:
Did you take this from cpython? Or update the one that we had to make it python3 compatible?

Lifan1998:
> Did you take this from cpython? Or update the one that we had to make it python3 compatible?

Cpython? Sorry, I am not good at Python and skulpt, today I found a raise statement on another document.
[What is New in Python 3?](https://www.tutorialspoint.com/python3/python3_whatisnew.htm):

    raise exception
    Python 2 accepts both notations, the 'old' and the 'new' syntax; Python 3 raises a SyntaxError if we do not enclose the exception argument in parenthesis.

    raise IOError, "file error" #This is accepted in Python 2
    raise IOError("file error") #This is also accepted in Python 2
    raise IOError, "file error" #syntax error is raised in Python 3
    raise IOError("file error") #this is the recommended syntax in Python 3


So my previous changes may be wrong, I am very sorry. I changed StringIO.py again.
You can fix it yourself.

I have encountered other problems, so I plan to compile my own skulpt. You can close this pr if you feel there is no need to fix it.

PR #925, opened 2019-07-09
fix cross origin error when ondrag with image turtle

willdonggg:
in turtle.ondrag function hitTest is called, in hitTest it calls context.getImageData(), if the image has diffrent origin, cross origin error will happen


albertjan:
@skulpt/committers this is a safe change afaics.

albertjan:
You've reused this branch for some other things, I'm closing this. Thanks for your time and sorry it took so long to action this.

PR #930, opened 2019-07-12
Performance: Add a fast path for unpacking non-vararg function calls

meredydd:
One performance casualties of our move to the Python 3 grammar was the process for assembling function arguments. Python 3 supports multi-star packed positional args (eg `foo(a,b,*c,d,*e)`, and we were using the code path for assembling those args, even for simple calls). This change adds a fast path when there are no varargs in the call.

This change yields approximately a 10% decrease in the unit test runtime for me (18.99s to 17.00s). Keen to see what others find on their workloads.


albertjan:
Hey @meredydd I think it would also make sense to add an optimisation here: 

https://github.com/skulpt/skulpt/blob/master/src/function.js#L301

```
if (co_argcount === posargs.length) {
       return this.func_code.apply(this.func_globals, args);
}
```

meredydd:
We'd need to be a bit careful -- off the top of my head, that suggested fast-path code won't handle varargs arguments or kw-only args or spurious keyword args correctly. (And it took me several minutes to notice all three of those issues; I would want to sit down and think about it carefully to be sure I hadn't missed one.)

I agree that `function.tp$call` could probably do with optimisation or a fast path or both, but that change is (happily) independent of this one.

rixner:
This is okay, but it's not mitigating the major source of slowdown.

For a program I care about in Chrome 75 that got killed by #905:
pre-905: ~7 seconds
post-905: ~14 seconds
with this pull request: ~13.5 seconds

Clearly not hurting anything, though, and it does get up to 10% on some workloads.  As I said in the initial pull request, ```tp$call``` was a much bigger culprit when I profiled things.

PR #937, opened 2019-07-27
Tokenisation speed-ups

### LINKED ISSUE #927: Parsing is slow because of regexes being repeatedly re-compiled in Chrome Devtools
#926 left me very curious about what was causing the slowdown. I did a lot of digging and manual profiling, and I think I've determined the main cause. A big chunk of time is spent calling `RegExp` inside the tokenizer, which is a very costly operation. I'm guessing that something in DevTools prevents it from caching the created regexes.

I'm not clear that this affects anyone not using DevTools in Chrome, but I know I'll be modifying this in my local fork. My plan is to simply extract out the literal values in `isidentifier`, and turn the Python3 dependent ones into something that gets setup during `Parser.setup`. Is there any interest in making that a pull request?

Offending lines: 

* https://github.com/skulpt/skulpt/blob/master/src/tokenize.js#L107
* https://github.com/skulpt/skulpt/blob/master/src/tokenize.js#L365
* https://github.com/skulpt/skulpt/blob/master/src/tokenize.js#L392
* https://github.com/skulpt/skulpt/blob/master/src/tokenize.js#L425

LeszekSwirski:
Caches a few regexes in the tokeniser. I see a unit test speedup from 16-17s to 12-13s (!) locally.

Fixes #927.

### DIFF
--- a/src/tokenize.js
+++ b/src/tokenize.js
@@ -79,14 +79,7 @@ function rstrip (input, what) {
     return input.substring(0, i);
 }
 
-/**
- * test if string is an identifier
- *
- * @param {str} string
- * @returns {boolean}
- */
-function isidentifier(str) {
-    var normalized = str.normalize('NFKC');
+const IS_IDENTIFIER_REGEX = (function() {
     var the_underscore = '_';
     var Lu = '[A-Z]';
     var Ll = '[a-z]';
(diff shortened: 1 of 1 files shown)


albertjan:
Nice one!

PR #938, opened 2019-07-28
Update Python 3 Tests

rixner:
This pull requests ports the majority of the test/run tests to Python 3 unit tests.  There are some that are not so easy to do (like testing quit, assert, etc.) that didn't make it.  But nearly all are now here.

Along with this come many small/medium fixes to Python 3 behavior, including changing filter/map/zip to work as they do in Python 3, changing the use of ```__nonzero__```, ```__cmp__```, and the like as per Python 3 semantics, not printing the "L" at the end of longs, fixing scientific notation formatting for Python 3, etc.

I would be happy to delete the entire test/run directory, but I'm not sure if others would like to keep it around.  Those tests are more finicky than the unit tests, and this pull request ensures that all those capabilities will still be tested (albeit in Python 3 mode).


src/builtin.js
```diff
@@ -536,10 +536,9 @@ Sk.builtin.int2str_ = function helper_ (x, radix, prefix) {
     var str = "";
     if (x instanceof Sk.builtin.lng) {
         suffix = "";
-        if (radix !== 2) {
+        if (radix !== 2 && (!(Sk.__future__.python3))) {
```
albertjan: there's an `L` suffix option on `__future__`

src/builtindict.js
```diff
@@ -110,4 +107,17 @@ Sk.builtins = {
     "coerce"    : Sk.builtin.coerce,
     "intern"    : Sk.builtin.intern
 };
-Sk.exportSymbol("Sk.builtins", Sk.builtins);
+
+Sk.setupObjects = function (py3) {
+    if (py3) {
```
albertjan: should we also remove `reduce` in python three? it's moved to functools.

src/long.js
```diff
@@ -814,7 +814,13 @@ Sk.builtin.lng.prototype.ob$ge = function (other) {
 };
 
 Sk.builtin.lng.prototype.$r = function () {
-    return new Sk.builtin.str(this.str$(10, true) + "L");
+    var suffix;
+    if (Sk.__future__.python3) {
```
albertjan: `L` suffix flag here too

src/misceval.js
```diff
@@ -624,12 +631,22 @@ Sk.misceval.isTrue = function (x) {
     if (x.constructor === Sk.builtin.float_) {
         return x.v !== 0;
     }
-    if (x["__nonzero__"]) {
-        ret = Sk.misceval.callsimArray(x["__nonzero__"], [x]);
-        if (!Sk.builtin.checkInt(ret)) {
-            throw new Sk.builtin.TypeError("__nonzero__ should return an int");
+    if (Sk.__future__.python3) {
+        if (x["__bool__"]) {
```
albertjan: why do you use `["__bool__"]` here instead of `__bool__`

src/builtin.js
rixner: I asked about this in #905, but I guess I wasn't specific enough.  What's the point of carrying around the specific flags anymore (except those that are in CPython 2 to allow you to turn on some Python 3 features individually)?

I can change this, but I'd prefer to delete the L suffix future flag and make it use the Python3 future flag in the other places it's used.

src/builtindict.js
rixner: We can do that, but shouldn't someone implement ```functools``` first and remove it from builtins as a part of that?  It seems like a better story if someone asks "where did reduce go?" to be able to answer "it moved to functools, where it belongs in Python 3", rather than "it's gone".

src/misceval.js
rixner: Because the access to ```["__len__"]``` right below this was already there, so we maintained consistency with the existing code.  I can trivially switch them all over if you'd like...

src/builtindict.js
albertjan: I agree I thought we had some functools already 😄

src/builtin.js
albertjan: Yeah that's fine with me too.

src/misceval.js
albertjan: If closure is happy with it I think that would make sense.

rixner: Just to be clear, do you think I should make them all like ```x["__bool__"]``` or ```x.__bool__```?  I think you mean the latter...

albertjan:
Sorry I let this sit for a while @rixner I was on holiday 😄. Please @skulpt/committers merge if you concur with me!

rixner:
@albertjan no problem.  (I was trying to decide if I should poke people, though.)  I hope you had a good time!

bnmnetp:
LGTM.

I am very happy to eliminate the test/run folder with real unittest replacements!

albertjan:
> I am very happy to eliminate the test/run folder with real unittest replacements!

Lets not do that they were very very very useful when updating the compiler.

rixner:
@albertjan can I perhaps remove the running of the tests in the run folder when testing/creating the distribution, but leave them there for people to run independently if they need to?  I am with @bnmnetp on eliminating them entirely, but defer to the collective wisdom of the group.

bnmnetp:
This PR does not eliminate them, so I can go ahead and merge I think....   Any last objections?

PR #946, opened 2019-08-18
add turtle.color and turtle.bgcolor for  turtle.js

Dustyposa:
The same as turtle in python.
The colormode  attribute is a global attribute in python.
And can use 
```python
import turtle
turtle.Screen().colormode()
turtle.bgcolor()
turtle.colormode()
````
the same use way as python

### DIFF
--- a/src/str.js
+++ b/src/str.js
@@ -939,7 +939,7 @@ Sk.builtin.str.prototype.nb$remainder = function (rhs) {
         fieldWidth = Sk.builtin.asnum$(fieldWidth);
         precision = Sk.builtin.asnum$(precision);
 
-        if (mappingKey === undefined || mappingKey === "") {
+        if ((mappingKey === undefined || mappingKey === "") && conversionType != "%") {
             i = index++;
         } // ff passes '' not undef for some reason
 
--- a/test/unit/test_format.py
+++ b/test/unit/test_format.py
@@ -201,6 +201,9 @@ class FormatTest(unittest.TestCase):
         self.doboth("%s",{'a':1}, "{'a': 1}")
         self.doboth("%s",[], "[]")
 
+        # test escaping
+        self.doboth("%d%%",(100),"100%")
+
         # alternate float formatting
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by albertjan** [comment]
Hi! thanks for your contribution! Can you undo the indentation changes you did? I agree that the indentation is incorrect but this creates very difficult merges for all our downstream consumers. 😄

**Comment by Dustyposa** [comment]
Oh, Ok. I use pycharm. used to push `ctrl + alt + l`.. you know,  :）

**Comment by albertjan** [comment]
totally understand! 😄

**Comment by Dustyposa** [comment]
I fixed indentation. :)

**Comment by albertjan** [comment]
@bnmnetp can you easily check how many times is `colorMode` on `Turtle` used in your students' code base?

**Comment by alexwenbj** [comment]
There is a problem for the colormode() function.This function can accept the any integer and float value as parameter.If the value is not equal to 255,the colormode will be set to 1.0. Your code only accept 1.0 and 255.It is not correct.

**Comment by Dustyposa** [comment]
@albertjan 
Oh, thanks. I missed it.
Now, I fixed it,  pushed a new version for turtle.
How about it?

**Comment by albertjan** [comment]
@Dustyposa I'm a little hesitant to remove the function from the Turtle object as it may be used in students code. Or am I misunderstanding that it's removed from turtle?

**Comment by Dustyposa** [comment]
@albertjan In python. `Turtle obj` hasn't the function `colormode`. 
```python
import turtle

t = turtle.Turtle()
t.colormode()
>>> AttributeError: 'Turtle' object has no attribute 'colormode'
```
I want students don't change their code, and can run their's code by python interpreter either.
By the way, I fixed the #954 in now version.
How about this code? I don't have a good idea for it. It's not a Class.

bnmnetp:
@albertjan -- not used very often at all.around 530 times out of 8+ million in the last 2 years.  Also, it is not mentioned in the text so I'm not explicitly telling students to use it.

PR #948, opened 2019-08-21
Stop %% escapes in %-formatted string from consuming a value

meredydd:
When formatting a string with the % operator, the '%' character
is encoded '%%' (eg '%d%%' % 100 -> '100%').

Skulpt's implementation was consuming an element from the value tuple
for '%%' literals, which was incorrect. This change fixes that, and
adds a test.


meredydd:
I say this change is small enough to merge with one approval. @skulpt/committers, object if you dare ;)

PR #950, opened 2019-08-23
[WIP] Fast path for function tp$call()

meredydd:
This change adds a couple of fast paths to the `tp$call` method of `Sk.builtin.func`. Although the don't break any of the tests, they also make no difference to the amount of time it takes to run the unit tests on my machine.

If it doesn't bring any benefit, we shouldn't carry the extra complexity, but I'm putting this out as a WIP so that @rixner and others with workloads that were slowed down a lot by the 2->3 migration can try it out.

I am inclined to interpret this as evidence that the `tp$call` function was not the bottleneck we were supposing it to be...

### DIFF
--- a/src/function.js
+++ b/src/function.js
@@ -290,8 +290,9 @@ Sk.builtin.func.prototype.tp$getname = function () {
 };
 
 Sk.builtin.func.prototype.tp$call = function (posargs, kw) {
-    // This function is a logical Javascript port of
-    // _PyEval_EvalCodeWithName, and follows its logic.
+    // The rest of this function is a logical Javascript port of
+    // _PyEval_EvalCodeWithName, and follows its logic,
+    // plus fast-paths imported from _PyFunction_FastCall* as marked
 
     let co_argcount = this.func_code.co_argcount;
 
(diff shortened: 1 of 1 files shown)


rixner:
@meredydd I have been swamped for the past month or so.  It is definitely on my todo list to come up with a benchmark I can release that demonstrates the slowdowns I am seeing.  I really don't think the tests serve as good performance benchmarks.  I think they are dominated by compile time.

I will run this and report back as soon as I can, but the semester starts for me on Monday and I have countless logistics to deal with at this point in time.

Regardless, I really appreciate the effort to try to get the performance back!  Thanks!

meredydd:
No problem. I'm happy to leave this PR open until you have time to take a swing at it!

albertjan:
I've run [`pystone.py`](https://github.com/python/cpython/blob/2.7/Lib/test/pystone.py) through this branch. There's a small performance gain. Results below: 

## Before

```
Pystone(1.1) time for 50000 passes = 5.184000015258789
This machine benchmarks at 9645.061700005408 pystones/second
-----
Run time: 5.611s

Pystone(1.1) time for 50000 passes = 5.021999835968018
This machine benchmarks at 9956.193077087632 pystones/second
-----
Run time: 5.428s

Pystone(1.1) time for 50000 passes = 5.148000001907349
This machine benchmarks at 9712.5097089112 pystones/second
-----
Run time: 5.561s
```

## After

```
Pystone(1.1) time for 50000 passes = 4.926999807357788
This machine benchmarks at 10148.163579249986 pystones/second
-----
Run time: 5.324s

Pystone(1.1) time for 50000 passes = 4.92799973487854
This machine benchmarks at 10146.104441954145 pystones/second
-----
Run time: 5.325s

Pystone(1.1) time for 50000 passes = 5.1600000858306885
This machine benchmarks at 9689.922319439398 pystones/second
-----
Run time: 5.553s

Pystone(1.1) time for 50000 passes = 4.995999813079834
This machine benchmarks at 10008.006779563308 pystones/second
-----
Run time: 5.39s

rixner:
When I run this in node, I don't see any measurable benefits.  When I run it in Chrome, I see maybe a 5% speedup?  This is on some code that I know exercises the fast path a significant amount.

I can't decide if what I'm seeing in Chrome is statistically significant or just my imagination (as the numbers bounce around a lot as usual).  It's not a large difference, but it does seem like the fast path numbers are consistently better than not having it.

The question is whether the small difference warrants the small increase in complexity.  I think I come down in favor of yes?

Regardless, I don't think we can argue that ```tp$call``` is not the bottleneck.  It still shows up at the very top of the profile output (both with and without this fast path).  It may or may not be the bottleneck, but I don't think this pull request is conclusive evidence that it is not.

Node: 8.10
Chrome: 79.0.3945.130

meredydd:
This PR is superseded by #1009

PR #952, opened 2019-08-23
[Don't merge yet] Support the <> operator (Python 2 only)

meredydd:
This change reintroduces support for the Python-2-only `<>` operator. Rather than mapping it to `NOTEQUAL` (like Python 2 does), we introduce a new token type for "not equal, but only if you're in Python 2", and then error out in `ast.js` if we see it in Python 3 mode.

This is really my first time touching the tokeniser, and I am still not 100% certain I know how all the different repetitions of the token table work together -- so, careful review requested from @albertjan!


albertjan:
:trollface: 

```
Python 3.6.8 (default, Jul 31 2019, 15:26:40) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.4)] on darwin
>>> from __future__ import barry_as_FLUFL
>>> 1<>2
True
>>> 
```

src/pgen/parser/Grammar.txt
```diff
@@ -93,8 +93,7 @@ or_test: and_test ('or' and_test)*
 and_test: not_test ('and' not_test)*
 not_test: 'not' not_test | comparison
 comparison: expr (comp_op expr)*
-# <> isn't actually a valid comparison operator in Python. It's here for the
```
albertjan: let's leave this in 😄

src/pgen/parser/token.py
```diff
@@ -67,10 +67,12 @@
 AWAIT = 54
 ASYNC = 55
 # Extras hacked in
-COMMENT = 56
-NL = 57
-ERRORTOKEN = 58
-N_TOKENS = 59
+NOTEQUAL_PY2ONLY = 56
```
albertjan: I feel slightly queasy about adding a token and diverging from python on token numbers, but I haven't been able to figure out what a problem would be.

gen/parse_tables.js
```diff
@@ -25,7 +25,7 @@ Sk.OpMap = {
 "@=": Sk.token.tokens.T_ATEQUAL,
 "==": Sk.token.tokens.T_EQEQUAL,
 "!=": Sk.token.tokens.T_NOTEQUAL,
-"<>": Sk.token.tokens.T_NOTEQUAL,
+"<>": Sk.token.tokens.T_NOTEQUAL_PY2ONLY,
```
albertjan: In cpython, they both have the `NOTEQUAL` token name

src/ast.js
```diff
@@ -293,6 +293,11 @@ function astForCompOp (c, n) {
                 return Sk.astnodes.GtE;
             case TOK.T_NOTEQUAL:
                 return Sk.astnodes.NotEq;
+            case TOK.T_NOTEQUAL_PY2ONLY:
+                if (Sk.__future__.python3) {
+                    throw new Sk.builtin.SyntaxError("The <> operator is not supported in Python 3", c.c_filename, n.lineno);
```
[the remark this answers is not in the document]
meredydd: I'm happy with being strictly more helpful than the normal Python parser, as long as the type of the exception is correct.

meredydd:
I think you're right, @albertjan: We should throw this in the tokenizer instead of the parser. This would align us with CPython, and also get rid of the ugly weird extra token.

However...I'm afraid the tokenizer is your turf, so I'd appreciate a little help implementing this the right way!

Anvil will carry this patch out-of-tree until we can get something nicer in.

Issue #954, opened 2019-08-26
enumerate() method not support "start" argument name


alexwenbj:
For example:
```
grocery = ['bread', 'milk', 'butter']
enumerateGrocery = enumerate(grocery, start = 10)
print(list(enumerateGrocery))
```
will raise an JS exception:`<native JS>() got an unexpected keyword argument 'start'`.

alexwenbj:
@Dustyposa It seems your fix to this issue still not works for me.Have you tested it?Thank you.

Dustyposa:
@alexwenbj can you share your test code? I add the code in test directory.

alexwenbj:
Just I listed in this issue:
```
grocery = ['bread', 'milk', 'butter']
enumerateGrocery = enumerate(grocery, start = 10)
print(list(enumerateGrocery))
```
Thank you.

Dustyposa:
@alexwenbj Did you can run cmd `npm run dist` and pass all test?  
I can't pass 
```Tests failed in  CounterTests.test_most_common
Ran 5 tests, passed: 4 failed: 1
```

alexwenbj:
I didn't use the new test case,just add your code to the enumerate.js,it passed and generated the built dist file.

Dustyposa:
emm..now I can't build dist file..I don't know way..I can't past all test. 
And even I only clone the master branch, then run `npm run dist`, didn't add any code.
It's so sad....

alexwenbj:
Run `npm install `first? @Dustyposa

Dustyposa:
@alexwenbj Only have one test can't pass, 
```
Tests failed in  CounterTests.test_most_common
Ran 5 tests, passed: 4 failed: 1
```

alexwenbj:
No idea for that...

Dustyposa:
Your Skulpt is latest version?

alexwenbj:
Not the latest version.I cloned it in July.

Dustyposa:
@alexwenbj  Sorry,It's a sand news,, my env have some thing wrong. Tomorrow, i use another computer to add test example.

Dustyposa:
@alexwenbj 
https://github.com/Dustyposa/skulpt/blob/af73ebfdeec61f19f5bed2df26912651bec9a514/test/unit3/test_enumerate.py#L98-L101
I add your eg in test, and can pass it, how about update a new version for you application.

alexwenbj:
[...]

PR #961, opened 2019-09-10
fix enumerate(Iterable, start=1)

Dustyposa:
fix enumerate(Iterable, start=1) add add test_case 
#954

### DIFF
--- a/src/enumerate.js
+++ b/src/enumerate.js
@@ -57,6 +57,11 @@ Sk.builtin.enumerate.prototype.next$ = function (self) {
     return self.tp$iternext();
 };
 
+Sk.builtin.enumerate.co_varnames = ["", "start"];
+Sk.builtin.enumerate.co_argcount = 2;
+Sk.builtin.enumerate.$defaults = [Sk.builtin.none.none$,0];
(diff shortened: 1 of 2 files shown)


src/enumerate.js
```diff
@@ -57,6 +57,11 @@ Sk.builtin.enumerate.prototype.next$ = function (self) {
     return self.tp$iternext();
 };
 
+Sk.builtin.enumerate.co_varnames = ["", "start"];
```
albertjan: lets call the first argument "iterable" 😄

albertjan:
Thank you! Looks good! One small comment. 😄

src/enumerate.js
Dustyposa: > Thank you! Looks good! One small comment.
Ok, I got it, I change the argument will get twice commit, 
Do I need to rebase the branch and make the commit to be once?

Dustyposa: > lets call the first argument "iterable"
Ok, I got it, I change the argument will get twice commit,
Do I need to rebase the branch and make the commit to be once?

albertjan:
Merging this as it's a small change.

PR #969, opened 2019-10-09
fix #968 

### LINKED ISSUE #968: str, repr problem
With feature Sk.python3 enabled, both `repr(0.1 + 0.2)` and `str(0.1 + 0.2)` return 0.3, but they should be `0.30000000000000004`

Dustyposa:
fix #968  and add test for it.
Please code review. :)
By the way, has any way to achieve the `eval` function? 
The function `Sk.builtin.eval_` hasn't achieve.

### DIFF
--- a/src/float.js
+++ b/src/float.js
@@ -884,9 +884,10 @@ Sk.builtin.float_.prototype.str$ = function (base, sign) {
 
     if (base === undefined || base === 10) {
         if (Sk.__future__.python3) {
+            // 17 would get another number string
             var tmpCheckString = work.toString();
             var tmpCheckPrecision = work.toPrecision(17);
-            tmp = tmpCheckString === tmpCheckPrecision?  tmpCheckPrecision: work.toPrecision(16);
+            tmp = tmpCheckString === tmpCheckPrecision ? tmpCheckPrecision : work.toPrecision(16);
         } else {
             tmp = work.toPrecision(12);
         }
--- a/src/float.js
+++ b/src/float.js
@@ -884,11 +884,13 @@ Sk.builtin.float_.prototype.str$ = function (base, sign) {
 
     if (base === undefined || base === 10) {
         if (Sk.__future__.python3) {
-            tmp = work.toPrecision(16);
+            // 17 would get another number string
+            var tmpCheckString = work.toString();
+            var tmpCheckPrecision = work.toPrecision(17);
+            tmp = tmpCheckString === tmpCheckPrecision?  tmpCheckPrecision: work.toPrecision(16);
         } else {
             tmp = work.toPrecision(12);
         }
-        
 
         // transform fractions with 4 or more leading zeroes into exponents
         idx = tmp.indexOf(".");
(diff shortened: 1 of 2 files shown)


src/float.js
```diff
@@ -884,11 +884,13 @@ Sk.builtin.float_.prototype.str$ = function (base, sign) {
 
     if (base === undefined || base === 10) {
         if (Sk.__future__.python3) {
-            tmp = work.toPrecision(16);
+            // 17 would get another number string
+            var tmpCheckString = work.toString();
+            var tmpCheckPrecision = work.toPrecision(17);
```
albertjan: why can't we use `16` here?

Dustyposa:
```javascript
var s = 0.1 + 0.2
s.toPrecision(17)
>>> "0.30000000000000004"
s.toPrecision(16)
>>> "0.3000000000000000"
```
If use 16, the tmp finally is 0.3 not 0.30000000000000004

Dustyposa: but when the number length more than 17:
```javascript
var s = 0.30000000000000000000000000004
s.toPrecision(17)
>>> "0.29999999999999999"
// except： 0.3
s.toString()
>>> 0.3
// this is right, but not all s.toString() is right, like str(1e1) -> 10.0
// but 1e1.toString() is 10 
// so I compare s.toString() with s.toPrecision(17) to guaranteed decimal.
```

albertjan: wow 😄

src/float.js
```diff
@@ -884,11 +884,13 @@ Sk.builtin.float_.prototype.str$ = function (base, sign) {
...
+            tmp = tmpCheckString === tmpCheckPrecision?  tmpCheckPrecision: work.toPrecision(16);
```
albertjan: can you clean up the white space here? thanks

Dustyposa: OK

Dustyposa: Now is ok？

Dustyposa:
@skulpt/committers  is not a 404 page?

Dustyposa:
that's so sad..... :pensive:

albertjan:
I can't change it but it's a group of people that have the ability to merge PR's @Dustyposa and I would like one of them to have a look at this before we do.

Dustyposa:
So that's it! :smiley: 
Got it.

bnmnetp:
I get it that 0.3 is not the same as what you see in "real" Python, but I'm constantly having to explain away the .0000000001 for real python to students, which quite honestly makes them a little skeptical about whether this python thing really works all that well. Why do we want to perpetuate Python's bad math further?

Dustyposa:
I got it, how about this way which make this to be a Skulpt feature.
```javascript
            if (Sk.CythonFloat){
                var tmpCheckString = work.toString();
                var tmpCheckPrecision = work.toPrecision(17);
                tmp = tmpCheckString === tmpCheckPrecision ? tmpCheckPrecision : work.toPrecision(16);
            } else {
                tmp = work.toPrecision(16);
            }
```
Sometimes, we hope some skulpt's feature with the same as cython. Because the user don't know that Python is a standard, CPython is an implementation of that standard. They only want to get the same result with cython( even they don't know what is cython, call it Python).

albertjan:
Since @rixner and @bnmnetp voted agains this I'm gonna close it. Thanks for your input though @Dustyposa.

PR #972, opened 2019-10-25
Obey __all__ in wildcard imports

meredydd:
`import *` wasn't obeying `__all__`. Now it does.

(NB - it does _not_ currently support importing submodules via `__all__`, but this is a strict improvement)


meredydd:
Small change, merging.

PR #974, opened 2019-10-25
Make tokenization errors come out as SyntaxErrors with correct line numbers

meredydd:
The new tokenizer currently emits weird raw `Error`s without consistent line information. Make these `SyntaxError`s instead, with filenames and line numbers.


meredydd:
Small, self-approving.

PR #975, opened 2019-10-28
F-strings!

meredydd:
Everybody's favourite Python 3 feature - now in Skulpt!

f-strings now parse, compile, and execute, calling the `format()` builtin to format their values.

One caveat: Most format strings do not yet work. This is because all our formatting implementation is stuck in `str.format` rather than being accessible via `format()`. I consider this to be a separate issue to the question of whether f-strings actually work, and I am happy to defer this until a new PR.


src/ast.js
```diff
@@ -2526,34 +2808,8 @@ function ast_for_atom(c, n)
             return new Sk.astnodes.Name(name, Sk.astnodes.Load, LINENO(n), n.col_offset,
                         n.end_lineno, n.end_col_offset);
         }
-        case TOK.T_STRING: {
-            var str = parsestrplus(c, n);
-            // if (!str) {
```
albertjan: ha, thanks for cleaning this up

src/ast.js
```diff
@@ -2663,114 +2919,6 @@ function ast_for_setdisplay(c, n) {
     return new Sk.astnodes.Set(elts, LINENO(n), n.col_offset);
 }
 
-
-function astForAtom(c, n) {
```
albertjan: 🔥

repl/repl.js
```diff
@@ -41,7 +41,11 @@ var compilableLines = [],
     lines = [],
     origLines;
 
-console.log("Python 2.6(ish) (skulpt, " + new Date() + ")");
+if (Sk.__future__.python3) {
+    console.log("Python 3.7(ish) (skulpt, " + new Date() + ")");
```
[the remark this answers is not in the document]
meredydd: It was necessary to get `datetime.datetime.__format__` working for the tests!

meredydd:
> How much of this: https://github.com/python/cpython/blob/3.7/Lib/test/test_fstring.py would pass if we added it @meredydd?

About 1/4, I'd guess. Most of those tests lean heavily on the `ast` module, and various `eval`-y things (to test SyntaxError generation), which we don't have. (Although I did consider implementing them for the sake of the tests!)

It also exercises the numerical formatting code, which I didn't include in this PR (although it could use a proper workout - we keep getting reports of flakiness.)

acbart:
Chiming in - I do have the `ast` module and `exec` working on my forks more or less. It'd be nice to get those integrated upstream at some point. They'd probably save you a lot of base effort, if nothing else.

https://github.com/blockpy-edu/skulpt/blob/master/src/lib/ast.js

https://github.com/blockpy-edu/skulpt/blob/master/src/builtin.js#L1424

meredydd:
Asking all @skulpt/committers -- How close are we to getting this approved?

Do we need to add enough functionality to satisfy (some of) the Python f-string tests first? Getting any useful portion of those working would probably require pulling out the numerical formatting code from `str`, which I had been planning to do as a separate PR.

(I don't think we should be mixing "get `eval` and/or `ast` working in Skulpt" into this PR, although I would love _love_ **_love_** to see a PR from @acbart for that!)

Having this unmerged is making it difficult to maintain other PRs/internal changes that build on it, because they'll conflict, so it would be good to know what we'll need to get this approved.

albertjan:
This looks great, I would like to see the tests from cpython added, we can comment out the broken ones for now. 😄

alexwenbj:
As @meredydd said, maybe you guys should approve this PR first and let things move on.

meredydd:
I'm happy to import the CPython tests, and comment out 90% of them. Just need to wait for my next time-slice for Skulpt work!

meredydd:
Full marks to @albertjan for insisting on the CPython test suite - that shook out a few nice bugs! I've got it running (with lots commented out), and passing, and I think this is now good to merge.

meredydd:
@skulpt/committers, can I re-up this for commit please?

It's passing all of CPython's tests (except the ones Skulpt can't run for unrelated reasons), and it's been approved by @albertjan.

meredydd:
(Rebased because #986 has now landed)

src/abstract.js
```diff
@@ -711,7 +711,7 @@ Sk.abstr.objectFormat = function (obj, format_spec) {
     // Find the (unbound!) __format__ method (a borrowed reference)
     meth = Sk.abstr.lookupSpecial(obj, Sk.builtin.str.$format);
     if (meth == null) {
-        throw new Sk.builtin.TypeError("Type " + Sk.abstr.typeName(obj) + " doesn't define __format__");
+        return Sk.misceval.callsimArray(Sk.builtin.object.prototype["__format__"], [obj, format_spec]);
```
rixner: Why do we need this here?  Shouldn't lookupSpecial have found it?  This feels like a step backwards to have to look it up on the prototype after lookupSpecial...  I thought we had moved towards not doing this in new places?

rixner:
@meredydd sorry I took so long to look at this.  I have the one comment, but otherwise, it seems reasonable (especially since you added a (small) subset of the CPython tests).

src/abstract.js
meredydd: It's to handle old-style classes in Python 2, which do not inherit from `object`.

rixner: Just to confirm, in Python 3, ```lookupSpecial``` will always succeed, right?

meredydd: Yep!

PR #976, opened 2019-11-02
Tokenise ellipses

meredydd:
Skulpt currently barfs on ellipses. This PR is a partial improvement -- it gets the `T_ELLIPSIS` as far as `ast.js`, where it still errors on the ellipsis literal (`...`), but at least gets it right in deeply nested relative imports (`from ... import foo`).

This is my first foray into the tokenizer, so review requested from @albertjan please!


meredydd:
I'm declaring this a minor change, and merging.

albertjan:
👍

PR #977, opened 2019-11-06
Support multiple decorators + class decorators (+ symtable fix)

meredydd:
Previously, we only supported up to 1 decorator per function, and did not support class decorators at all. This PR fixes this.

---

In the process of writing the tests for this PR (which are quite closure-heavy), I discovered a nasty `symtable` issue -- namely, that `ste.hasFree` is not set for functions that _contain_ functions with free variables. For example:
```python
def f():
    v = 4
    def g():
        def h():
            return v+1
```
`symtable.js` did not set the `hasFree` flag on `g()`, even though it correctly notices that it has a free variable (`v`), because it computed the `hasFree` flag *before* walking child scopes (in this case, `h()`).

This actually appears to be a logic bug inherited from Python 2's `symtable.c`, which doesn't matter there because CPython's compiler doesn't rely on the `ste_free`(/`hasFree`) flag. Our compiler does, so I have added logic to keep that flag up to date when we gain free variables from child blocks.

(Why yes, it has been a long afternoon. Why do you ask? ;) )

### DIFF
--- a/test/unit3/test_decorators.py
+++ b/test/unit3/test_decorators.py
@@ -312,13 +312,13 @@ class TestDecorators(unittest.TestCase):
                 return wrapped
             return wrap
 
-        @decorate('first')
-        @decorate('second')
+        @decorate('outer')
+        @decorate('inner')
         def f(x):
             return x
 
         self.assertEqual(42, f(42))
-        self.assertEqual(['first', 'second'], calls)
+        self.assertEqual(['inner', 'outer'], calls)
         
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


meredydd:
Declaring this a minor change, merging.

test/unit3/test_decorators.py
```diff
@@ -302,5 +302,39 @@ def f(cls, arg): return (cls, arg)
...
+            return wrap
+
+        @decorate('outer')
+        @decorate('inner')
+        def f(x):
+            return x
+
+        self.assertEqual(42, f(42))
+        self.assertEqual(['inner', 'outer'], calls)
+
```
s-cork, line 322: @meredydd I think this should be `['outer', 'inner']`

see #1147

PR #983, opened 2019-12-04
Proper unicode support (str/bytes)

### LINKED ISSUE #871: Some character make result wrong in format.

meredydd:
Guido's original reason for moving to Python 3 was so that `str` could be sensible and Unicode-y. This PR makes Skulpt Unicode-aware, and provides the Python 3 `str`/`bytes` interface.

How this change works:

* Skulpt strings are still stored as Javascript strings (which means UTF-16, bletch, but that's the format expected by everything else we touch). Python source code is a Javascript string. No UTF-8 anywhere. (Except if the user calls `encode()` -- but that's a `bytes`, not a `str`.)

* `bytes` is a JS "byte-string" -- a JS string using only chars 0-255. This keeps them very similar to `str`s, which means we can share most of the implementation :)

* `str` is now indexed/sliced/etc by codepoints, not JS string indices. These are _usually_ the same thing, but if you have an astral character they diverge. Eg the string `'\U0001f355!'` ("🍕!") is 2 codepoints long but 3 JS chars long (because the pizza emoji is represented by a surrogate pair). If you're manipulating that string in Python, `s[1]` needs to be `'!'`, not the second half of a surrogate pair.

  Working out which codepoint index corresponds to which JS string index is slow (you have to grovel over the whole string looking for surrogate pairs), so we only compute it if necessary. The first time we do anything that needs to know how many codepoints are in a string, or where they are, the `$hasAstralCodePoints()` method searches for astral characters. _If_ this string has any, we compute an array of codepoints and their position in the string. Once computed, we use this array for all length calculations, slicing, iteration, etc.

  Example: For the string `"🍕!"`, the `codepoints` array is `[0, 2]` (the first codepoint starts at string index 0 and is 2 JS-chars long, and the second codepoint starts at string index 2).

* We explicitly _do not_ make much effort to support Python 2 unicode semantics. `unicode` is an alias for `str`, and all of Python 2's implicit encoding/decoding basically doesn't happen. The goal, as for all our Python 2 support, is "no more broken than we already were".

* I've cleared up some encoding confusion in the parser (Kill UTF-8 with fire! All input data is JS strings! Also, the `\uXXXX`/`\UXXXXXXXX` escape codes now work.)

I would particularly like input from our international users (hi @Dustyposa - This should fix #871 for you!). This should improve things significantly for you!

----

**Note:** This change is built on top of my F-strings branch (#975), because they both touch the string parsing code.

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -2241,7 +2241,7 @@ function astForIfexpr (c, n) {
  * s is a python-style string literal, including quote characters and u/r/b
  * prefixes. Returns [decoded string object, is-an-fstring]
  */
-function parsestr (c, s) {
+function parsestr (c, n, s) {
     var encodeUtf8 = function (s) {
         return unescape(encodeURIComponent(s));
     };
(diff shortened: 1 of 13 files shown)


alexwenbj:
Great!Hope all these 4 PR will merge soon.I just can't wait to pull it.:-)

LeszekSwirski:
> bytes is a JS "byte-string" -- a JS string using only chars 0-255. It's a little inefficient, but it's the simplest way to do it

Why inefficient? Most JS engines optimise the string representation to one byte (Latin 1) if all the string values are < 256.

Dustyposa:
Oh awesome ! It's a great progress,we can get new features that make it's easy to use.
Thanks so much！
I can't wait to pull it.：）

meredydd:
> > bytes is a JS "byte-string" -- a JS string using only chars 0-255. It's a little inefficient, but it's the simplest way to do it
> 
> Why inefficient? Most JS engines optimise the string representation to one byte (Latin 1) if all the string values are < 256.

I stand corrected! In which case, it is a brilliantly efficient scheme, all of whose advantages I definitely thought of beforehand. ;)

rixner:
@meredydd, I have an implementation of ```bytes``` that uses ```Uint8Array```.  We had it at the beginning of the fall, but didn't get around to creating a pull request (you may have noticed I dropped off the planet this fall - I've been hugely busy!).  I don't know if you want to see it as an alternative or not.  I've been using it in my classes for files all fall.

meredydd:
Given that so very many methods are conserved between `str` and `bytes`, I think it'll be easier to keep the JS representation the same. I don't know how much else you've done (eg by way of properly indexing astral `str`s), but everything in this change is a thing we desperately need in Anvil, and I'd like to ship this functionality fairly soon.

rixner:
I need to look at yours and mine a little bit more to see if there are things missing from one that the other has.  Also, I do not like that you put the bytes class in str.js.  It should be in bytes.js.

I'm not sure I agree that using the same underlying representation is all that advantageous...

meredydd:
> I do not like that you put the bytes class in str.js. It should be in bytes.js.
>
> I'm not sure I agree that using the same underlying representation is all that advantageous...

Both of these choices stem from the fact that `bytes` still has an enormous number of functions you or I would think of as string-manipulation functions. These include `l`/`rstrip()`, `l`/`rjust()`, `center()` `partition()`, `count()`, `find`, `index`, `replace`, etc.

Given the choice between constructing an entirely fresh implementation of these functions for `bytes` (operating on `Uint8Array`s), and re-using the ones we already have, I went with the latter. The former would be both a lot of work and increase Skulpt's code size even further.

I'm not philosophically opposed to having the `bytes` and `str` _classes_ defined in different files, but (absent a stonking rewrite) they're going to share the vast majority (by numbers and LoC) of their method implementations, which is why I put them together.

meredydd:
(Apologies for all the force-pushing; I'm just rebasing on top of #975)

meredydd:
@rixner, do we have an ETA on being able to look at your alternative implementation and review this one?

meredydd:
Rebasing again, because #986 landed and all builtins that are conditional on 2-vs-3 cause conflicts in `Sk.setupObjects`

rixner:
Sorry for my slowness.  I continue to be swamped.

You can now look at #989.  I know they have similar methods.  However, I still believe they should be in independent files.  Now that we are using webpack, one could refactor code and "require" it in multiple places, if that makes the separation more palatable...

I do apologize for not having the time to look at this pull request carefully.  We are in the middle of finals week and that is sucking up my time.

meredydd:
Superseded by #1162. Boo-yeah!

PR #984, opened 2019-12-10
Refactor string formatting

meredydd:
This is a follow-up to #975, which refactors the string formatting code so that it can be used by f-strings (and the built-in `format()` function too).

In other words, formatting like `f"{4+5:.2f}"` now works :-D

I took the opportunity to do a bunch of cleanup and bugfixing in our `str.format()` implementation, and re-enable/add a couple more tests to `test_strformat.py`, as well as enabling all the format-string-related tests in `test_fstring.py`.

(All the changes for this PR are in the most recent commit - everything else GitHub is showing us is part of #975.)

### DIFF
--- a/repl/repl.js
+++ b/repl/repl.js
@@ -41,7 +41,11 @@ var compilableLines = [],
     lines = [],
     origLines;
 
-console.log("Python 2.6(ish) (skulpt, " + new Date() + ")");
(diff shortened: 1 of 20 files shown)


meredydd:
Rebased onto #975 again.

PR #986, opened 2019-12-12
Small Python 3 fixes

meredydd:
A grab-bag of small fixes to improve our Python 3 fidelity.

### DIFF
--- a/src/list.js
+++ b/src/list.js
@@ -171,6 +171,9 @@ Sk.builtin.list.prototype.tp$richcompare = function (w, op) {
             return true;
         }
 
+        if (Sk.__future__.python3) {
+            return Sk.builtin.NotImplemented.NotImplemented$;
+        }
         // todo; other types should have an arbitrary order
         return false;
     }
--- a/src/ast.js
+++ b/src/ast.js
@@ -408,6 +408,10 @@ function astForExceptClause (c, exc, body) {
         return new Sk.astnodes.ExceptHandler(ast_for_expr(c, CHILD(exc, 1)), null, astForSuite(c, body), exc.lineno, exc.col_offset);
     }
     else if (NCH(exc) === 4) {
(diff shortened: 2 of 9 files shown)


src/list.js
```diff
@@ -171,6 +171,9 @@ Sk.builtin.list.prototype.tp$richcompare = function (w, op) {
             return true;
         }
 
+        if (Sk.__future__.python3) {
+            return Sk.builtin.NotImplemented.NotImplemented$;
```
albertjan: It is strange that we're returning these here? Instead of raising them?

meredydd: Yes, it is, but that's what the caller is expecting! I'm guessing we return a sentinel value because raising exceptions is expensive, and on this code path you can end up trying a lot of handlers before one of them works.

This is a `tp$` function -- ie it's in JS-land -- so Skulpt can make up whatever protocols it likes. (For example, `tp$getattr()` returns `undefined` rather than throwing an `AttributeError`, for similar reasons.)

meredydd:
It says "small" right in the name. Merging.

alexwenbj:
@meredydd You mean in Python 3, `list` data can't compare to `dict` data and so on?But when I run this code below,I can get the result `False`:
```
import sys
a = [1,2]
b = {1:1,2:2}
print(sys.version)
print(a == b)
```
Result:
```
3.6.3 |Anaconda, Inc.| (default, Oct  6 2017, 12:04:38) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
False
```

meredydd:
@alexwenbj -- What this means is that you can't use `<` or `>` to compare `list` and `dict` objects. Equality still works, but whereas Python 2 thinks this code is valid:

```python
{'1':2} < [1,2,3]
```

Python 3 raises a `TypeError`, and now so does Skulpt.

alexwenbj:
@meredydd Got it.Thank you.

PR #989, opened 2019-12-13
Bytes Class

rixner:
This pull request is an implementation of the bytes class using Uint8Array and TextEncoder/Decoder.  This should be looked at alongside #983.  I believe it is only a subset of what is in #983, but I haven't been able to look as carefully as I would like.


alexwenbj:
@rixner @albertjan and other maintainers, how about this PR progress? I think this can make up the loss of bytes supports, and it is a base class for Python.I am look forward to this can be merge if it is OK soon.Thank you all.

bnmnetp:
@rixner -- would love to see this merged.  I don't feel qualified to resolve the conflicts though.

rixner:
There are now three pull requests on bytes.  I am happy to update this branch if that's where we want to go.  I think we, as a project, simply need to decide if we want #983, #989, or #1107.

Personally, I'd like to see the backlog of pull requests cleared, and bytes is one of the things that should be in, given that we have multiple decent implementations sitting around.

As I said in #983, I feel that bytes should be in bytes.js for consistency across the project. No one other than @meredydd and I weighed in on that, which is partially where I think the deadlock came from...

s-cork:
Does this pr necessarily require text-encoding? I ask because it seems that we might be creating an unnecessary large dist file here? When I ran the dist it seemed much larger than master - 1mb vs 500kb. But I don’t know if that’s accurate or if I’m mistaken. If that is accurate - maybe it’s possible to adapt the approach and use javascripts builtin TextEncoder and the appropriate polyfill?

I don’t have strong opinions about the internal representation of a bytes object - but it seems that this question should also be part of the decision - Should the internal representation be a string or an array like object.

If the v property is a string then methods between bytes and str can and should be shared... whether this means that they should live in the same file 🤷‍♂️

rixner:
It was my understanding (at the time) that TextEncoder is a polyfill.  In the browser, I believe it uses the builtin TextEncoder if it's there.  I saw no reason to re-implement functionality that exists in modern browsers.  But, that is quite a bit of bloat.  If there is a more efficient polyfill, I believe it would work just fine.

s-cork:
Ah yes - I see that was linked at the top of [MDN](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder) page as a polyfill 
the other polyfill on the same page: https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder#Polyfill
is taken from https://github.com/anonyco/FastestSmallestTextEncoderDecoder which seems to do necessary without the bulk.

rixner:
I tried the "fastest and smallest text encoder decoder".  It appears to only work with 'utf-8' encodings. Python allows a much richer set of encodings than this polyfill (https://docs.python.org/3/library/codecs.html#standard-encodings).  I guess the question is what we want to support in Skulpt.

rixner:
I believe the spec says it should support more than just 'utf-8' (https://encoding.spec.whatwg.org/#concept-encoding-get), which is more compatible with Python.  By using a restricted polyfill and/or by rolling our own, we necessarily restrict the supported encodings.

s-cork:
How likely are text encodings other than utf8 for skulpt users?


Am I right in thinking this pr covers utf8 and ascii only? Was the other polyfill insufficient for this pr?

rixner:
This PR only implements/tests utf8 and ascii. But, I think it could easily be extended (as I think the polyfill we chose supports more).   Switching to the smaller polyfill only allows utf8.  I have no idea how likely it is that people will want other encodings.

s-cork:
i've had more of a look at the `text-encoding` polyfill. I don't think it's appropriate for this pr.
- it adds approx `470kb` to `skulpt.min.js`. 

The majority of the bulk is from the non-standard encodings - which this pr does not utilise. 

The ascii encoding in this pr is defaulting to `utf-8` encoding. 

```javascript
x = new textEncoding.TextEncoder('ascii')
x.encoding // 'utf-8'
```
In order to utilise the ascii encoding it would have to be written [with legacy encodings set to true](https://github.com/inexorabletash/text-encoding#non-standard-behavior)
```javascript
x = new textEncoding.TextEncoder('ascii', { NONSTANDARD_allowLegacyEncoding: true })
x.encoding // windows-1252
```

So I think the smaller polyfill will do just fine for this pr by simply removing the `'ascii'` argument from `new TextEncoder`.
It has the benefit of being significantly smaller (adding just `2kb` to `skulpt.min.js`) and using the browser `TextEncoder` if it is available.

---

Other potential issues: 
- `text-encoding` is designed to be used with script tags and not with a package manager - the readme states
  _Use through [npm] is not really supported_
- the polyfill doesn't seem to use the browser `TextEncoding` if it is available (when used with a package manager)
  `textEncoding.TextEncoder === TextEncoder // false`
- according to [npmjs](https://www.npmjs.com/package/text-encoding) `text-encoding` is no longer maintained.

rixner:
I updated to fastestsmallesttextencoderdecoder and got everything working.  It passes all the tests for me locally.  I'm not sure why it's failing on travis.  I need to investigate further.

rixner:
The "problem" is that I'm using node v12 and Travis is using v10.  I've got v10 installed now, so can debug.

rixner:
This now works in node 10, node 12, and the browser for me.

rixner:
@skulpt/committers there are two bytes PRs and an AST PR from December 2019 that have been languishing.  I personally would like to see both bytes and ast in Skulpt.  However, @s-cork 's inhuman ability to continuously create useful PRs (keep it up!) makes it so that these old PRs are constantly out of date.  I would like to suggest the following:

We refrain from merging any more new PRs (now matter how tempting they are) until we act on bytes and ast.  For bytes, this means once and for all selecting either #983 or #989 and rejecting the other (or rejecting both).  For ast, it means evaluating and merging #991 (or just declaring we don't want it).

If we keep making Skulpt a moving target and ignoring these pull requests, we are passively saying we don't want them in Skulpt.  And as I said, I do want them in.  If I am the outlier and everyone else is okay with them not being in, that so be it.  But, please let's actively make that decision and close the pull requests, then.

Thanks.

rixner:
In light of my previous comment, I'd like to once again argue for this PR over #983.  I think that it is much better to store things in a byte array than a string.  While we know Chrome optimizes strings of ASCII characters, we don't know what other browsers do, and there's no reason to think that will necessarily always be the case.  Uint8Arrays are meant for this purpose and are a more direct match to the Python type.

I also think it makes sense to make sure we incorporate all of the tests from both PRs, regardless of which PR we actually merge.  Although, I would argue that should be a second step, given that these already have been sitting around for 8 months.  Let's get one of them in and then we can improve it.

s-cork:
I would also suggest that a more complete set of [cpython tests](https://github.com/python/cpython/blob/master/Lib/test/test_bytes.py) be incorporated and tests that don't pass should be commented out in the usual way...


This seems to have become somewhat of a stalemate. ♟️ 


I have, perhaps a naive question, one of the benefits of #983 seems to be the shared methods. How feasible (in a future pr) would it be to share methods with the approach here (#989) do you think? (and is there utility in doing so?) 

e.g.
```javascript
// roughly something like
str_lib.isdigit = function(self) {
    Sk.buitin.pyCheckArgsLen(...)
    return new Sk.builtin.bool( /^\d+$/.test(self.$jsstr()));
})

Sk.builtin.bytes.prototype.isdigit = new Sk.builtin.func(str_lib.isdigit)

// bytes version of $jsstr
Sk.builtin.bytes.prototype.$jsstr = function () {
    return new TextDecoder().decode(this.v.buffer);
}
```

rixner:
@s-cork perhaps another naive statement, but bytes and strings are not the same thing. #983 jumps through a lot of hoops to implement that sharing.  Operations on bytes are operating on bytes, not characters.  So, I don't think naively converting to a string is going to work in all situations (I think that's verified by some of the things that need to be done in #983 ).  Do I think some things could be pulled out and used in common?  Yes.  Do I think it's a good idea to just blindly do it for everything like #983?  No.  While code replication is generally bad, I think the problem here is the different possible encodings.  The method in this PR is encoding agnostic.  Bytes are treated as raw bytes and the operations all just work on raw bytes.  I believe #983 would require pervasive changes each time support for a new encoding is added, whereas here, just the encoding and decoding logic would need to be updated (which also would need to be updated in #983 ).  I could be wrong about that, but that's my impression.  Am I wrong?  I think it is a bad call to say that we are only ever going to want to support utf-8.  While that's the most important first step, I think we should be thinking about the fact that people might want other encodings in the future.

I feel like the "stalemate" is easily resolved.  The @skulpt/committers need to take a position on which approach they prefer. We can then get the preferred approach merged and work together to improve it.  I don't speak for @meredydd , but I assume that we are both mature enough to be okay with either PR being merged. What I am not okay with is letting them continue to languish and repeatedly get out of date by inaction.  If everyone hates both approaches, so be it, but they should speak up so we can just close them and someone else can take a crack at it.  I personally think that's counterproductive, though, given that once one of them is in, anyone can work to improve them.

I'm fine with continuing to work on this to satisfy people, but I would like to know I'm not wasting my time and that there's an end in sight.  So, the question is, what do I need to do to get two people from @skulpt/committers to approve/merge this?

meredydd:
Hi all,

I'm sorry to have been absent from this discussion -- Anvil has been keeping me busy!

Two things:

1. After some thought, and talking to @s-cork, I am willing to be convinced that `Uint8Array` is a better backing store for `bytes`. I'm still a little concerned about the code-size implications of duplicating so many of the methods that `str` and `bytes` have in common (once using strings and once using a `Uint8Array`), but I can trade off something I'm concerned about (code size) in favour of correctness/cleanliness, just as @rixner has been willing to trade off performance for suspension-correctness in the past :)

2. I also agree with @rixner that we should make sure that the eventual PR we accept should pass all the tests from both PRs. In particular, this will require supporting astral code points in `str`, and supporting all of the methods `str` and `bytes` have in common. This is a practical consideration -- we have had customers in production using #983 for months, and we can't impose regressions on them.

@s-cork has expressed interest in bringing over those features from #983 and building them on top of this PR, for which I _definitely_ owe him several drink. I agree with @rixner that having these two duelling PRs waiting for an uncomfortable conflict resolution has been holding us back for months, and I am excited that Skulpt is finally about to get a good, solid, Python-3-ic implementation of both `bytes` and `str`!

src/bytes.js
```diff
@@ -0,0 +1,1522 @@
...
+        if (val1.v != val2.v) {
+            return Sk.builtin.bool.false$;
+        }
+    }
+
+    return Sk.builtin.bool.true$;
+};
+
+Sk.builtin.bytes.prototype.ob$ne = function (other) {
+    return (!(this.ob$eq(other)));
```
s-cork, line 252: this will always return false since `ob$eq` will return a `pyBool` I would suggest making `ob$eq` return js `boolean`s

Would also be good to implement the other comparisons
https://github.com/python/cpython/blob/46e19b61d31ba99f049258efa4ff1334856a3643/Lib/test/test_bytes.py#L223

meredydd: Changing the semantics of `ob$eq` is probably out of scope for this PR - we should make `bytes`'s implementation match the rest of the codebase for now!

rixner:
Okay, I believe that means we have @s-cork , @meredydd , and I agreeing on this PR over #983?  Given that, I am happy to invest the time to do what's needed to get this accepted.  The question I have, though, is what should be in this PR and what should be in future PRs?  Should I "steal" @meredydd 's test cases (and add in whatever is missing from CPython, with failing tests commented out) and make them all pass?  Should I just address the explicit comments on the code here and then @meredydd , @s-cork , and/or I can create additional PRs pulling in missing features from #983 and whatever other improvements?  Something else?  As I've said, if there's an end game, I'm willing to do what's needed.  But, the sooner we get this in, the sooner people can productively build upon it.

@meredydd I totally understand and agree with not introducing regressions for either of us.  I've been using this for what seems like forever, as well.  The question is whether that needs to be accomplished in a single PR, or if it can be spread out over multiple PRs.

src/bytes.js
```diff
@@ -0,0 +1,1522 @@
...
+                            data.push(val);
+                        }
+                    }
+                    ret = Uint8Array.from(data);
+                } else if (encoding === "utf-8") {
+                    ret = new TextEncoder().encode(source);
+                } else {
+                    throw new Sk.builtin.LookupError("unknown encoding: " + encoding.v);
+                }
+                view = new DataView(uint8ToBuffer(ret));
```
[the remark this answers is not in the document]
rixner: It was so long ago that I honestly don't remember.  I was able to convert it to just use Uint8Array directly and I agree that's better.

s-cork:
I'd say... address the comments in this pr.

I'm then happy to merge #983 onto this branch
(I took the afternoon and have a working version with combined tests. I can then rebase it on top of this branch with whatever changes result from the review here)

- get this merged
- a separate pr to critique incorporating #983 
And hopefully check and mate...

src/bytes.js
s-cork: oops sorry yes - stick with the `pyBool` return values.

meredydd:
I'm a _little_ nervous about merging this PR without #983's capabilities (particularly astral codepoints and the ability to parse Unicode string literals), because from the moment this PR is merged until we backfill those features, Anvil is unable to use Skulpt master, which is never a fun place to be!

But if @s-cork is going to fill in the rest Really Soon Thereafter, then let's hold our breath and push through it:)

rixner:
@meredydd while I feel your pain, I haven't been able to use Skulpt master in years.  Every time I try to get back in line, something changes...  I'll address all the comments and we can go from there.  Perhaps @s-cork adjusts his and we merge this only when he says he's ready?  That way the two pull requests can be nearly back-to-back?

meredydd:
Yes, that sounds like a good plan! @s-cork says his is already nearly ready, so I'm good for merge as soon as the current review comments have been addressed :)

src/bytes.js
```diff
@@ -0,0 +1,1522 @@
...
+    var i;
+    for (it = this.tp$iter(), i = it.tp$iternext(); i !== undefined; i = it.tp$iternext()) {
+        if (Sk.misceval.richCompareBool(i, item, "Eq")) {
+            return true;
+        }
+    }
+    return false;
+};
+
+Sk.builtin.bytes.prototype.$decode = function (self, encoding, errors) {
```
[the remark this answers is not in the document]
rixner: It's also used by the compiler...

rixner:
I believe I have addressed all of the comments.  I did not add commented missing CPython tests (beyond the few directly relevant to the comments).  @s-cork did you already do that in your PR, or should I do it here?

s-cork:
the changes look good to me. I haven't added any cpython tests - can if you like or if you do then 👍 . 

I've basically just done a merge #983 onto this branch, removed references to `bytes.prototype.foo =` in the #983 str.js file and then adapted the conflicting approaches to compile.js/ast.js...
I will need to do a little more scouring of the `str.js` to do a deeper bytes cleanse... but I do have all the tests working which was the initial aim... 

i'll need to just rebase now. As i was reviewing, I was playing with the code in this branch here and there to understand it better so will need to just revert anything not appropriate to the PR.

Will then put two pull requests on my fork (one for this branch and one for #983) to more easily see the diffs between the two bytes prs. Along with a pr to master I guess.

src/bytes.js
```diff
@@ -0,0 +1,1443 @@
...
+/**
+ * @constructor
+ * @param {*} source
+ * @param {Sk.builtin.str=} encoding
+ * @param {Sk.builtin.str=} errors
+ * @return {Sk.builtin.bytes}
+ */
+
+Sk.builtin.bytes = function (source, encoding, errors) {
+    let view;
```
s-cork, line 30: can we remove view now? I don't think it's referenced in this function anymore

src/bytes.js
```diff
@@ -0,0 +1,1443 @@
...
+    if (arguments.length === 1) {
+        if (Sk.builtin.checkInt(source)) {
+            source = Sk.builtin.asnum$(source);
+            arr = new Uint8Array(source);
+        } else if (source instanceof Sk.builtin.bytes) {
+            return source;
+        } else if (Array.isArray(source)) {
+            // Internal fast path
+            Sk.asserts.assert(source.every((x) => (x >= 0) && (x < 256)),
+                              "Bad internal call to bytes with array object");
```
s-cork, line 63: what did you think about this btw? A worthwhile sanity check or a check that has zero probability of failing?

src/bytes.js
```diff
@@ -0,0 +1,1522 @@
...
+                    break;
+                case 92:
+                    ret += "\\\\";
+                    break;
+            }
+        } else {
+            ret += String.fromCharCode(num);
+        }
+    }
+    ret = "b'" + ret + "'";
```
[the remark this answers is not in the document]
s-cork: I'm happy to address this one - I think i fixed it in a version I was working on.

src/bytes.js
```diff
@@ -0,0 +1,1443 @@
...
+            if (i < 0 || i >= this.v.byteLength) {
+                throw new Sk.builtin.IndexError("index out of range");
+            }
+            return new Sk.builtin.int_(this.v[i]);
+        }
+    } else if (index instanceof Sk.builtin.slice) {
+        ret = [];
+        index.sssiter$(this.v.byteLength, (i) => {
+            ret.push(this.v[i]);
+        });
```
s-cork, line 208: love this

src/bytes.js
```diff
@@ -0,0 +1,1443 @@
...
+        if (this.v[i] != other.v[i]) {
+            return Sk.builtin.bool.false$;
+        }
+    }
+
+    return Sk.builtin.bool.true$;
+};
+
+Sk.builtin.bytes.prototype.ob$ne = function (other) {
+    return Sk.misceval.isTrue(this.ob$eq(other)) ?
```
s-cork, line 234: I think we also need to do a `instantceof => NotImplemented$` type pre-check here.

src/bytes.js
```diff
@@ -0,0 +1,1522 @@
...
+
+    }
+    index = new Sk.builtin.slice(0, val);
+    final1 = self.mp$subscript(index);
+    index = new Sk.builtin.slice(val, val + sep.v.byteLength);
+    final2 = self.mp$subscript(index);
+    index = new Sk.builtin.slice(val + sep.v.byteLength, self.v.byteLength);
+    final3 = self.mp$subscript(index);
+
+    return new Sk.builtin.tuple([final1, final2, final3]);
```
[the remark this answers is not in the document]
s-cork: same approach as per changes in `partition`?

src/bytes.js
```diff
@@ -0,0 +1,1443 @@
...
+                    if (item >= 0 && item <= 256) {
+                        final.push(item);
+                    } else {
+                        throw new Sk.builtin.ValueError("bytes must be in range(0, 256)");
+                    }
+                } else {
+                    throw new Sk.builtin.TypeError("'" + Sk.abstr.typeName(item) + "' " + "object cannot be interpreted as an integer");
+                }
+            }
+            arr = new Uint8Array(final);
```
s-cork, line 84: how careful do we need to be here
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/Uint8Array#Browser_compatibility
`new Uint8Array(iterable)` support

src/bytes.js
```diff
@@ -0,0 +1,1443 @@
...
+                                const hexval = makehexform(val);
+                                throw new Sk.builtin.UnicodeEncodeError("'ascii' codec can't encode character '" + hexval + "' in position " + i + ": ordinal not in range(128)");
+                            } else if (errors === "replace") {
+                                data.push(63); // "?"
+                            }
+                        } else {
+                            data.push(val);
+                        }
+                    }
+                    arr = Uint8Array.from(data);
```
s-cork, line 114: same question 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/from#Browser_compatibility

src/bytes.js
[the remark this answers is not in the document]
rixner: I "fixed" it, in the sense that there will not be an error now.  (It will do b'\''.)  If you have a fix that converts that to b"'", great.  I didn't think that was worth it, but I guess it's actually not that much more difficult.

src/bytes.js
rixner: Asserts are no-ops in production, so I don't think it hurts anything to have it here and can only help.

src/bytes.js
rixner: Good point.

src/bytes.js
rixner: I wrote a really simple polyfill for Uint8Array.from and I use "from" everywhere an array is converted to a Uint8Array now.  The polyfill is good enough for the uses in the file, but I'm not sure if people are comfortable with that or not.   I don't know if we want a full Uint8Array polyfill, though a cursory look didn't turn up one that I was happy with.

rixner: The more I think about it, the less okay I am with a sketchy polyfill. I'll look harder for a better one, and if there isn't one, then perhaps just write a function that does what we want with a clear comment about why it doesn't use the from method.

rixner: The more I think about it, the less okay I am with a sketchy polyfill. I'll look harder for a better one, and if there isn't one, then perhaps just write a function that does what we want with a clear comment about why it doesn't use the from method.

rixner: The more I think about it, the less okay I am with a sketchy polyfill. I'll look harder for a better one, and if there isn't one, then perhaps just write a function that does what we want with a clear comment about why it doesn't use the from method.

rixner: The more I think about it, the less okay I am with a sketchy polyfill. I'll look harder for a better one, and if there isn't one, then perhaps just write a function that does what we want with a clear comment about why it doesn't use the from method.

rixner: The more I think about it, the less okay I am with a sketchy polyfill. I'll look harder for a better one, and if there isn't one, then perhaps just write a function that does what we want with a clear comment about why it doesn't use the from method.

rixner: The more I think about it, the less okay I am with a sketchy polyfill. I'll look harder for a better one, and if there isn't one, then perhaps just write a function that does what we want with a clear comment about why it doesn't use the from method.

s-cork:
Excellent - I think i'm there with the unicode support - and I've added a handful of cpython tests commenting out a few that were out of scope. When you think this pr is ready i'll do a separate pr for that branch.

rixner:
@s-cork I think it's ready.  We can always do smaller cleanup PRs if anything comes up.  I will stop modifying it, so you can get your PR ready.  That will allow us to review yours, then we can merge them both.

s-cork:
Here's the diff. 

https://github.com/s-cork/skulpt/pull/7

Most of the changes outside of bytes are the merge of #983 to support Unicode and astralcodepoints in ast/compile/str. I adjusted the approach from #983 in compile.js to account for the Uint8Array representation of bytes. And fixed an error for compiling non ascii byte literals (0xff => 0x7f)

With str.js I really didn't do much apart from rebase #983 on top of master and cleanse references to bytes. Any changes to that file are from the #983 branch except a couple related to making the two prs work together. 

The main change in bytes is to the constructor and distinguishes between an internal call and a py call. Support added for `__bytes__` which can suspend and iterators that can suspend. The diff for the constructor isn't so easy to read but looking at the file on its own should hopefully clear up what I was aiming to do. 

Will make a proper pr tomorrow/Monday. Rebasing any further tweaks here.

rixner:
@s-cork, I will look at this more carefully later, but after a quick read, I have two questions/comments:

1) Why did you eliminate the bytes object in the ast code?  A bytes literal should be a bytes object, not a string object.

2) You reintroduced Uint8Array.from, which you pointed out should go.  I am assuming this is because you didn't grab my last few commits?

As you said, I am confused by the diff on the constructor, but I'll look it over directly, as you suggested.  Ultimately, I feel like there are far more changes in bytes.js than I would have expected, since as @meredydd pointed out, his PR is really about unicode handling, which seems like it should mostly, if not entirely, be confined to strings, and not really impact the bytes implementation.

s-cork:
Yes just needed to adjust uint8array.from. On my Todo. 

Good point. 

I wasn't totally sure what to do I guess with the ast module here to marry the branches. The conflict between branches was in ast and handling the bytes str literal and the parsestr algorithm. decode-utf8 was removed in #983 and I didn't want to change the algorithm in #983. The compromise i went for was that ast now handled the byte literal as a str object but uses the case Ast.Bytes... compile then makes it a bytes object. It meant that `b'\xf0\x9f\x8d\x95'` could be compiled... happy to to move in a different direction here. It just seemed the easiest way to marry the approaches.

rixner:
I guess it only matters in how you make the constant.  Looking more closely, I see that you create an array of integers for the constant, which makes sense.  I understand your approach now, and it seems reasonable.  I'll look at the rest more carefully.

src/bytes.js
```diff
@@ -0,0 +1,1458 @@
...
+
+Sk.builtin.bytes.prototype.bytes_copy_ = function () {
+    var i;
+    var final;
+    final = [];
+    for (i = 0; i < this.v.byteLength; i++) {
+        final.push(this.v[i]);
+    }
+    return new Sk.builtin.bytes(final);
+};
```
s-cork, line 268: do we still need this function? I couldn't see that it gets called.

rixner: It's no longer used.  Removed.

src/bytes.js
```diff
@@ -0,0 +1,1455 @@
...
+        return new Sk.builtin.int_(-1);
+    } else {
+        throw new Sk.builtin.TypeError("argument should be integer or bytes-like object, not '" + Sk.abstr.typeName(sub) + "'");
+    }
+};
+
+Sk.builtin.bytes.prototype["find"] = new Sk.builtin.func(function (self, sub, start, end) {
+    Sk.builtin.pyCheckArgsLen("find", arguments.length - 1, 1, 3);
+
+    return Sk.builtin.bytes.prototype.find_item_.call(self, sub, start, end);
```
s-cork, line 661: suggestion
change this to `Sk.builtin.bytes.prototype.find$item`
and instead of doing `Sk.builtin.bytes.prototype.find_item_.call` do:
```javascript
self.find$item(sub, start, end)
```
or use it as a standalone function in this module rather than on the `prototype` of bytes.

(i only suggest `$` because in theory a trailing underscore could be overridden by a subclass)

rixner:
Okay, I think I'm done messing with this branch and it is ready to be merged, as far as I'm concerned, once #1162 is good to go right after it.  @s-cork, @meredydd let me know if you think there's anything further that I should do here.  @s-cork, I'm totally fine if you want to just add things to #1162 instead of waiting for me if you find anything else.

meredydd:
Merged! :tada:

PR #991, opened 2019-12-18
AST module

acbart:
With my semester wrapping up, I wanted to make some pull requests for features I added during the semester. This one is probably not the most important, but its a big part of BlockPy's toolchain, so I wanted to get a sense of your interest. Others include `exec`, more info in exceptions, and a fix to special method lookup in classes.

This PR adds in an importable `ast` module that has the most of the core functionality you'd expect from CPython's `ast` library, including AST node classes, a `NodeVisitor` class, and a `parse` function. It also expands nodes to include ending lines and columns in addition to the starting attributes.

If you're interested, I'm willing to spend some time cleaning up some of the code further - I am passing all the test cases I wrote, at least, and none of the existing tests seem to be broken. Not sure how compatible this is with the other pull requests open right now. Probably should merge them in first and then make further modifications as required.

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -402,10 +402,10 @@ function astForExceptClause (c, exc, body) {
     REQ(exc, SYM.except_clause);
     REQ(body, SYM.suite);
     if (NCH(exc) === 1) {
-        return new Sk.astnodes.ExceptHandler(null, null, astForSuite(c, body), exc.lineno, exc.col_offset);
+        return new Sk.astnodes.ExceptHandler(null, null, astForSuite(c, body), exc.lineno, exc.col_offset, exc.end_lineno, exc.end_col_offset);
     }
     else if (NCH(exc) === 2) {
-        return new Sk.astnodes.ExceptHandler(ast_for_expr(c, CHILD(exc, 1)), null, astForSuite(c, body), exc.lineno, exc.col_offset);
+        return new Sk.astnodes.ExceptHandler(ast_for_expr(c, CHILD(exc, 1)), null, astForSuite(c, body), exc.lineno, exc.col_offset, exc.end_lineno, exc.end_col_offset);
     }
     else if (NCH(exc) === 4) {
         if (Sk.__future__.python3 && CHILD(exc, 2).value == ",") {
@@ -415,7 +415,7 @@ function astForExceptClause (c, exc, body) {
         var expression = ast_for_expr(c, CHILD(exc, 1));
         e = ast_for_expr(c, CHILD(exc, 3));
         setContext(c, e, Sk.astnodes.Store, CHILD(exc, 3));
-        return new Sk.astnodes.ExceptHandler(ast_for_expr(c, CHILD(exc, 1)), e, astForSuite(c, body), exc.lineno, exc.col_offset);
+        return new Sk.astnodes.ExceptHandler(ast_for_expr(c, CHILD(exc, 1)), e, astForSuite(c, body), exc.lineno, exc.col_offset, exc.end_lineno, exc.end_col_offset);
     }
     Sk.asserts.fail("wrong number of children for except clause");
 }
@@ -463,7 +463,7 @@ function astForTryStmt (c, n) {
     }
 
     Sk.asserts.assert(!!finally_ || handlers.length != 0);
-    return new Sk.astnodes.Try(body, handlers, orelse, finally_, n.lineno, n.col_offset);
+    return new Sk.astnodes.Try(body, handlers, orelse, finally_, n.lineno, n.col_offset, n.end_lineno, n.end_col_offset);
 }
 
 function astForDottedName (c, n) {
@@ -476,10 +476,10 @@ function astForDottedName (c, n) {
     lineno = n.lineno;
     col_offset = n.col_offset;
     id = strobj(CHILD(n, 0).value);
-    e = new Sk.astnodes.Name(id, Sk.astnodes.Load, lineno, col_offset);
+    e = new Sk.astnodes.Name(id, Sk.astnodes.Load, lineno, col_offset, n.end_lineno, n.end_col_offset);
     for (i = 2; i < NCH(n); i += 2) {
         id = strobj(CHILD(n, i).value);
-        e = new Sk.astnodes.Attribute(e, id, Sk.astnodes.Load, lineno, col_offset);
+        e = new Sk.astnodes.Attribute(e, id, Sk.astnodes.Load, lineno, col_offset, n.end_lineno, n.end_col_offset);
     }
     return e;
 }
@@ -497,7 +497,7 @@ function astForDecorator (c, n) {
     }
     else if (NCH(n) === 5) // call with no args
     {
-        return new Sk.astnodes.Call(nameExpr, [], [], null, null, n.lineno, n.col_offset);
+        return new Sk.astnodes.Call(nameExpr, [], [], null, null, n.lineno, n.col_offset, n.end_lineno, n.end_col_offset);
     }
     else {
         return ast_for_call(c, CHILD(n, 3), nameExpr);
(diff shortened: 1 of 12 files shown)


acbart:
The parser also collects comments now too. That's a necessary bit for BlockMirror's block/text conversion algorithm, though it might be esoteric for others.

Oh, and I think some code for modules ended up in here, which I can move to the `exec` PR. That's related to the crazy global mangling that has to go on to make things `exec` keep namespaces separate.

rixner:
I am in favor of this pull request.  I looked over the code and it seems like this doesn't disrupt anything else in the core and it adds new functionality we don't currently have.

However, I can't run it right now, perhaps because I resolved the conflicts incorrectly (I didn't try very hard, admittedly)?  I'm also not sure why Travis is saying it passes the tests when there is a conflict?

Also, you changed ```Sk.builtin.module``` to take a module name as an argument, but you didn't change existing callers (I think there's only one in import.js) to properly pass the name.

Anyway, @acbart  if you can resolve the conflicts and get things in a condition you are happy with, I will be happy to approve this!

acbart:
Sorry I've been utterly unable to work on this lately! Our semester started a few weeks ago, which means my long winter break ended without the chance for me to contribute anything more while I was waiting for this to be approved. It seems like you've been able to incorporate some of my changes on your own, anyway. When the summer rolls around, hopefully I can get the rest of my changes integrated into wherever you end up.

rixner:
I would still very much like to see this merged. @acbart do you have time to bring it up to date?

acbart:
I plumb forgot about this, but I do have time now. When integrating @s-cork 's #1092 pull request I did have to make some changes to this module.  Is it worth waiting till that PR is merged in before trying to get this merged in? Or vice versa?

rixner:
I'd vote to clear the backlog of pull requests before we even consider #1092.  So, my vote is this should go first.

albertjan:
@acbart can you do this module without adding `end_line_no` and `end_col_offset` change?

rixner:
@albertjan is there a reason you don't want those changes?  I think we want those.  It's one of the things I needed to add for #1044.

albertjan:
Sure but I would rather see them as a separate PR. As they aren’t strictly necessary for the ast module. And it makes this PR harder to merge.

Get Outlook for iOS<https://aka.ms/o0ukef>

acbart:
I appreciate wanting to have them be separate, but that would be an awful lot of work on changing something I've already done. Plus, I'd really like to have the line endings in upstream Skulpt, it's currently making it much harder for me to merge in upstream changes. I have a little time today to make the already requested changes and resolve the simple merge conflict that was there, but if you want that change, it'll be a lot longer before I'd have the time needed.

acbart:
Okay, want to take another look? I'm passing all test cases and I don't think there are any conflicts to resolve.

src/ast.js
```diff
@@ -3363,5 +3378,115 @@ Sk.astDump = function (node) {
     return _format(node, "");
 };
 
+
+Sk.INHERITANCE_MAP = {
```
albertjan, line 3382: what does this do is this automatically generated?

src/parser.js
```diff
@@ -68,6 +69,11 @@ function findInDfa (a, obj) {
     return false;
 }
 
+// Add a comment
```
albertjan, line 72: our parser throws out comments? Does the cpython parser leave them in?

albertjan:
Can you see how many of these tests pass? https://github.com/python/cpython/blob/3.8/Lib/test/test_ast.py

src/parser.js
acbart: If I recall, both Skulpt and cpython toss the comments out.

src/ast.js
acbart: Not automatically generated. I believe this data ended up here largely out of convenience that came when authoring it, and I'm thinking now it needs to be moved into the `src/lib/ast.js` file instead. The idea is that this encodes the inheritance relationships that exist between nodes. Want it moved to the library AST file?

acbart:
To answer the question of how many tests pass: apparently almost none of them (I think 3 passed).

1. I haven't implemented any of the functionality for things like `fix_missing_locations`, `get_source_segment`, and `get_docstring`.
2. I added in a little functionality to create nodes manually (limited utility for my purposes), but that doesn't prepopulate their fields.
3. There's some wacky inheritance things that happen with the new `Constant` nodes that I don't think we should bother with until we have proper metaclassing. It's weird.
4. I believe I messed up the keyword arguments for `parse` and `dump`, but I don't know what the exact style is supposed to be. Is that documented somewhere?

Only #4 seems worth fixing to me, right now. I think someone else will have to feel strongly about 1-3, since I have no particular reason to make them happen.

albertjan:
Thanks @acbart. I'll try to have look tomorrow.

src/parser.js
albertjan: But you keep them in here? But I don't see where you use them.

src/module.js
```diff
@@ -1,7 +1,13 @@
...
  */
-Sk.builtin.module = function module () {
+Sk.builtin.module = function module (name) {
+    if (!(this instanceof Sk.builtin.module)) {
+        return new Sk.builtin.module(name);
+    }
+    this["$d"] = {__name__: name};
+    this["$d"]["__dict__"] = this["$d"];
+    return this;
 };
```
s-cork, line 11: Should we have the `__dict__` here? this would return a javascript object to the user no?

acbart: I totally defer to you on this one, I figured whatever I had in here would have to be heavily changed for your new PRs anyway. This is just what I had at this point in time. I note that my own local Skulpt (which integrates most of your changes, but not the latest ones) actually looks like this:

```javascript
Sk.builtin.module = Sk.abstr.buildNativeClass("module", {
// ...
        tp$init: function(args, kwargs) {
            Sk.abstr.checkArgsLen(this.tp$name, args, 1, 3);
            this["$d"] = {
                "__name__": args[0],
                "__package__": Sk.builtin.none.none$,
            };
            return Sk.builtin.none.none$;
        },
```

s-cork: Nice. Yeah, For this pr I'd remove the `__dict__` since the user won't get a python object if they try to access it. I guess this really should be a python dictionary in the future...

acbart: Does b92c363d246cd672250208165c4ad4a3ac9fbae4 handle this? I feel silly for letting my module stuff slip into this PR, it's not really all that relevant, ultimately.

test/unit3/test_ast.py
```diff
@@ -0,0 +1,1984 @@
...
+('Expression', ('Subscript', (1, 0), ('Name', (1, 0), 'a', ('Load',)), ('Slice', ('Name', (1, 2), 'b', ('Load',)), ('Name', (1, 4), 'c', ('Load',)), None), ('Load',))),
+('Expression', ('Name', (1, 0), 'v', ('Load',))),
+('Expression', ('List', (1, 0), [('Constant', (1, 1), 1, None), ('Constant', (1, 3), 2, None), ('Constant', (1, 5), 3, None)], ('Load',))),
+('Expression', ('List', (1, 0), [], ('Load',))),
+('Expression', ('Tuple', (1, 0), [('Constant', (1, 0), 1, None), ('Constant', (1, 2), 2, None), ('Constant', (1, 4), 3, None)], ('Load',))),
+('Expression', ('Tuple', (1, 0), [('Constant', (1, 1), 1, None), ('Constant', (1, 3), 2, None), ('Constant', (1, 5), 3, None)], ('Load',))),
+('Expression', ('Tuple', (1, 0), [], ('Load',))),
+('Expression', ('Call', (1, 0), ('Attribute', (1, 0), ('Attribute', (1, 0), ('Attribute', (1, 0), ('Name', (1, 0), 'a', ('Load',)), 'b', ('Load',)), 'c', ('Load',)), 'd', ('Load',)), [('Subscript', (1, 8), ('Attribute', (1, 8), ('Name', (1, 8), 'a', ('Load',)), 'b', ('Load',)), ('Slice', ('Constant', (1, 12), 1, None), ('Constant', (1, 14), 2, None), None), ('Load',))], [])),
+]
+main()
```
s-cork, line 1984: for this pr we'd need this unit test module to pass. I'd suggest commenting out all the tests that don't pass here...

test/test_ast.py
```diff
@@ -0,0 +1,94 @@
+import ast
```
s-cork, line 1: these tests don't fit with the style of skulpt. 
Can we turn these into unit tests?

src/parser.js
kolowy: Is there a better solution to parse comment ? I think this is not the optimal way

PR #994, opened 2019-12-26
chr() python3 mode awareness

cgibin:
Making possible to use chr() with unicode characters in python3 mode

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -500,6 +500,10 @@ Sk.builtin.ord = function ord (x) {
 };
 
 Sk.builtin.chr = function chr (x) {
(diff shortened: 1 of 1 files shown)


cgibin:
what about Sk.builtin.unichr.apply(this, arguments) to avoid double check?
Or this is some sort of code agreement?

meredydd:
Hi @cgibin,

Thanks for the pull request! This functionality is actually already available in PR #983, which makes the `str` class properly Unicode-aware. Can you try checking out that branch and seeing if it does what you need?

cgibin:
@meredydd sorry for delay, everything is fine in your branch, I'll wait till merge and closing this request

PR #998, opened 2020-01-06
Un-break string equality in Python 2

meredydd:
PR #986 broke string equality in Python 2 (doh). This PR adds tests to replicate and then fixes.

I'd like to fast-track this in; it's a small, tested fix to something previously accepted.


albertjan:
go ahead @meredydd

PR #1004, opened 2020-02-01
Pr/math.js

s-cork:
First pull request to open source project - feedback welcome...
_(sorry about the reordering - I wanted it to match the python math.rst - I was learning git at the same time but I now realise this will probably make it difficult to compare some changes to existing functions)_

This updates the math.js file with the following:
```
copysign    --> previously incorrectly implemented  **existing function**
                (was the wrong way round according to docs - tests now uncommented)
erf         --> NotImplementedError
erfc        --> NotImplementedError
exp         --> raises OverflowError tests uncommented **existing function**
expm1       --> Implemented with tests
factorial   --> adapted to use BigInt for x > 18 and ValueErrors raised  **existing function**
fmod        --> Implemented with tests
frexp       --> Implemented with tests
fsum        --> Implemented with tests
gamma       --> NotImplementedError
gcd         --> Implemented with tests
isclose     --> Implemented with tests
isinf       --> Corrected isinf(nan) is now False  **existing function**
ldexp       --> Implemented with tests
lgamma      --> NotImplementedError
log         --> works for large ints e.g. log(2**1024)  **existing function**
log10       --> works for large ints e.g. log10(2**1024)  **existing function**
log1p       --> Implemented with tests
log2        --> Implemented with tests
modf        --> Implemented with tests
nan         --> Implemented (respective tests uncommented)
inf         --> Implemented (respective tests uncommented)
pow         --> Tests uncommented for infs, and OverflowError  **existing function**
remainder   --> Implemented with tests
tanh        --> corrected for signed -0.0  **existing function**
tau         --> Implemented with test(s)
trunc       --> Corrected for large ints should return the int
                previously returned inf e.g. trunc(2**2048)  **existing function**
```

the change of copysign has resulted in failed tests in test/unit/test_complex.py
I think this is because test_complex uses math.copysign for comparison which was previously incorrectly implemented.

### DIFF
--- a/src/complex.js
+++ b/src/complex.js
@@ -161,10 +161,10 @@ Sk.builtin.complex = function (real, imag) {
         ci.real += cr.imag;
     }
 
-    // adjust for negated imaginary literal
-    if (cr.real === 0 && (ci.real < 0 || Sk.builtin.complex._isNegativeZero(ci.real))) {
-        cr.real = -0;
-    }
+    // // adjust for negated imaginary literal
+    // if (cr.real === 0 && (ci.real < 0 || Sk.builtin.complex._isNegativeZero(ci.real))) {
+    //     cr.real = -0;
+    // }
 
     // save them as properties
     this.real = new Sk.builtin.float_(cr.real);
@@ -823,8 +823,14 @@ Sk.builtin.complex.prototype.tp$setattr = function (name, value) {
  * This functions assumes, that v is always instance of Sk.builtin.complex
  */
 Sk.builtin.complex.complex_format = function (v, precision, format_code){
-    function copysign (a, b) {
-        return b < 0 ? -Math.abs(a) : Math.abs(a);
+    function copysign(a, b) {
+        let sign;
+        if (b) {
+            sign = b < 0 ? -1 : 1;
+        } else {
+            sign = 1 / b < 0 ? -1 : 1;
+        };
+        return sign * Math.abs(a);
     }
 
     if (v == null || !Sk.builtin.complex._complex_check(v)) {
--- a/src/complex.js
+++ b/src/complex.js
@@ -161,11 +161,6 @@ Sk.builtin.complex = function (real, imag) {
         ci.real += cr.imag;
     }
 
-    // // adjust for negated imaginary literal
-    // if (cr.real === 0 && (ci.real < 0 || Sk.builtin.complex._isNegativeZero(ci.real))) {
-    //     cr.real = -0;
-    // }
-
     // save them as properties
     this.real = new Sk.builtin.float_(cr.real);
     this.imag = new Sk.builtin.float_(ci.real);
--- a/test/unit3/test_math.py
+++ b/test/unit3/test_math.py
@@ -79,6 +79,7 @@ def ulp(x):
 # multiplying by all j in {n >> i+1 < j <= n >> i; j odd}.  In Python terms,
 # this set is range((n >> i+1) + 1 | 1, (n >> i) + 1 | 1, 2).
 
+
 def count_set_bits(n):
     """Number of '1' bits in binary expansion of a nonnnegative integer."""
     return 1 + count_set_bits(n & n - 1) if n else 0
@@ -187,7 +188,7 @@ class MathTests(unittest.TestCase):
         # Ref: Abramowitz & Stegun (Dover, 1965)
         self.ftest('pi', math.pi, 3.141592653589793238462643)
         self.ftest('e', math.e, 2.718281828459045235360287)
-        # self.assertEqual(math.tau, 2*math.pi)
+        self.assertEqual(math.tau, 2*math.pi)
 
     def testAcos(self):
         self.assertRaises(TypeError, math.acos)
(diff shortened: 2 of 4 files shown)


s-cork:
fixes #1005

s-cork:
fixes #1010

src/complex.js
```diff
@@ -161,10 +161,10 @@ Sk.builtin.complex = function (real, imag) {
...
     }
 
-    // adjust for negated imaginary literal
-    if (cr.real === 0 && (ci.real < 0 || Sk.builtin.complex._isNegativeZero(ci.real))) {
-        cr.real = -0;
-    }
+    // // adjust for negated imaginary literal
+    // if (cr.real === 0 && (ci.real < 0 || Sk.builtin.complex._isNegativeZero(ci.real))) {
+    //     cr.real = -0;
+    // }
```
s-cork: I commented this out. 
any time the second argument has a negative real part and the first argument is zero then the first argument becomes a negative zero. 

e.g.
```
>>> complex(0, -3)
-0-3i
```
the real 0 becomes a negative zero because the second argument real part (-3) is negative... 
but the first argument 0 should remain the same and not depend on the real part of the second argument at all. 

fixes #1010

src/complex.js
```diff
@@ -823,8 +823,14 @@ Sk.builtin.complex.prototype.tp$setattr = function (name, value) {
...
-    function copysign (a, b) {
-        return b < 0 ? -Math.abs(a) : Math.abs(a);
+    function copysign(a, b) {
+        let sign;
+        if (b) {
+            sign = b < 0 ? -1 : 1;
+        } else {
+            sign = 1 / b < 0 ? -1 : 1;
+        };
+        return sign * Math.abs(a);
```
s-cork: copysign should respect -0.0

src/complex.js
s-cork: fixes #1012

test/unit3/test_math.py
```diff
@@ -584,48 +584,51 @@ def testfrexp(name, result, expected):
...
+            # a little unsafe because str -> float conversion can't be
+            # relied upon to do correct rounding on all platforms.
+            tail = max(len(bin(abs(tmant)))-2 - mant_dig, etiny - texp)
+            if tail > 0:
+                h = 1 << (tail-1)
+                tmant = tmant // (2*h) + bool(tmant & h and tmant & 3*h-1)
+                texp += tail
+            
+            # weird hack that fixes a bug when tmant is converted to a float incorrectly by skulpt
+            return math.ldexp(int(str(tmant)), texp)
```
s-cork: I wasn't sure why I had to do this but I did?
`tmant` was a large negative `int` and doing `float(tmant)` vs `float(int(str(tmant))` gave wildly different values.
when testing this I had a value `tmat = -2492302492096719`
```
float(tmat)              --> -1689704655.0
float(-2492302492096719) --> -2492302492096719.0
float(str(int(tmat))     --> -2492302492096719.0
```

the adapted code which highlights this issue is... :
```python
import math
from math import fsum

mant_dig = 53
etiny = -1024 - 53


def msum(iterable):
    """Full precision summation.  Compute sum(iterable) without any
    intermediate accumulation of error.  Based on the 'lsum' function
    at http://code.activestate.com/recipes/393090/
    """
    tmant, texp = 0, 0
    for x in iterable:
        mant, exp = math.frexp(x)

        mant, exp = int(math.ldexp(mant, mant_dig)), exp - mant_dig
        if texp > exp:
            tmant <<= texp-exp
            texp = exp
        else:
            mant <<= exp-texp
        tmant += mant
    # Round tmant * 2**texp to a float.  The original recipe
    # used float(str(tmant)) * 2.0**texp for this, but that's
    # a little unsafe because str -> float conversion can't be
    # relied upon to do correct rounding on all platforms.
    tail = max(len(bin(abs(tmant)))-2 - mant_dig, etiny - texp)
    if tail > 0:
        h = 1 << (tail-1)
        tmant = tmant // (2*h) + bool(tmant & h and tmant & 3*h-1)
        texp += tail
    print(tmant, texp)

    print(float(tmant))
    print(float(-2492302492096719))
    print(float(int(str(tmant))))

    return math.ldexp(tmant, texp)


from random import random, gauss, shuffle, seed
seed(0)
# print(random())
for j in range(1):
    vals = [7, 1e100, -7, -1e100, -9e-20, 8e-20] * 3
    # s = 0
    # for i in range(200):
    #     # print(gauss(0, 0.8444218515250481))
    #     # v = gauss(0, 0.8444218515250481) ** 7 - s
    #     v = 0.044072055380373884 ** 2 - s
    #     print(v)
    #     s += v
    #     vals.append(v)
    # shuffle(vals)
    # print(vals)

    s = msum(vals)
```

test/unit3/test_math.py
```diff
@@ -635,28 +638,44 @@ def testfrexp(name, result, expected):
...
+                self.fail("test %d failed: got OverflowError, expected %r "
+                          "for math.fsum(%.100r)" % (i, expected, vals))
+            except ValueError:
+                self.fail("test %d failed: got ValueError, expected %r "
+                          "for math.fsum(%.100r)" % (i, expected, vals))
+            self.assertTrue(math.isclose(actual, expected, rel_tol=10**-15)) 
+            # AssertEqual failed on testcase 3,4,6,10 only in the units digit
+            # correct in first 15 significant figures!
+            # hence using isclose rather than assertEqual 
+
```
s-cork: I used `isclose` rather than `assertEqual` see comment

test/unit3/test_math.py
```diff
@@ -635,28 +638,44 @@ def testfrexp(name, result, expected):
...
+            s = 0
+            for i in range(200):
+                v = gauss(0, random()) ** 7 - s
+                s += v
+                vals.append(v)
+            shuffle(vals)
+        
+            s = msum(vals)
+            self.assertAlmostEqual(msum(vals), math.fsum(vals), 13)
+            # assertEqual failed - this seemed the best compromise
```
s-cork: `assertEqual` failed, `assertAlmostEqual` at the 13th decimal places worked

src/complex.js
albertjan: :+1: please remove the code

test/unit3/test_math.py
albertjan: this is really weird.

s-cork: done with commit abc066f 👍

src/complex.js
s-cork: done with commit abc066f 👍

meredydd:
Well, this is *excellent*! Approved to merge...once that weird test failure gets sorted out.

(I think that's just another equality->`isclose` thing, that got shaken loose by a different ordering of some other Javascript deep in V8)

albertjan:
thank you @s-cork !

rixner:
This pull request "silently" pulled in ```BigInt```.  In ```package.json``` it still says that we work for node versions >=8.10.  However, ```BigInt``` wasn't added to node until 10.4.  I'm not sure if we care about old browser compatibility, as well.  But, at a minimum, package.json should have been updated.  I discovered that because I'm still running node 8.10 on some machines.  I'm fine with updating (I'm running an old version out of laziness more than anything else), but this should be communicated.

(@s-cork otherwise, thanks for the contributions!)

alexwenbj:
The function round()/floor()/ceil() output not same as Python 3.6, for example:
```
from math import *
a = 2.13
b = 3.7777
c = -3.12
print(int(a), floor(b), ceil(c), round(c)) 
```
print out:`(2, 3.0, -3.0, -3.0)`
but in Python 3.6, it shows:
`2 3 -3 -3`

@s-cork Would you please confirm this?Thank you.

PR #1009, opened 2020-02-01
Performance improvements

meredydd:
This PR combines a series of performance improvements, totalling a ~45%~ **85%** speedup (aka a ~31%~ 46% reduction in runtime) on @rixner's student project ([`field.py`](https://gist.githubusercontent.com/rixner/f79dfa52a0385e5c82d147ae3d06c50c/raw/971a46b80c9ff79734e80f7bab49f66ecdc3ae11/field.py)).

Tagging @rixner because this is relevant to your interests, @albertjan because @rixner is busy, and @LeszekSwirski for arcane mysteries of V8 performance.

I'd suggest reviewing each commit individually: Each improvement is a single commit, with a paragraph of notes in the commit message.

Performance numbers throughout use the median of 10ish measurements on my laptop, of 500 ticks of `field.py`. (I made a desultory effort to get the CPU throttling into a consistent state, but it's pretty rough; T tests averaged around p=0.10.)

----
**Notes and Highlights:**

The hottest code was in the attribute-getting infrastructure:

* We win 16% by reworking an inefficient fallback from `klass.prototype.tp$getattr()` to `object.GenericPythonGetAttr`, which is of course the common case.

* Removing `klass.prototype.tp$getattr()` entirely when unnecessary (ie when the class doesn't define a custom `__getattribute__`) got us another 10%

* `Sk.abstr.gattr()` is an anomalously hot spot. Even after streamlining, it's still nearly 4% of CPU (according to the Chrome profiler), despite doing very little!

  - I initially tried inlining the `tp$getattr()` call straight into generated code, which won us quite a lot (4.5% speedup). I'm guessing this is because V8 could now do call-site-specific optimisation (each line of user code pretty consistently calls one `tp$getattr` implementation).

  - However, the benefit from that inlining basically disappeared after I removed `tp$getattr()` from most classes. This is presumably because, now that most classes share a single `tp$getattr()` implementation (in fact, _all_ classes in this particular program do), V8 can do call-site-specific optimisation on the `tp$getattr()` call in `Sk.abstr.gattr()` itself.

    I am slightly concerned that more complex programs that manipulate other types -- modules, exceptions, classes that implement `__getattribute__`, etc -- might prevent V8 from optimising that site. So perhaps we should still be inlining those calls, but I spent 5 minutes messing around with microbenchmarks and couldn't find evidence to justify it, so I'm holding off for now. (It's in this PR, but reverted.)

* Most attrs are looked up in the object dictionary, so a fast path for string comparisons in `Sk.builtin.dict.prototype.key$lookup` proved well worth it, netting us another 21% (harvested in two parts: 8.8% from remarkably small tweaks to `richCompareBool`, and then 16.3% from bypassing it altogether when doing string lookups)

The function-call infrastructure is next on my hit list. I've grabbed a couple of pieces of low-hanging fruit, but there's plenty still to get:

* `Sk.misceval.callsimOrSuspendArray()` has an API almost identical to `tp$call()`, but is much slower (it calls `applyOrSuspend()`, which goes round the houses of all the things Skulpt knows how to invoke). So I've added a fast path that invokes `tp$call` directly from generated code when possible.

  - I'm not actually sure we _should_ support invoking anything that doesn't have a `tp$call` method. We shouldn't sacrifice common-case performance to make things marginally easier for JS devs (eg allowing you to invoke JS functions). Forcing _all_ calls to use `tp$call` would speed up everything and clear out a bunch of code, but it would break our external API so I've held off.

* I've cherry-picked #950 into this, and the fast path is indeed catching most executions, but `func.tp$call` (and `method.tp$call`) are still taking ~4% of runtime each. Plenty to be getting on with here!

### DIFF
--- a/src/dict.js
+++ b/src/dict.js
@@ -76,6 +76,18 @@ Sk.builtin.dict.prototype.key$lookup = function (bucket, key) {
     var eq;
     var i;
 
+    // Fast path: We spend a *lot* of time looking up strings
+    // in dictionaries. (Every attribute access, for starters.)
+    if (key.ob$type === Sk.builtin.str) {
+        for (i = 0; i < bucket.items.length; i++) {
+            item = bucket.items[i];
+            if (item.lhs.ob$type === Sk.builtin.str && item.lhs.v === key.v) {
+                return item;
+            }
+        }
+        return null;
+    }
+
     for (i = 0; i < bucket.items.length; i++) {
         item = bucket.items[i];
         eq = Sk.misceval.richCompareBool(item.lhs, key, "Eq");
--- a/src/misceval.js
+++ b/src/misceval.js
@@ -854,9 +854,9 @@ Sk.exportSymbol("Sk.misceval.callsim", Sk.misceval.callsim);
  * Does the same thing as callsim without expensive call to Array.slice.
  * Requires args to be a Javascript array.
  */
-Sk.misceval.callsimArray = function(func, args) {
+Sk.misceval.callsimArray = function(func, args, kws) {
     var argarray = args ? args : [];
-    return Sk.misceval.apply(func, undefined, undefined, undefined, argarray);
+    return Sk.misceval.apply(func, undefined, undefined, kws, argarray);
 };
 Sk.exportSymbol("Sk.misceval.callsimArray", Sk.misceval.callsimArray);
 
@@ -884,14 +884,24 @@ Sk.exportSymbol("Sk.misceval.callsimOrSuspend", Sk.misceval.callsimOrSuspend);
 
 /**
  * @param {Object} func the thing to call
- * @param {Array=} args an array of arguments to pass to the func
+ * @param {Array} args an array of arguments to pass to the func
+ * @param {Array=} kws an array of keyword arguments to pass to the func
  *
  * Does the same thing as callsimOrSuspend without expensive call to
- * Array.slice.  Requires args to be a Javascript array.
+ * Array.slice.  Requires args+kws to be Javascript arrays.
  */
-Sk.misceval.callsimOrSuspendArray = function (func, args) {
-    var argarray = args ? args : [];
-    return Sk.misceval.applyOrSuspend(func, undefined, undefined, undefined, argarray);
+Sk.misceval.callsimOrSuspendArray = function (func, args, kws) {
+    if (!args) {
+        args = [];
+    }
+    if (func.tp$call) {
+        return func.tp$call(args, kws);
+    } else {
+        // Slow path handles things like calling native JS fns
+        // (perhaps we should stop supporting that), and weird
+        // detection of the __call__ method (everything should use tp$call)
+        return Sk.misceval.applyOrSuspend(func, undefined, undefined, kws, args);
+    }
 };
 Sk.exportSymbol("Sk.misceval.callsimOrSuspendArray", Sk.misceval.callsimOrSuspendArray);
 
--- a/src/misceval.js
+++ b/src/misceval.js
@@ -257,8 +257,8 @@ Sk.misceval.richCompareBool = function (v, w, op, canSuspend) {
     Sk.asserts.assert((v !== null) && (v !== undefined), "passed null or undefined parameter to Sk.misceval.richCompareBool");
     Sk.asserts.assert((w !== null) && (w !== undefined), "passed null or undefined parameter to Sk.misceval.richCompareBool");
 
-    v_type = new Sk.builtin.type(v);
-    w_type = new Sk.builtin.type(w);
+    v_type = v.ob$type;
+    w_type = w.ob$type;
 
     // Python 2 has specific rules when comparing two different builtin types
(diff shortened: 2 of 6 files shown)


src/misceval.js
```diff
@@ -884,14 +884,24 @@ Sk.exportSymbol("Sk.misceval.callsimOrSuspend", Sk.misceval.callsimOrSuspend);
...
-    return Sk.misceval.applyOrSuspend(func, undefined, undefined, undefined, argarray);
+Sk.misceval.callsimOrSuspendArray = function (func, args, kws) {
+    if (!args) {
+        args = [];
+    }
+    if (func.tp$call) {
+        return func.tp$call(args, kws);
+    } else {
+        // Slow path handles things like calling native JS fns
+        // (perhaps we should stop supporting that), and weird
```
albertjan: we should stop supporting raw js functions

rixner:
@meredydd you are on fire here!  I will look at these in detail this week!

rixner:
I am seeing a 2X speedup for the whole thing on a different benchmark.  Very nice work!  I'll look commit-by-commit.

src/dict.js
```diff
@@ -76,6 +76,18 @@ Sk.builtin.dict.prototype.key$lookup = function (bucket, key) {
     var eq;
     var i;
 
+    // Fast path: We spend a *lot* of time looking up strings
+    // in dictionaries. (Every attribute access, for starters.)
+    if (key.ob$type === Sk.builtin.str) {
+        for (i = 0; i < bucket.items.length; i++) {
+            item = bucket.items[i];
+            if (item.lhs.ob$type === Sk.builtin.str && item.lhs.v === key.v) {
```
rixner: I marginally prefer using ```item.lhs.$jsstr() === key.$jsstr()``` here to get away from assuming ```.v``` is the representation, but this is internal and it is so pervasive, that it probably doesn't matter...

src/misceval.js
rixner: I agree.  Do any still exist in the core of Skulpt?

src/type.js
```diff
@@ -459,6 +441,30 @@ Sk.builtin.type = function (name, bases, dict) {
...
+                );
+                return canSuspend ? r : Sk.misceval.retryOptionalSuspensionOrThrow(r);
+            };
+        } else if (!klass.prototype.tp$getattr) {
+            // This is only relevant in Python 2, where
+            // it's possible not to inherit from object
+            // (or perhaps when inheriting from builtins? Unclear)
+            klass.prototype.tp$getattr = Sk.builtin.object.prototype.GenericGetAttr;
+        }
+
```
rixner: What happens if you monkey patch the class to add a ```__getatttribute__``` method here?  Personally, I don't care, because you deserve what you get (in my opinion!) if you do that, but does that create a correctness issue?

src/dict.js
meredydd: That function dispatch would be expensive (this code is red-hot); if there were _anywhere_ to inline `$jsstr()`, it's here.

src/misceval.js
meredydd: I didn't spot any, but I'm not confident I turned over every rock.

src/type.js
meredydd: You're right; it now breaks, like every other dunder-method (monkey-patching of `__len__`, `__getitem__`, or others isn't reflected in the JS slot functions either).

I can imagine a custom `tp$setattr` for classes that updates the prototype if you do a monkey-patch like that...although that won't work if you do it to a multiply-inherited ancestor (JS prototypical inheritance does not follow the Python MRO!). Subclasses could register a callback to be notified if it happens, I suppose, but that makes subclasses non-garbage-collectable, which I'm not too keen on.

Or we could set a global "someone monkey-patched something important" flag, which causes _all_ slot functions to deoptimise forevermore...although that doesn't help if the slot functions aren't even defined (which is the case for `sq$len` and friends).

Basically, it's a gnarly one to support, and our support for all the _other_ slot functions is already intolerant of monkey-patching, so I'm inclined to let it stay broken unless we can work out how to fix them all at once.

meredydd:
> However, I would at least like you to think about/respond to the issue of adding a `__getattribute__` method to a class dynamically and whether it is actually a problem (and/or if you care about it).

See my reply to your comment. Yes, this breaks a piece of monkey-patching, but we don't support monkey-patching anywhere else in the dunder/slot functions, so I'm (reluctantly, and while holding my nose) calling it an acceptable cost.

If there were a reasonably performant way of making monkey patching actually work, I'd be up for it.

src/dict.js
rixner: Fair enough.

rixner:
As I implied, I'm not a fan of monkey patching, so am not at all offended by this.  Especially since we don't fully support it elsewhere anyway.  This was a minor step backward in that regard, but the win is large and I agree with the sentiment that if we believe this should be supported, they all should support it, not just selected methods.  I'm going to very happily merge this!  Again: great work!

PR #1013, opened 2020-02-02
Performance: Directly call functions

meredydd:
Summary: By changing the calling convention of Skulpt's function objects, we can get a ~6% performance improvement. (And we could get a lot more if we could mess with the external API.)

(This PR is on top of #1009)

----

Hitherto, every time we call a function, we run `func.tp$call()`, which unpacks the argument values (posargs and kwargs), matches them against the function's formal arguments, and then calls the implementation (eg generated code). This is a big hot-spot (7-8% of runtime in `field.py`), and it's quite inefficient (among other reasons: It's extra layer of indirection, it uses `apply()`, and it looks up all the argument flags [`co_varnames` etc] on a very polymorphic value `func_code`).

This change turns that control flow inside out: The generated code *becomes* the `tp$call`, which (combined with #1009) means that one compiled function calls another directly -- that's less indirection _and_ easier for V8 to optimise.

Argument unpacking becomes a library function (`func.prototype.$resolveArguments()`), which simple functions can avoid calling altogether!

(While I'm at it, I also optimise the lookups of the function flags - `co_argnames` and friends - by saving them into the `func` object, where they are much faster to look up.)

---

Unfortunately, lots of native JS functions (both inside and outside Skulpt) are using the old calling convention, so we can't replace it wholesale with fast calls. This costs us a lot of performance on `field.py`, as it makes heavy use of math functions (`max()`, `sqrt()`, etc), which still have to bounce through `tp$call()`. I've added another fast path to the rump `tp$call()` to deal with native functions, which helps. I think it would be worth making a breaking change to the API to clean this up.


src/function.js
```diff
@@ -260,6 +260,19 @@ Sk.builtin.func = function (code, globals, closure, closure2) {
     if (code.co_fastcall) {
         this.tp$call = code;
     }
+    // Because our external API allows you to set these flags
+    // *after* constructing the function (grr), we can only
```
albertjan: should we make them part of the `Sk.func` constructor instead of plonking them on the prototype of `func_code`?

src/compile.js
```diff
@@ -2043,7 +2050,7 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
 
     // inject __class__ cell when running python3
     if (Sk.python3 && class_for_super) {
-        this.u.varDeclsCode += "$gbl.__class__=this." + class_for_super.v + ";";
```
albertjan: this is an odd change 😄

src/function.js
```diff
@@ -430,6 +434,7 @@ Sk.builtin.func.prototype.$resolveArgs = function (posargs, kw) {
 
 
 Sk.builtin.func.prototype.tp$call = function (posargs, kw) {
+    console.log("Legacy tp$call for", this.tp$getname(), this.func_code);
```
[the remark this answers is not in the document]
meredydd: Yes; I'm planning to drop an issue today with all the API changes I'd like to make, and this is one of them :)

src/function.js
meredydd: Yes, we should! That will make constructing functions from JS much nicer, too. But again, it's an API-breaking change, so I'm leaving it out of this PR.

src/compile.js
meredydd: It's because, in `co_fastcall` land, `this` is no longer the globals object; it's the `func` object.

albertjan: ah yes that makes sense

rixner:
Do we want to consider this?  Or do we want to think about a breaking API change first?  If we all agree to a breaking API change, isn't that better?  If we don't, then we can come back to this...  The impact of this was not immediately obvious to me on a quick read.  I'm happy to spend more time with it, though, if we think this path should be pursued independent of an API change.

meredydd:
> Do we want to consider this? Or do we want to think about a breaking API change first?

This is an incremental change that is still entirely valid if we make the API change.

Changing the API would allow us to _extend_ this optimisation to native JS-defined functions (all native funcs could become fastcalls - or at least we could have an even faster `tp$call` shim for the common case of native functions that do their own argument checks).

It wouldn't change the meat of this PR. After that API change, we could tidy up what is currently `$memoiseFlags()` into the `func` constructor, and factor out the "unchecked native function" fast-path into its own `tp$call`, but the rest of this PR remains a good idea.

rixner:
This all seems fine, but it has very little effect.  I do see small gains on some programs, but on many others, it's a wash.  On its own, I'm not sure this is worth the added complexity.  As a pre-curser to #1014 it makes sense.  I'm fine with either pulling this or waiting for further discussion on #1014.

s-cork:
This LGTM, since this is not a breaking change I think we should merge. (after travis is happy)

I found the performance improvements running `fields.py` were significant. 

```
100 ticks (avg of 10 iterations using npm run repl py2)
this_branch: ('avg', 1.18779995441)
master   : ('avg', 2.7121999979)
```

src/compile.js
s-cork: This seems to be the change that is preventing it passing the tests.
For some reason this change got lost in the version travis is testing...

albertjan:
Oh no now we have a failing test:

```
UNCAUGHT EXCEPTION: KeyError: __class__ on line 275
test/unit3/test_complex.py
```

src/compile.js
```diff
@@ -2051,8 +2051,8 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
     this.u.varDeclsCode += "}";
 
     // inject __class__ cell when running python3
-    if (Sk.python3 && class_for_super) {
-        this.u.varDeclsCode += "$gbl.__class__=$gbl." + class_for_super.v + ";";
+    if (Sk.__future__.python3 && class_for_super) {
+        this.u.varDeclsCode += "$gbl.__class__=this." + class_for_super.v + ";";
     }
```
s-cork: `this` should be changed back go `$gbl` here for this pr.

albertjan: Could you push this change? 😄

rixner:
Sorry, I screwed this up when merging master, I think!

src/compile.js
s-cork: Done - I had no idea I could push changes to someone else branch like that!

s-cork:
is this one mergeable now?

PR #1016, opened 2020-02-04
Added small fix to scientific notation string formatting

darkaardvark2:
Previously, the following statement 
`print("{:e}".format(123456789))`
would show
`1.234568e+8`

Now fixed power number padding to show
`1.234568e+08`


src/formatting.js
```diff
@@ -189,6 +189,16 @@ let formatNumber = function(num, formatSpec, isFractional) {
                     result += ".0";
                 }
             }
+            // Python's 'e' pads the power with 0 if it is a single digit
+            if (conversionType.toLowerCase()==="e") {
```
meredydd: We should probably also handle the `g` conversion type too!

src/formatting.js
```diff
@@ -189,6 +189,16 @@ let formatNumber = function(num, formatSpec, isFractional) {
...
+                let scientificNotation = result.match(/[-+]?[0-9]*\.?[0-9]+[eE][-+]?([0-9]+)?/);
```
meredydd: Code-size nit: Can we collapse this into one `replace()` call? Something like:

```javascript
result = result.replace(/^([-+]?[0-9]*\.?[0-9]+[eE][-+]?)([0-9])?$/, "$10$2");
```

rixner:
Can we get this updated so that we can get it merged?

meredydd:
Hi @darkaardvark2 - would you be able to update this PR? We'd love to get this merged!

s-cork:
closed via #1226

PR #1017, opened 2020-02-04
Update node packages and engines.

rixner:
#1004 uses ```BigInt``` which wasn't introduced into node until version 10.4.  Update engines in package.json.  Also update some other packages that npm complains have security issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "closure-webpack-plugin": "^2.0.1",
     "commander": "^2.20.0",
     "compression-webpack-plugin": "^2.0.0",
-    "copy-webpack-plugin": "^5.0.3",
+    "copy-webpack-plugin": "^5.1.1",
     "ejs": "^2.6.1",
     "eslint": "^5.16.0",
     "eslint-loader": "^2.1.2",
@@ -58,7 +58,7 @@
     "git-revision-webpack-plugin": "^3.0.3",
     "google-closure-compiler": "^20190415.0.0",
     "js-beautify": "^1.10.0",
-    "jsdoc": "~3.5.5",
+    "jsdoc": "^3.6.3",
     "micro-strptime": "^0.2.3",
     "open": "^6.3.0",
     "readline-sync": "^1.4.9",
@@ -69,7 +69,7 @@
     "webpack-cli": "^3.3.2"
   },
   "engines": {
-    "node": ">=8.10"
+    "node": ">=10.4"
   },
   "dependencies": {}
 }
--- a/src/lib/math.js
+++ b/src/lib/math.js
@@ -217,25 +217,30 @@ var $builtinmodule = function (name) {
                 return a;
             };
             return _gcd(b, a % b);
-        };
+        }
 
-        let _a = Math.abs(Sk.builtin.asnum$(a));
-        let _b = Math.abs(Sk.builtin.asnum$(b));
-        let max_safe = false;
+        function _biggcd(a, b) {
+            if (b.trueCompare(Sk.builtin.biginteger.ZERO) === 0) {
+                return a;
+            }
+            return _biggcd(b, a.remainder(b));
+        }
 
-        if (_a >= Number.MAX_SAFE_INTEGER || _b >= Number.MAX_SAFE_INTEGER) {
-            _a = BigInt(Sk.builtin.str(a).$jsstr());
-            _b = BigInt(Sk.builtin.str(b).$jsstr());
-            max_safe = true;
-        };
+        if ((a instanceof Sk.builtin.lng) || (b instanceof Sk.builtin.lng)) {
+            let _a = Sk.builtin.lng(a).biginteger;
+            let _b = Sk.builtin.lng(b).biginteger;
+            let res = _biggcd(_a, _b);
+            res = res.abs(); // python only returns positive gcds
 
-        let res = _gcd(_a, _b);
-        res = res < 0 ? -res : res; // python only returns positive gcds
+            return new Sk.builtin.lng(res);
+        } else {
+            let _a = Math.abs(Sk.builtin.asnum$(a));
+            let _b = Math.abs(Sk.builtin.asnum$(b));
+            let res = _gcd(_a, _b);
+            res = res < 0 ? -res : res; // python only returns positive gcds
 
-        if (max_safe) {
-            return new Sk.builtin.lng(res.toString());
-        };
-        return new Sk.builtin.int_(res);
+            return new Sk.builtin.int_(res);
+        }
     });
 
 
--- a/src/lib/math.js
+++ b/src/lib/math.js
@@ -68,27 +68,40 @@ var $builtinmodule = function (name) {
         Sk.builtin.pyCheckType("x", "number", Sk.builtin.checkNumber(x));
 
         let _x = Sk.builtin.asnum$(x);
+        x = Math.floor(_x);
 
-        if (!Number.isInteger(_x)) {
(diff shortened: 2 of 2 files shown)


albertjan:
Ah sorry about that. I missed that. Weren't we already using a transparent polyfill for BigInt?

rixner:
I don't think we use ```BigInt``` anywhere else...  Perhaps we now need a polyfill?  We've never really settled on what level of support we want for older browsers...

albertjan:
Ah! I knew I had done this work somewhere. In my first attempt to update to node I had removed our biginteger.js and replaced it with a polyfill for BigInt. The browser support for BigInt is still very small so I think we'll have to do this I'll see if I can excavate that code and do a PR.

albertjan:
https://caniuse.com/#search=bigint

rixner:
Yeah, we definitely need a polyfill.  Pulling #1004 was premature. :)

If we have the polyfill, biginteger.js should go, so I hope you can extract that part easily!

rixner:
I suspect it is going to be difficult to drop in ```BigInt``` with a polyfill, especially since you can't use a transparent polyfill.  While I think we should try to replace ```Sk.builtin.biginteger``` with ```BigInt```, I think that's a larger undertaking.  So, I replaced the use of ```BigInt``` in math.js with ```Sk.builtin.biginteger```.  I think that's the easier/better short term solution.

I favor pulling this quickly so that we can get back to being compatible with all browsers (Safari, I'm looking at you) and then if someone is excited about doing so, we can replace ```Sk.builtin.biginteger``` with ```BigInt``` somehow with an appropriate polyfill in another pull request.

I'm leaving the node engines declaration at 10.4, but can reset it to 8.10 if people think that's the right thing to do.

albertjan:
@rixner see: https://github.com/skulpt/skulpt/pull/857/commits/6f07236faf6cbadfbce7b387c9feda0cd8bd5540

rixner:
@albertjan that looks good.  I didn't realize you found one that has the same interface we already use.  That isn't "transparent" in the sense that we still can't have code directly using ```BigInt``` like math.js does, so math.js still needs to be "fixed" as I did here.


I do believe we all discussed this in the past and JSBI was suggested instead, because babel can convert it directly to ```BigInt``` once there is universal support and we no longer want a polyfill.  I'll try to dig up that conversation. 

How do you want to proceed?

rixner:
(Also we can't require/import node packages in libraries, so whatever polyfill we use will have to be exported via ```Sk``` in order for math.js to be able to use it.)

rixner:
The previous discussion (#729) wasn't as extensive as I had remembered.  @albertjan you and I agreed we should use a polyfill, and then we didn't. :smile:

rixner:
@albertjan, using your code as a guide, I updated everything to use ```big-integer``` and it seems to work fine with the dev build.  However, the closure compiler does not like the ```big-integer``` library, so I can't do a production build.  Not sure what to do at this point...

rixner:
See my [bigint branch](/rixner/skulpt/tree/bigint).

albertjan:
wow @rixner that's a lot of errors

s-cork:
Nice job @rixner Sorry for the small headache I caused 🤕 ... still learning ... 📘 ✏️

rixner:
@s-cork no problem - the reviewers missed it too!  This pull request fixes it, so no real damage was done.  The "headache" is in getting a library/polyfill to work, which is entirely unrelated, as far as I'm concerned.

rixner:
@albertjan yeah, I was surprised.  I think the only way to get it to work is to not run it through the closure compiler.  That's a level of webpack/closure gymnastics that I'm not sure I want to undertake.  I'm going to look a little bit at other libraries.

albertjan:
there must be a way to tell closure this is an external dependency and it shouldn't be typechecked

rixner:
JSBI is no better closure-wise.  I'll see if I can get closure to skip big-integer.

rixner:
I am unable to get it to work at this point.  Both webpack and the closure compiler have the notion of externals and third party libraries, but I can't find the right set of magic incantations to get everything to build.  

One initial problem is that ```BigInteger.js``` is under a node package called ```big-integer```.  The dash can't be used in a valid Javascript identifier, which bedevils webpack.  Even if I move the file itself and just work with that, I still run into mystifying errors from the closure compiler.

Unless anyone else knows how to get this all to work, I'd like to just get this pull request merged so that we are back to being browser compatible and someone else can take a run at a BigInt library and hopefully have better luck than I am...  At this point, I'm stuck and don't have additional time to work on it.

s-cork:
is this relevant at all?  https://github.com/google/closure-compiler/issues/3167

albertjan:
Alright lets merge this first and figure out how to deal with the polyfill later. Maybe @meredydd has an idea.

rixner:
I'll come back to this at some point, as well, but I don't think we should block this merge until that happens...

@s-cork that does reinforce that we can't run the library through the closure compiler, but I was trying (unsuccessfully) to not run it through the closure compiler for other reasons.

rixner:
I figured out the problem.  And it's not good.

In ```biginteger.js``` we have a ```@supress {checkTypes}``` directive.  That was turning off the closure compiler's type checking for code outside that file.  When I stopped including that file, I started getting errors all over the place, because the closure compiler is now actually type checking all of the code.  So, it's not that I was doing the wrong thing in terms of the big integer library, but rather that we have a lot of latent type errors that get exposed.  I'm not sure where to go with that.

rixner:
There are a bunch of issues with the ast and type checking.  I am wondering if some of the cleanup in #991 would help...

There are also several legitimate errors, rather than just incomplete/bad type annotations on our part, that closure is catching.  I don't know why they don't cause perceivable problems.

rixner:
I'm down to 17 errors from the closure compiler, which is significantly better than where things started.  I will see if I can get rid of any more, but a good deal of them are references to AST nodes and Tokens that are just not defined anywhere.  I'm not sure if they used to be defined and we lost them somewhere along the line or if there's speculative code that assumes they'll be implemented some day or what...  Presumably I could delete them without any negative impact since they're not doing anything useful, but I'm not sure if that's the right approach.

Examples:  T_BACKQUOTE, Sk.astnodes.Repr, exec_stmt

rixner:
I believe I have eliminated all of the closure compiler errors (maybe not in the best way possible), but I have as of yet been unable to get it to rebuild a working ```skulpt.min.js```.  The issues have to do with webpack and splitting and merging the files.  It still works with the devbuild so I don't think the issue is that I broke the code, I really think it's just not bundled correctly anymore.

Webpack also injects some code that has (1) closure compiler errors in it and (2) references to ```window``` which won't work in node.  I don't know that there's any way around that other than to modify the webpack source, which is less than desirable.  As a test, I did modify the webpack source to eliminate (1) and then tried to run the result in a browser, where ```window``` should work, but that wasn't working for me either.

At this point, I think we should just merge this pull request.  We do have the somewhat serious issue, though, that the closure compiler is not doing type checking anywhere, as far as I can tell.  So, we have to resolve that (in my opinion), whether we use a big integer library or not.

So, @skulpt/committers please merge this so that we regain browser compatibility while we work out the other issues (or ignore them, I guess)!

albertjan:
@rixner did you forget to push to this branch? Or do you want to update the type stuff in a different PR.

rixner:
@albertjan I have another PR that I'll submit after this one with my fixes (including the type stuff).  I have it "working" now.  But, that's going to require careful review and discussion, whereas this does not.  Let's merge this, which is the minimum to get back to where we were, and then I'll open that one.

PR #1019, opened 2020-02-06
Add copy() function for Python 3 list

alexwenbj:
Add copy() function for Python 3 list.


s-cork:
doesn't list already have a copy method on line `655`?

in both 2 and 3 environments, I get
```python
>>> l = [1,2,3]
>>> l.copy()
[1,2,3]
```

alexwenbj:
@s-cork Yes,I made a mistake.This is a wrong implementation.Thank you.I will close this PR.

PR #1020, opened 2020-02-07
Pr/itertools

### LINKED ISSUE #468: itertools permutations
Planning on working on this with a tight turnaround- hoping to get this deployed for a user by Weds evening ET for an event Thurs AM GMT.  Thought I'd drop an issue in case anyone has an interest in permutations and may be able to get to this before I can (which may be Weds PM).

I found some JS that returns all permutations of any array.  Should be very doable to implement using something like the below, though tests may be tough because I'm not sure the ordering will be the same as Python's.  Haven't dug into the C implementation yet to check- running will probably be easier.

``` javascript
function permutator(inputArr) {
  var results = [];

  function permute(arr, memo) {
    var cur, memo = memo || [];

    for (var i = 0; i < arr.length; i++) {
      cur = arr.splice(i, 1);
      if (arr.length === 0) {
        results.push(memo.concat(cur));
      }
      permute(arr.slice(), memo.concat(cur));
      arr.splice(i, 0, cur[0]);
    }

    return results;
  }

  return permute(inputArr);
}
```

May be time to stub out some of the rest of itertools as well- haven't looked too deeply at it yet.

### LINKED ISSUE #971: Itertools not implemented
It seems that `itertools` hasn't been implemented yet, I checked in `src/lib/` and there was no `itertools.py`. Am I looking in the right place?

I can start work on this if `itertools` really has not been implemented, yet.

s-cork:
This PR implements `itertools`. Closes #971 and closes #468 

The following functions are implemented:
```
accumulate
chain
chain.from_iterable
combinations
combinations_with_replacement
compress
count
cycle
dropwhile
filterfalse
groupby
islice
permutations
product
repeat
starmap
takewhile
zip_longest
```

The following functions raise a `NotImplementedError`:
```
tee
```

_(I couldn't see a good reason to implement `tee`)_

---

- I used `Sk.builtin.func` and `Sk.builtin.generator` to implement each function.
- The code is based on the 'pure python' versions from the [itertools docs](https://docs.python.org/3/library/itertools.html) and adapted for Skulpt. 
- The tests are from https://github.com/python/cpython/blob/3.7/Lib/test/test_itertools.py
- There is one change to the `complex.js` file, namely the `prototype["$r"]` property

### DIFF
--- a/src/complex.js
+++ b/src/complex.js
@@ -865,7 +865,7 @@ Sk.builtin.complex.complex_format = function (v, precision, format_code){
 };
 
 Sk.builtin.complex.prototype["$r"] = function () {
-    return Sk.builtin.complex.complex_format(this, 0, "r");
+    return Sk.builtin.complex.complex_format(this, null, "g");
 };
 
 Sk.builtin.complex.prototype.tp$str = function () {
--- a/src/lib/itertools.js
+++ b/src/lib/itertools.js
@@ -41,7 +41,7 @@ var $builtinmodule = function (name) {
 
     var _accumulate_gen = function ($gen) {
         it = $gen.gi$locals.it;
-        func = $gen.gi$locals.func;
+        f = $gen.gi$locals.func;
         total = $gen.gi$locals.total;
         initial = $gen.gi$locals.initial;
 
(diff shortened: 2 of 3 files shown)


s-cork:
I did a quick test as I was curious about what the speed differences would be from simply using the pure python equivalents would be... 

```python
# skulpt itertools.js
list(permutations('abcdefghi'))
# 0.4767 sec

# skulpt using a pure python equivalent from the python itertools docs
list(permutations1('abcdefghi'))
# 48.7961 sec
```

rixner:
@albertjan I agree, merging.

albertjan:
:tada:

PR #1021, opened 2020-02-07
Py3 list copy

alexwenbj:


alexwenbj:
@s-cork The list copy() function was added in Python 3.3, it is not in Python 2.But there was an error in my previous implementation: the list copy is shallow copy.I modified it in this version.

rixner:
I feel like this is not something that should have it's own flag.  I think moving forward we had agreed to use the ```python3``` flag instead of adding new flags (like ```list_copy```).  I'd favor removing ```list_clear``` as well, and just using ```python3``` for both of those features.

Otherwise, this looks pretty straightforward.

alexwenbj:
@rixner What you mean is just like the below in copy() function:
```
if(Sk.__future__.python3){
    //copy implementation
}
```
and remove the list_copy config in env.js?

src/env.js
```diff
@@ -118,6 +120,7 @@ Sk.configure = function (options) {
     Sk.bool_check(Sk.__future__.dunder_next, "Sk.__future__.dunder_next");
     Sk.bool_check(Sk.__future__.dunder_round, "Sk.__future__.dunder_round");
     Sk.bool_check(Sk.__future__.list_clear, "Sk.__future__.list_clear");
```
albertjan: And this:

```suggestion
```

src/env.js
```diff
@@ -118,6 +120,7 @@ Sk.configure = function (options) {
...
+    Sk.bool_check(Sk.__future__.list_copy, "Sk.__future__.list_copy");
```
albertjan: And this: 
```suggestion
    Sk.bool_check(Sk.__future__.list_copy, "Sk.__future__.list_copy");
```

src/env.js
```diff
@@ -205,6 +208,7 @@ Sk.configure = function (options) {
     Sk.switch_version("round$", Sk.__future__.dunder_round);
     Sk.switch_version("next$", Sk.__future__.dunder_next);
     Sk.switch_version("clear$", Sk.__future__.list_clear);
```
albertjan: Change this to use `python3`
```suggestion
    Sk.switch_version("clear$", Sk.__future__.python3);
```

src/env.js
```diff
@@ -205,6 +208,7 @@ Sk.configure = function (options) {
...
+    Sk.switch_version("copy$", Sk.__future__.list_copy);
```
albertjan: And do the same here:

```suggestion
    Sk.switch_version("copy$", Sk.__future__.python3);
```

alexwenbj:
@albertjan Got it & updated it.

rixner:
@alexwenbj this looks good now!  Sorry for not being more clear about my requested changes.

alexwenbj:
@rixner Thank you.

PR #1024, opened 2020-02-10
Implement Python3 range object

rixner:
This is an implementation of Python3's range object.  It allowed me to uncomment many of the commented out tests in test_range.py.

It computes the entire range ahead of time instead of being a generator object, which is no worse than what we do now.  But, that means it is inefficient for large ranges.  But, a lot of other capabilities are improved.

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -7,20 +7,20 @@
 
 Sk.builtin.range = function range (start, stop, step) {
     var ret = [];
+    var lst;
     var i;
 
     Sk.builtin.pyCheckArgsLen("range", arguments.length, 1, 3);
-    Sk.builtin.pyCheckType("start", "integer", Sk.builtin.checkInt(start));
+    Sk.builtin.pyCheckType("start", "integer", Sk.misceval.isIndex(start));
+    start = Sk.misceval.asIndex(start);
     if (stop !== undefined) {
-        Sk.builtin.pyCheckType("stop", "integer", Sk.builtin.checkInt(stop));
+        Sk.builtin.pyCheckType("stop", "integer", Sk.misceval.isIndex(stop));
+        stop = Sk.misceval.asIndex(stop);
     }
     if (step !== undefined) {
-        Sk.builtin.pyCheckType("step", "integer", Sk.builtin.checkInt(step));
+        Sk.builtin.pyCheckType("step", "integer", Sk.misceval.isIndex(step));
+        step = Sk.misceval.asIndex(step);
     }
-
-    start = Sk.builtin.asnum$(start);
-    stop = Sk.builtin.asnum$(stop);
-    step = Sk.builtin.asnum$(step);
 
     if ((stop === undefined) && (step === undefined)) {
         stop = start;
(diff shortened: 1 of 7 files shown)


s-cork:
@rixner - might not be helpful at all. This is what I've come up with for what the generator version might look like...? This was based on my approach in the itertools/pr.

It would then require a different approach for methods like `sq$contains` etc...  

```javascript
Sk.builtin.range = function range (start, stop, step) {
    Sk.builtin.pyCheckArgsLen("range", arguments.length, 1, 3);
    Sk.builtin.pyCheckType("start", "integer", Sk.misceval.isIndex(start));
    start = Sk.misceval.asIndex(start);
    if (stop !== undefined) {
        Sk.builtin.pyCheckType("stop", "integer", Sk.misceval.isIndex(stop));
        stop = Sk.misceval.asIndex(stop);
    }
    if (step !== undefined) {
        Sk.builtin.pyCheckType("step", "integer", Sk.misceval.isIndex(step));
        step = Sk.misceval.asIndex(step);
    }

    if ((stop === undefined) && (step === undefined)) {
        stop = start;
        start = 0;
        step = 1;
    } else if (step === undefined) {
        step = 1;
    }

    if (step === 0) {
        throw new Sk.builtin.ValueError("range() step argument must not be zero");
    }


    _range_gen = function ($gen) {
            const start = $gen.gi$locals.start;
            const step = $gen.gi$locals.step;
            const stop = $gen.gi$locals.stop;
            const compare = step.nb$ispositive() ? Sk.misceval.isTrue(start.ob$lt(stop)) : Sk.misceval.isTrue(start.ob$gt(stop));
            if (compare) {
                try {
                    return [ /*resume*/ , /*ret*/ Sk.builtin.int_(start)];
                } finally {
                    $gen.gi$locals.start = start.nb$add(step);
                }
            }
            return [ /*resume*/ , /*ret*/ ];
    };   
    _range_gen.co_varnames = ["start", "stop", "step"];
    _range_gen.co_name = new Sk.builtin.str("range");

    pystart = typeof start === "number" ? Sk.builtin.int_(start) : Sk.builtin.lng(start);
    pystop = typeof stop === "number" ? Sk.builtin.int_(stop) : Sk.builtin.lng(stop);
    pystep = typeof step === "number" ? Sk.builtin.int_(step) : Sk.builtin.lng(step);

  range_gen = new Sk.builtin.generator(_range_gen, undefined, [pystart, pystop, pystep])

    if (Sk.__future__.python3) {
        return new Sk.builtin.range_(start, stop, step, range_gen);
    }

    return range_gen;
};

```

albertjan:
@s-cork does your PR include this?

s-cork:
> @s-cork does your PR include this?

#1020 `pr/itertools`? - no I haven't touched range in that pr. But I did make a lot of generators there 🙂 

when @rixner mentioned that the range object would ideally be a generator the above code was a back of the napkin approach to making `range` a generator object. 

I tested it today. It works very well on python 2. It will require some thought for creating a python 3 `range` object that includes the relevant methods like `contains` and `index` and `slice` since you have to work with the `start`, `stop`, `step` values rather than the generator. 

I found you also have to be a little careful, since the python 3 `range object` is different from a generator in that it can be reused. 

here's a [branch](https://github.com/s-cork/skulpt/tree/range-pr) I made today based on this... it doesn't' work for large numbers yet. And slicing a range was causing 🤕.

rixner:
@s-cork, the issue is not making a generator, which I knew how to do, but rather dealing with the complexity of subscript, slicing, index, count, etc.  It was not trivial getting all of that working.  But, it looks like you are finding that out. :smile:

In Python 2, range is supposed to return a list, so you shouldn't make it return a generator.  (You could add xrange, but given that Python 2 is EOL, I don't see the point of adding new features there.)

Also, by using ```nb$add``` instead of Javascript addition, you are going to slow down the range generation.  While a generator will make large ranges more efficient, I'd prefer not to lose performance on the much, much more common small ranges that appear everywhere in Python.

s-cork:
> looks like you are finding that out. 😄

yeah, it's very not trivial! 😄  👍

rixner:
I think we should evaluate this request and merge it (assuming the reviewers don't find any issues), and @s-cork can keep chipping away at using a generator if he wants and we can evaluate that independently.

As it stands, this adds functionality improvements that we don't have, so is a step forward.

PR #1026, opened 2020-02-13
Pr/textwrap

### LINKED ISSUE #261: textwrap
`fill()` and `wrap()` would be quite useful for getting well-line-broken text into the console.  Could be useful in printouts from an eventual `help()` function or other docstring dumps.  (#249)  

In the classroom: We've got some users who are teaching computational poetry.  textwrap would be very useful here. 

Some research is in order: if/how the module differs 2 vs 3 being the big one. And which to implement.  Any other decisions that need to be made or considerations?

[the 2 docs](https://docs.python.org/2/library/textwrap.html) 
[the 3 docs](https://docs.python.org/3/library/textwrap.html)

At first glance looks like 3 just adds a few things.

Edit: per @rixner's suggestion, a list of re and string things I found that need implementation for textwrap 2.x to work in skulpt:

`string`
- [ ] maketrans
- [x] expandtabs

`re`
- [ ] compile
- [ ] sub
- [ ] .pattern

s-cork:
This pr takes the cpython `textwrap.py` and changes some expressions where Skulpt does not yet support them. 
Namely - `re.sub`, `re.escape`, `str.translate`, `re.compile(pattern).search`
It also replaces the positive look-behind assertion --- `(?<` --- with a workaround due to browser incompatibility 

---

It implements the tests from https://github.com/python/cpython/blob/3.7/Lib/test/test_textwrap.py

Some tests were commented out due to failure:
- non-breaking spaces tests (not in the 2.7 test_textwrap.py)
- 4 tests fail that were added in python 3 but not in python 2.7 -- likely because the `wordsep-re` expression is lifted from [2.7](https://github.com/python/cpython/blob/2.7/Lib/textwrap.py) and not [3.7](https://github.com/python/cpython/blob/3.7/Lib/textwrap.py)
- umlaut tests
- `"foo (--option) bar"` test failed and split like `['foo', '(', '--', 'option)', 'bar']` because the positive look behind workaround will fail with an em-dash that immediately follows non-sentence ending punctuation...
  - the workaround does pass all other em-dash tests like` "foo --option-- bar"`, `"foo!--option-- bar"`,  `"foo--option-- bar"`.

--- 

It also removes the keyword-only argument (`*,`) syntax present in the 3.7 version since this will fail the 2.7 compilation... 

---

I think it closes #261 but there were also 5 todos on that issue...

### DIFF
--- a/src/lib/re.js
+++ b/src/lib/re.js
@@ -17,7 +17,7 @@ var $builtinmodule = function (name) {
     // mod.X = 64;
     // mod.VERBOSE = 64;
 
-    validGroups = ["(?:", "(?=", "(?!"];
+    validGroups = ["(?:", "(?=", "(?!", "(?<"];
 
     convert = function (pattern) {
         var newpattern;
--- a/src/lib/re.js
+++ b/src/lib/re.js
@@ -17,7 +17,7 @@ var $builtinmodule = function (name) {
     // mod.X = 64;
     // mod.VERBOSE = 64;
 
-    validGroups = ["(?:", "(?=", "(?!", "(?<"];
+    validGroups = ["(?:", "(?=", "(?!"];
 
     convert = function (pattern) {
         var newpattern;
--- a/src/lib/textwrap.py
+++ b/src/lib/textwrap.py
@@ -1,1 +1,456 @@
-raise NotImplementedError("textwrap is not yet implemented in Skulpt")
+"""Text wrapping and filling.
+"""
+
+# Copyright (C) 1999-2001 Gregory P. Ward.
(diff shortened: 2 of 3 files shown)


s-cork:
I think it might close #261 but there were also 5 todos on that issue...

s-cork:
💥 
and if #1048 is approved first I can uncomment all the `with self.assertRaises` tests 😉

rixner:
@s-cork as you probably realize, I merged #1048, so you can uncomment the tests.

s-cork:
> @s-cork as you probably realize, I merged #1048, so you can uncomment the tests.

Awesome! I've uncommented the relevant tests so good to go when you are.

Issue #1029, opened 2020-02-13
lack of frozenset() in built-in funcitons


alexwenbj:
lack of frozenset() in builtin funcitons.

s-cork:
# close 💥 great work @alexwenbj

PR #1030, opened 2020-02-14
Add frozenset to builtin function(Got things done!)

alexwenbj:
1. Add frozenset;
2. Add `__contains__ `to set;
3. Modify lib/keyword.py,same as the Python implementation.

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -1379,10 +1379,6 @@ Sk.builtin.execfile = function execfile () {
     throw new Sk.builtin.NotImplementedError("execfile is not yet implemented");
 };
 
-Sk.builtin.frozenset = function frozenset () {
-    throw new Sk.builtin.NotImplementedError("frozenset is not yet implemented");
-};
-
 Sk.builtin.help = function help () {
     throw new Sk.builtin.NotImplementedError("help is not yet implemented");
 };
--- a/src/env.js
+++ b/src/env.js
@@ -34,7 +34,6 @@ Sk.python2 = {
     unicode_literals: false,
     // skulpt specific
     python3: false,
-    set_repr: false,
     class_repr: false,
     inherit_from_object: false,
     super_args: false,
@@ -56,7 +55,6 @@ Sk.python3 = {
     unicode_literals: true,
     // skulpt specific
     python3: true,
-    set_repr: true,
     class_repr: true,
     inherit_from_object: true,
     super_args: true,
@@ -106,7 +104,6 @@ Sk.configure = function (options) {
     Sk.bool_check(Sk.__future__.print_function, "Sk.__future__.print_function");
     Sk.bool_check(Sk.__future__.division, "Sk.__future__.division");
     Sk.bool_check(Sk.__future__.unicode_literals, "Sk.__future__.unicode_literals");
-    Sk.bool_check(Sk.__future__.set_repr, "Sk.__future__.set_repr");
     Sk.bool_check(Sk.__future__.class_repr, "Sk.__future__.class_repr");
     Sk.bool_check(Sk.__future__.inherit_from_object, "Sk.__future__.inherit_from_object");
     Sk.bool_check(Sk.__future__.super_args, "Sk.__future__.super_args");
--- a/src/env.js
+++ b/src/env.js
@@ -35,7 +35,6 @@ Sk.python2 = {
     // skulpt specific
     python3: false,
     set_repr: false,
-    frozenset_repr: false,
     class_repr: false,
     inherit_from_object: false,
     super_args: false,
@@ -58,7 +57,6 @@ Sk.python3 = {
     // skulpt specific
     python3: true,
     set_repr: true,
-    frozenset_repr: true,
     class_repr: true,
     inherit_from_object: true,
     super_args: true,
@@ -109,7 +107,6 @@ Sk.configure = function (options) {
     Sk.bool_check(Sk.__future__.division, "Sk.__future__.division");
     Sk.bool_check(Sk.__future__.unicode_literals, "Sk.__future__.unicode_literals");
     Sk.bool_check(Sk.__future__.set_repr, "Sk.__future__.set_repr");
-    Sk.bool_check(Sk.__future__.frozenset_repr, "Sk.__future__.frozenset_repr");
     Sk.bool_check(Sk.__future__.class_repr, "Sk.__future__.class_repr");
     Sk.bool_check(Sk.__future__.inherit_from_object, "Sk.__future__.inherit_from_object");
     Sk.bool_check(Sk.__future__.super_args, "Sk.__future__.super_args");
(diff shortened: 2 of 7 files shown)


alexwenbj:
This PR close #1029 .

rixner:
@alexwenbj, I haven't had a chance to look over this carefully yet, but please make sure that you add tests.  They should be unit tests that go in test/unit3/test_frozenset.py.  You probably should look at the tests in CPython (https://github.com/python/cpython/tree/master/Lib/test).  We generally copy the tests into Skulpt and then comment out the ones that don't work yet because of other missing features.

alexwenbj:
@rixner That is why I didn't add unit tests:

1. All these frozenset functions are copy from set functions.In fact,I copy the set.js file totally then delete the unused functions from it.
2. I can't find the unit test for frozenset in CPython unit test directory.

alexwenbj:
@rixner Add unit test cases.

rixner:
@alexwenbj you have exposed methods to Python programs that shouldn't exist:

```
s = frozenset([1, 2, 3])
print(s)
s.frozenset_add_(4)
print(s)
```

If you need internal functions, they should just be Javascript functions, you should not wrap them with ```Sk.builtin.func```, which makes them into Python functions.  You can use ```$``` in the Javascript function name for internal-use only functions.

Also, it seems like it might be better/easier to just use ```Sk.builtin.set``` as the internal representation for frozensets.  Then in the fronzenset methods (including the constructor) you can just call the appropriate set methods.  That way you won't need to reimplement methods, you won't need to have mutation methods that shouldn't be there, etc.  This internal set won't be exposed to Python programs, so as long as you never mutate it, things should work.  Am I missing something which makes it better/necessary to replicate the behavior of sets?

Finally, I don't think you need to add a new future flag here, you should just be able to use ```Sk.__future__.python3``` to determine which representation to use.

alexwenbj:
> @alexwenbj you have exposed methods to Python programs that shouldn't exist:
> 
> ```
> s = frozenset([1, 2, 3])
> print(s)
> s.frozenset_add_(4)
> print(s)
> ```
> 
> If you need internal functions, they should just be Javascript functions, you should not wrap them with `Sk.builtin.func`, which makes them into Python functions. You can use `$` in the Javascript function name for internal-use only functions.
I think this method will not and should not be invoked by programmers, cause when you use the dir():
```
f = frozenset([1, 2, 3])
print(dir(f))
```
what you get:
`['__contains__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__iter__', '__le__', '__lt__', '__ne__', '__new__', '__repr__', '__setattr__', '__str__', 'copy', 'difference', 'intersection', 'isdisjoint', 'issubset', 'issuperset', 'symmetric_difference', 'union']`
you won't get the frozenset_add_ method. I guess the function naming way with a "_" in the end is a way like $?

> 
> Also, it seems like it might be better/easier to just use `Sk.builtin.set` as the internal representation for frozensets. Then in the fronzenset methods (including the constructor) you can just call the appropriate set methods. That way you won't need to reimplement methods, you won't need to have mutation methods that shouldn't be there, etc. This internal set won't be exposed to Python programs, so as long as you never mutate it, things should work. Am I missing something which makes it better/necessary to replicate the behavior of sets?
> 
Yes, maybe it is a better way. But about the mutation methods, see my answer above.
> Finally, I don't think you need to add a new future flag here, you should just be able to use `Sk.__future__.python3` to determine which representation to use.
This is just like the set implementation.

rixner:
@alexwenbj, I ran that program and it mutated the frozenset by adding 4 to it.  Did you run it?

alexwenbj:
> @alexwenbj, I ran that program and it mutated the frozenset by adding 4 to it. Did you run it?

What I mean is the users will not know the frozenset_add_() method if they don't read the code, because the dir() will not expose the method to them.

rixner:
Regardless of whether users will see those methods when they use ```dir```, they do not belong there.  They are internal functions so should not be exposed, and there is no advantage of doing so.  Just make them internal functions.

```
Sk.builtin.frozenset.frozenset$add = function (self, item) {
    self.v.mp$ass_subscript(item, true);
    return Sk.builtin.none.none$;
};
```

That way, you can call them normally, they won't be exposed to Python, and you can use whatever helper functions you want.  However, it also seems like functions like this are completely unnecessary and you could just put the one line of code to call ```mp$ass_subscript``` directly where you need it.

alexwenbj:
@rixner Sorry,I am not familiar with the base mechanism of the method. When I modified the `Sk.builtin.frozenset.prototype["frozenset_difference_update_"] `to `Sk.builtin.frozenset.frozenset$difference_update` and invoked it with `Sk.builtin.frozenset.frozenset$difference_update(null, new_args);`, or copied the code inside difference_update to the invocation place, the result is not correct.

rixner:
If you construct new_args in the same way, I assume you would still need to call it as ```Sk.builtin.frozenset.frozenset$difference_update.apply(null, new_args)``` to call it with an array.  But, in that case, it probably makes more sense to change it to just take an array as an argument.  Or better yet, inline the code.  I'm not sure why inlining it wouldn't work, that must just be a bug in the way you inlined it, right?

alexwenbj:
`Sk.builtin.frozenset.prototype["frozenset$update"]` definition + `Sk.builtin.frozenset.frozenset$update.func_code.apply(null, new_args);` invocation = Cannot read property 'func_code' of undefined

`Sk.builtin.frozenset.prototype["frozenset$update"]` definition + `Sk.builtin.frozenset["frozenset$update"].func_code.apply(null, new_args);` invocation = Cannot read property 'func_code' of undefined

`Sk.builtin.frozenset["frozenset$update"] = function (self, other)` definition + `Sk.builtin.frozenset["frozenset$update"].func_code.apply(null, new_args);` invocation = Cannot read property 'func_code' of undefined

`Sk.builtin.frozenset["frozenset$update"] = function (self, other) ` definition + `Sk.builtin.frozenset.frozenset$update(null, new_args);` invocation = TypeError: '<invalid type>' object is not iterable (Because in passed the array into forzenset$update() function, not arguments.)

I think, maybe let this PR open until someone can help improve it.
To the ones need the frozenset, it is a not so bad alternative solution.

alexwenbj:
@rixner I have removed all the unnecessary python methods from the code.Please review them again.

rixner:
The operators should have type checks for Python 3.  See #1053 - you should just be able to copy those checks here and add similar tests.  May as well put them in now, rather than leaving them for another pull request.

alexwenbj:
@rixner Add the test cases.

rixner:
The operator error messages all have "&" instead of the correct operator.  Otherwise, this looks good! Fix the messages and I'll merge this. Thanks @alexwenbj.

alexwenbj:
@rixner Corrected them.Thank you for your kindness and patient.

PR #1034, opened 2020-02-16
Update str.js

alexwenbj:
Modify startsWith()/endsWith().


rixner:
Thank you for the contributions!  However, please add tests that used to fail, but now pass with your changes (and ensure that they pass with Python 2 and 3).

alexwenbj:
@rixner It seems the Python itself do not check the length of string.For example:
```
s ="hello world"
print(s.startswith("hell", 0, 100))
```
print out True.

rixner:
I understand.  But, what I meant is to possibly replace this:

```
    var substr = self.v.slice(start, end);
    return new Sk.builtin.bool( substr.indexOf(prefix.v) === 0);
```

with something like this (totally untested, but should be close):

```
    if ((end - start) < prefix.v.length) {
        return Sk.builtin.bool.false$;
    }
    return new Sk.builtin.bool(self.v.indexOf(prefix.v, start) === start);
```

I believe that accomplishes the same thing without copying the string.  Something similar could be done for endswith.  I was just asking if it makes sense to pursue that approach?  As I said, I am fine with the way it is, as well.

alexwenbj:
@rixner Yeah, Thank you for your patient.I will check it when I am  not so busy.Because there are more things to do in your way, for example, maybe we should check whether the "begin" and "end" is in the normal order, etc.What my code doing now is let the JS slice() function do these things.

rixner:
That's fine!  If there are good reasons for doing it the way you have, then you should leave it.

If end is before start, though, won't the if statement catch it and return ```False```?   Which is usually correct, but I guess it might be incorrect in the case where the prefix is the empty string?

(Although it might be nice to remove that extra space that I'm now noticing between the '(' and 'substr.indexOf'.)

alexwenbj:
@rixner Yes, you are right. I fixed the bug when prefix/suffix is "" and start>end return False.And added test cases.

rixner:
@skulpt/committers I think this should be merged, as it is additional functionality and is now well tested.  Please take a look.

alexwenbj:
Let startswith/endswith accept the tuple arg.Please review again.(test cases add)

alexwenbj:
Add CPython startswith/endswith test cases.
From https://github.com/python/cpython/blob/master/Lib/test/string_tests.py

rixner:
Can we get this updated and merged?  @skulpt/committers, I approved this a long time ago...

src/str.js
```diff
@@ -628,17 +633,121 @@ Sk.builtin.str.prototype["rindex"] = new Sk.builtin.func(function (self, tgt, st
...
+        start = Sk.builtin.asnum$(start);
+        start = start >= 0 ? start : self.v.length + start;
+    }
+
+    if (end === undefined || Sk.builtin.checkNone(end)) {
+        end = self.v.length;
+    } else {
+        end = Sk.builtin.asnum$(end);
+        end = end >= 0 ? end : self.v.length + end;
+    }
```
s-cork, line 721: For most cases this is not a problem but since the error message suggests that start and end can be instances of a class with an ‘__index__‘ method it might be more correct to use misceval.isIndex and misceval.asIndex rather than checkInt and asnum

src/str.js
```diff
@@ -628,17 +633,121 @@ Sk.builtin.str.prototype["rindex"] = new Sk.builtin.func(function (self, tgt, st
...
+
+    //take out the substring
+    var substr = self.v.slice(start, end);
+
+    if(Sk.abstr.typeName(suffix) == "tuple"){
+        var tmpBool = false, resultBool = false;
+        var it, i;
+        for (it = Sk.abstr.iter(suffix), i = it.tp$iternext(); i !== undefined; i = it.tp$iternext()) {
+            if(start > end){
+                tmpBool = start <= 0;
```
s-cork, line 735: This seems to be a one time check. Could we move this logic outside of the loop and return earlier if this check is false. Then there wouldn’t be a need for tmpBool

alexwenbj: Yes,you are right.I have modified it.Thank you.

albertjan:
Thanks @alexwenbj

PR #1038, opened 2020-02-18
fix #1032 :memo:

### LINKED ISSUE #1032: str functions startswith() and endswith() take more than one arguments
For example,in Skulpt:
```
s = "hello world!"
print(s.endswith("!", 0, 5))
```
will raise an error,but it is OK in python.

Dustyposa:
add two arguments for endswith and startswith
Please review the code.
It's necessary to create function for reduce the duplicated code?
:smile:


rixner:
Please look at #1034

Dustyposa:
got it

PR #1039, opened 2020-02-18
fix #1035 :memo:

### LINKED ISSUE #1035: a small bug for str format
When format a string and set align left:
`print('%-10s' % ('hello'))  # align left`
it will print out a string:
` hello    `
there is a space char in front of the 'hello',and only 4 space chars behind the 'hello'.It is OK with format() function and f-strings.

Another small bug:
`print("The temperature today is %d degrees outside !" % ( 23.4 ))`
should convert the 23.4 to 23, but Skulpt do not do it.

CC:  @meredydd Maybe you can fix these. :-)

Dustyposa:
fix `%d` and `%-10s` format


Dustyposa:
emm
has any one can review this code?

src/str.js
```diff
@@ -1040,7 +1046,11 @@ Sk.builtin.str.prototype.nb$remainder = function (rhs) {
             }
             return prefix + r;
         };
-
+        throwFormatError = function (res, formatType, baseData) {
```
albertjan: why do you have this separate function?

src/str.js
```diff
@@ -1056,7 +1066,11 @@ Sk.builtin.str.prototype.nb$remainder = function (rhs) {
         }
         base = 10;
         if (conversionType === "d" || conversionType === "i") {
-            return handleWidth(formatNumber(value, 10));
+            let tmpData = formatNumber(value, base);
```
albertjan: wouldn't it make more sense to call `.toFixed(0)` on this? instead of doing the `slice` below?

albertjan:
two small questions 😄

src/str.js
Dustyposa: Oh, I merged it.

src/str.js
Dustyposa: `tmpData ` is a `string array`, if want to use `toFixed(0)` need to make the sting to be a number.
Like use `parseInt, parseFloat, Number etc`.. Normally, it's work fine for small number. But for big  number which like '"123456789127123546890"` , that's can't work. So i use slice to operate the `string`

albertjan: Alright I hadn't thought of that

Dustyposa: Yes, at the beginning I used `parseInt etc`, but can't pass all tests.
So I decided use this way.
:smile:

Dustyposa: got it.

s-cork: looks good.

albertjan:
Thanks @Dustyposa

PR #1041, opened 2020-02-18
Load parent package before checking cache for children.

meredydd:
Some packages (such as Plotly) automatically load their children
(eg `import plotly` causes `plotly.graph_objs` to be imported automatically).

This means that we need to check the `sys.modules` cache *after* loading
the parent module, not before - otherwise we can end up trying to load the
child module twice, which can break things (such as...Plotly).

This changes a handful of lines of code, and makes a big indentation change.
`git diff -w` will be much more helpful than GitHub's built-in diff! I'll paste
its output into a comment.


meredydd:
Here's the output of `git diff -w`, to help with review:

```
meredydd@biggles:~/src/skulpt$ git diff -w HEAD^ | cat
diff --git a/src/import.js b/src/import.js
index c488454..739bfda 100644
--- a/src/import.js
+++ b/src/import.js
@@ -188,19 +188,6 @@ Sk.importModuleInternal_ = function (name, dumpJS, modname, suppliedPyBody, rela
 
     modNameSplit = name.split(".");
 
-    // if leaf is already in sys.modules, early out
-    try {
-        prev = Sk.sysmodules.mp$subscript(modname);
-        // if we're a dotted module, return the top level, otherwise ourselves
-        if (modNameSplit.length > 1) {
-            return Sk.sysmodules.mp$subscript(absolutePackagePrefix + modNameSplit[0]);
-        } else {
-            return prev;
-        }
-    } catch (x) {
-        // not in sys.modules, continue
-    }
-
     if (modNameSplit.length > 1) {
         // if we're a module inside a package (i.e. a.b.c), then we'll need to return the
         // top-level package ('a'). recurse upwards on our parent, importing
@@ -211,12 +198,22 @@ Sk.importModuleInternal_ = function (name, dumpJS, modname, suppliedPyBody, rela
     }
 
     ret = Sk.misceval.chain(topLevelModuleToReturn, function(topLevelModuleToReturn_) {
+        topLevelModuleToReturn = topLevelModuleToReturn_;
+
+        // if leaf is already in sys.modules, early out
+        try {
+            prev = Sk.sysmodules.mp$subscript(modname);
+            // if we're a dotted module, return the top level, otherwise ourselves
+            return topLevelModuleToReturn || prev;
+        } catch (x) {
+            // not in sys.modules, continue
+        }
+
+        return Sk.misceval.chain(undefined, function() {
             var codeAndPath, co, googClosure;
             var searchFileName = name;
             var result;
 
-        topLevelModuleToReturn = topLevelModuleToReturn_;
-
             // If we're inside a package, look search using its __path__
             if (modNameSplit.length > 1) {
                 if (!topLevelModuleToReturn) {
@@ -397,6 +394,7 @@ Sk.importModuleInternal_ = function (name, dumpJS, modname, suppliedPyBody, rela
             // otherwise we return the actual module that we just imported
             return module;
         });
+    });
 
     return canSuspend ? ret : Sk.misceval.retryOptionalSuspensionOrThrow(ret);
 };
```

alexwenbj:
Does it mean when I import one module, it will load the sub-modules automatically?For example, `import pygame`,need import all the sub-modules automatically.

ebertmi:
You can also ignore white space in github diffs using `?w=1`

https://github.blog/2018-05-01-ignore-white-space-in-code-review/

rixner:
@meredydd I think the only outstanding "issue" is whether it is feasible to add a test.  Given that both @albertjan and I have approved this, I'm fine with you merging it yourself either with or without the test once you make a decision on that (despite our "ban" on merging your own pull requests :smile:).

alexwenbj:
@meredydd Hi,I have a question relate to this issue but not about this issue: Can we use the plotly.js module directly in Skulpt,or should we wrap it in skulpt way by ourselves?

albertjan:
Merging this one, has been open with approval for a while 😄

PR #1042, opened 2020-02-19
fix __dict__ self-referencing __dict__ self-referencing __dict__ ...

s-cork:
fixes this issue
```python
>>> class A:
...    pass
...
>>> a = A()
>>> a.x = 1
>>> a.__dict__
{'__dict__': {...}, 'x' : 1}
>>> a.__dict__["__dict__"]
{'__dict__': {...}, 'x' : 1}
>>> a.__dict__ is a.__dict__["__dict__"]
True
>>> d = a.__dict__
>>> for i in range(1000000): 
...     print(d["__dict__"])
...     d = d["__dict__"]
...
!@#
```

this pr aims to ensure that `a.__dict__` is not self-referencing.

---
It solves the issue in the following way.
removes the line that caused the self-referencing - effectively -  `this.$d["__dict__"] = this.$d`

adds a shortcut so that if `x.__dict__` is called then `x.$d` is returned

---

I know that this is a hot function so the solution might not be favourable. It simply aimed to ensure that `y.__dict__` always returns `this.$d` without `"__dict__"` needing to be a key inside `this$d`

---

My first solution (3e5d6f7) initially failed with objects that were copied because of how GenericGetattr prioritised its lookup...

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -63,7 +63,7 @@ Sk.builtin.object.prototype.GenericGetAttr = function (pyName, canSuspend) {
         }
         if (res !== undefined) {
             return res;
-        } else if (jsName == "__dict__") {
+        } else if (jsName == "__dict__" && dict instanceof Sk.builtin.dict) {
             return dict;
         }
     }
--- a/src/object.js
+++ b/src/object.js
@@ -63,6 +63,8 @@ Sk.builtin.object.prototype.GenericGetAttr = function (pyName, canSuspend) {
         }
         if (res !== undefined) {
             return res;
+        } else if (jsName == "__dict__") {
+            return dict;
         }
     }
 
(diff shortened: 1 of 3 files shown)


s-cork:
Looks like this will need a bit of work... not sure what I broke...

s-cork:
unintended consequence:

```
>>> x = list()
>>> x.__dict__
{'__bases__' : (,), '__mro__': <class 'list'>}
```

This is because of the function `setUpClass` in import.js which adds a `this.$d` dict object to all builtin objects...

see comment in #1049

s-cork:
@meredydd  Is there a reason that in the genericGetAttr function that we use the 

```
dict = this["$d"] || this.constructor["$d"];
```
If I run all the tests with 
```
dict = this["$d"];
```
it still works??


alternatively could only return the dict if `this.$d` actually belongs to the instance and not the contructor... 

something like:

```javascript
else if (jsName == "__dict__" && dict instanceof Sk.builtin.dict && dict === this["$d"]) {
            return dict;
        }
```

this would prevent instances of builtins having a `__dict__` property...

PR #1044, opened 2020-02-20
Add big integer library and clean up (and mess up!) the code base to do so

rixner:
Disclaimer: I am not sure this should be merged.  There are a few things that are ugly, and there must be some better way of accomplishing them!

This pull request incorporates several things:

1. It replaces ```Sk.builtin.biginteger``` with the npm ```big-integer``` library.  The major advantage of this is that it will use native ```BigInt``` when it is available.  This is the motivation for this PR.

2. It cleans up a bunch of closure compiler type errors.  It turns out that we have not been letting the closure compiler type check the code for a while.  ```biginteger.js``` had a directive to suppress type checking for the file.  However, webpack consolidates the code into a single "file" before passing it to the closure compiler, so the compiler was suppressing type checking for everything.  This is likely a strict improvement.  However, there were a few cases where I couldn't get things to type check, so I loosened the typing.  That's arguably a step back, but again, things weren't being type checked at all, before.  I don't think there were actual bugs, rather the type inference wasn't able to guarantee the desired type.

3. It fixes a few actual errors.  There were references to symbols that just plain do not exist.  Presumably that code was never executing, or it wasn't working correctly if it was.  Some of it is fixed, some of it was deleted (since it wasn't doing anything any way and didn't belong).  This is likely a strict improvement.  For example, look at https://github.com/skulpt/skulpt/blob/51bf1a386de5a7590143b19a3f8aca7ce1fd79a5/src/ast.js#L1499 
This can't possibly work.  ```TOK``` is ```Sk.token.tokens```, which is simply an object mapping token names to numbers.  There are no functions.  ```T_NEW_TYPE_COMMENT``` is not even one of the keys in the object.

4. It does not pass external libraries through the closure compiler.  The libraries in npm and elsewhere are generally incompatible with closure, as they have all kinds of type errors in them and other "problems". This PR uses webpack to split the build into two bundles (```vendor.js``` and ```internal.min.js```).  The internal code is run through the closure compiler and the vendor code is just bundled by webpack for use by the internal bundle.  This is just to get the vendor code in there without closure getting mad about it, so I view this as neither positive nor negative.

5. The vendor code is minified using Babel Minify.  This is potentially problematic.  If the vendor code uses features in a more recent version of Javascript than you are trying to build, then they will get though and end up in skulpt.  It does not force ES3/ES5/whatever compliant code to be output, like we have the closure compiler doing.  (We have the same problem with the libraries, so it is unclear how big of a deal this is, but it is unsatisfying at a minimum.)

6. I have an ugly, horrible hack in ```webpack.config.js``` to modify the code inserted by webpack into the bundles.  It changes the internal code to make it closure friendly.  And both the internal and vendor code to make them compatible with both node and browsers.  If webpack changes it's injected code, this hack would need to be modified to deal with it.  This is strictly bad in my opinion.  I couldn't find another way, though.

7. It merely concatenates the final minified vendor bundle and internal bundle into ```skulpt.min.js```.  One consequence of this is that I think the source maps are rendered useless.  I think it also likely requires people to test the validity of the final bundle across machines and browsers.

I tried to separate the closure fixes into independent commits where the same type of fixes were consolidated into a single commit.  In a hope to make it easier to understand what I did and catch errors.

As I said initially, I'm not really sure this should be merged.  @albertjan and @meredydd I would appreciate a careful look at this and some thought about what should be merged, what should not be merged, and whether there are better approaches to fixing the underlying problem that we need to suppress closure warnings/errors on vendor code, but make sure all internal code is checked.


rixner:
I should add that I think that the best/right way to deal with the third party libraries is to continue to pass them through the closure compiler with everything else, but to tell the compiler to ignore issues in those files (that we know not to be a problem).  I just wasn't able to figure out how to get that to work.  Presumably one could modify the library code to make it happen, but that also seems less than desirable.

src/float.js
```diff
@@ -2,6 +2,8 @@
  * @namespace Sk.builtin
  */
 
+const biginteger = require("big-integer");
```
albertjan: Lets call this BigInt so we only have to remove this
line in the future.

webpack.config.js
```diff
@@ -31,9 +31,37 @@ module.exports = (env, argv) => {
...
+            include: /^skulpt\.min\.js$/,
+            algorithm: 'gzip'
+        }),
+    ];
+
 
     if (argv.mode === 'production') {
         opt = {
+            splitChunks: {
+                cacheGroups: {
```
albertjan: Does this mean the bitinterger.js is split off into vendor.js?

src/ast.js
```diff
@@ -212,9 +215,9 @@ function setContext (c, e, ctx, n) {
         case Sk.astnodes.Compare:
             exprName = "comparison";
             break;
-        case Sk.astnodes.Repr:
-            exprName = "repr";
-            break;
+        // case Sk.astnodes.Repr:
```
albertjan: Do we not need this for backwards compat with py2

src/ast.js
```diff
@@ -572,29 +572,12 @@ function ast_for_with_stmt(c, n0, is_async) {
...
     if (is_async) {
-        return new Sk.astnodes.AsyncWith(items, body, LINENO(n0), n0.col_offset);
+        return new Sk.astnodes.AsyncWith(items, body, LINENO(n0), n0.col_offset, n0.end_lineno, n0.end_col_offset);
     } else {
-        return new Sk.astnodes.With(items, body, LINENO(n), n.col_offset);
+        return new Sk.astnodes.With(items, body, LINENO(n), n.col_offset, n.end_lineno, n.end_col_offset);
     }
 }
 
-function astForExecStmt (c, n) {
```
albertjan: Ah sorry for leaving this. We used oldStyleNaming and new_style_naming when moving the ast code so we could see which functions we had done. 😊

test/unit3/test_counter.py
```diff
@@ -3,6 +3,39 @@
 import collections
 
 class CounterTests(unittest.TestCase):
+    def _mc_compare(self, result, expected):
```
albertjan: Thanks! I feel we explained often enough that the sorting algorithms in node change between versions. 😊

src/ast.js
```diff
@@ -334,12 +337,9 @@ function seq_for_testlist (c, n) {
     var seq = [];
     Sk.asserts.assert(n.type === SYM.testlist ||
         n.type === SYM.testlist_star_expr ||
-        n.type === SYM.listmaker ||
```
albertjan: Which version would of python did you get this from? Lets change the name to be snake case too.

webpack.config.js
rixner: Yes.  biginteger, strftime, strptime, and setimmediate all end up in vendor.js.  None of them can get through the closure compiler.

src/ast.js
rixner: I don't know.  I do know that it is not defined in Python 2 or Python 3, so if we do "need" it, it wasn't being used before this pull request.  Removing it changed nothing in the code base, as it was dead code that could never execute.

test/unit3/test_counter.py
rixner: Yes, the tests were clearly broken and should have been fixed like this a long time ago. :smile:

src/ast.js
rixner: I do not know where any of that came from...

rixner:
@albertjan are you sure you wanted to approve this?  Again, points 5-7 are potentially problematic.  And maybe very problematic...

src/float.js
rixner: I don't think that will work the way you are proposing.  The "biginteger" object doesn't have the same interface as "BigInt".  I feel that calling it BigInt would be problematic because the rest of the code would look broken, as it would be calling methods that native BigInts don't have...  And then when we do go to native BigInts, we would still have to change all the code relating to big integers.

albertjan: I was under the impression it was a drop in replacement

src/ast.js
albertjan: Yeah you're right, I think it was left inplace because the python 2 ast/compiler does something special for repr. I'll try and find out. but it seems indeed best to comment this out for now.

src/ast.js
albertjan: @meredydd what do you think are we ever going to support the py2 `exec` statement?

src/ast.js
```diff
@@ -1447,7 +1442,7 @@ function ast_for_funcdef_impl(c, n0, decorator_seq, is_async) {
     var name_i = 1;
     var end_lineno, end_col_offset;
     var tc;
-    var type_comment = NULL;
+    var type_comment = new Sk.builtin.str("");
```
albertjan: was this to placate closure?

src/ast.js
albertjan: we should port this from https://github.com/python/cpython/blob/3.7/Python/ast.c#L1186-L1213 can you add a comment to the source. it doesn't make sense to do it in this PR but we're probably missing some edge cases 😄

albertjan:
How does the parser / tokeniser know about the column positions? or are they just filled with undefined until we update them?

src/float.js
rixner: It's not possible to build a polyfill that is a drop in replacement.  With native BigInts, you add/subtract/etc. with the "+", "-", etc. operators.  They can't be overloaded.  So, there are two realistic options for dealing with browsers that don't support BigInt:

1) Using a library like this that has a function-based API for big int operations, where they internally use native BigInts if/when they can.  The code looks different than it would if it solely used native BigInts, but the performance should be close on systems that support BigInt.

2) Write the source code using native BigInts and use a compiler to convert the appropriate operations to library calls.  This is known to not work in a variety of cases.  The only advantage being that when you switch to all native BigInts, the code doesn't need to change.  But, the correctness problems in the interim would seem to be enough of a deal breaker.

src/ast.js
rixner: Yes, but this code never worked, so I don't know if any of it is correct at all.  This is just a guess at what is supposed to be here.  It probably breaks something else, but I don't know what.  Do you have any tests that actually exercise this code? There's nothing in our test suite that does.  And the few things I tried didn't actually end up going through here.

So, I probably moved this sideways.  From not working at all, to passing closure, but probably still not working at all.

rixner:
I didn't look into where the column positions come from.  But, they are there (at least in the places that I looked at).  I didn't feel the need to investigate how they got there... :)

src/ast.js
rixner: I'm not sure what you're asking for.  This is correctly handling "testlist", I think.  It just also had a check for "listmaker", which doesn't exist anywhere.  It's not in https://github.com/python/cpython/blob/3.7/Python/ast.c anywhere, either.

I'm fine with adding a comment if there is something missing, but I'm not sure what that comment should be...

albertjan: I didn't mean the listmaker specifically I meant that we should make sure this whole function is aligned with the py3.7 version.

test/unit3/test_counter.py
s-cork: tagging @bnmnetp who just asked a related question in #1066

rixner:
I converted this to a draft because ultimately, I'm not comfortable with the solution to exclude third party code from the closure compiler. The changes throughout our codebase, though, are valuable.  So, I'm not sure what to do with this at this point.

meredydd:
Tidying up: We now have big integers implemented, using JSBI

Issue #1046, opened 2020-02-22
builitin inheritance not recognised


s-cork:
I noticed this when I was working with Error handlers but it's true of all builtins

```python
>>> issubclass(TypeError, Exception)
False # should be true
>>> issubclass(int, object)
False # should be true
```

PR #1048, opened 2020-02-22
assertRaises as a context manager and kwarg support

### LINKED ISSUE #1027: contexts for assertRaises
Many python tests use the syntax `with self.assertRaises(ValueError):` but this leads to the error message:

```
NotImplementedError: assertRaises does currently not support assert contexts
```
It is not always easy to adapt the test without contexts since tests that use contexts usually have kwargs, and kwargs can't be passed to `assertRaises`. 

It would be useful to have this feature for fewer commented out python tests.

### LINKED ISSUE #1047: BaseException
All exceptions inherit from BaseException but BaseException is not a valid name in skulpt 

```
NameError: name 'BaseException' is not defined
```

I think it's just a case of adding it to the `builtindict.js`

s-cork:
It is common to have to comment out `with self.assertRaises` or any `self.assertRaises` that has `kwargs` from Python unit tests 😢 . 

This pull request aims to close #1027, it:
- adds context manager support to `assertRaises`. 
- allows `kwargs` for `assertRaises`
- fixes the `repr` when an `assertRaises` test fails - currently you just get `Fail:` since `str(TypeError())` is empty
- uses code from cpython as a template and adapts to the Skulpt unittest module.

I uncommented a few of the `test_math.py` that included `with self.assertRaises` - I wasn't sure what other tests to add...?

### DIFF
--- a/src/lib/unittest/__init__.py
+++ b/src/lib/unittest/__init__.py
@@ -6,11 +6,16 @@ the unittest module from cpython.
 '''
 
 
-class _AssertRaisesBaseContext:
-
+class _AssertRaisesContext:
+    """A context manager used to implement TestCase.assertRaises* methods."""
     def __init__(self, expected, test_case):
         self.test_case = test_case
         self.expected = expected
+
+    def _is_subtype(self, expected, basetype):
+        if isinstance(expected, tuple):
+            return all(_is_subtype(e, basetype) for e in expected)
+        return isinstance(expected, type) and issubclass(expected, basetype)
 
     def handle(self, args, kwargs):
         """
@@ -20,6 +25,8 @@ class _AssertRaisesBaseContext:
         arguments.
         """
         try:
+            if not self._is_subtype(self.expected, BaseException):
+                raise TypeError('assertRaises() arg 1 must be an exception type or tuple of exception types')
             if not args:
                 return self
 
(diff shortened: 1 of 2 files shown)


s-cork:
```bash
Fail: Expected <type 'StandardError'> but got <type 'AttributeError'>
Fail: Expected <type 'StandardError'> but got <type 'AttributeError'>
Fail: Expected <type 'StandardError'> but got <type 'AttributeError'>
Fail: Expected <type 'StandardError'> but got <type 'AttributeError'>
Tests failed in  TestErrorHandlingPython.test_get_only  
Fail: Expected <type 'StandardError'> but got <type 'TypeError'>
Fail: Expected <type 'StandardError'> but got <type 'TypeError'>
Fail: Expected <type 'StandardError'> but got <type 'TypeError'>
Fail: Expected <type 'StandardError'> but got <type 'TypeError'>
Tests failed in  TestErrorHandlingPython.test_len_only  
Ran 5 tests, passed: 3 failed: 2
```

A couple more tests fail because of #1046

src/lib/unittest/__init__.py
```diff
@@ -186,26 +240,13 @@ def appendResult(self,res,actual,expected,feedback):
...
-            actualerror = str(inst)
-            print("ACT = ", actualerror, str(exception()))
-        else:
-            actualerror = "No Error"
-
-        self.appendResult(res, str(exception()), actualerror, "")
+            return context.handle(args, kwargs)
+        finally:
+            # bpo-23890: manually break a reference cycle
+            context = None
```
meredydd: This is an odd-shaped invocation. Are you copying code from Python mainline, and if so where/which version? (I'm guessing you are, because reference cycles aren't a thing with a proper GC like Javascript has)

s-cork: yeah - the code is from https://github.com/python/cpython/blob/3.7/Lib/unittest/case.py#L757


There's also a similar invocation in the `_AssertRaisesContext `
```
        finally:
            # bpo-23890: manually break a reference cycle
            self = None
```
https://github.com/python/cpython/blob/3.7/Lib/unittest/case.py#L179

I wasn't sure about them - so I just left them in...

s-cork:
> It's possible we should just make the inheritance correct in `error.js`

I think the inheritance setup in the error.js is correct... I think maybe all builtins don't know about their parent classes... which seems like a separate issue...
```python
>>> issubclass(int, object)
False
>>> class C(int):
...     pass
>>> issubclass(C, int)
True
>>> issubclass(C, object)
False
```

s-cork:
re inheritance issue - side note - `error.js` inheritance is correct from a Python2 standpoint but includes `StandardError` as part of the inheritance tree which is removed in Python 3...

s-cork:
Please consider committing #1049 before this commit - then I will tidy up this commit.

`issubclass` will then work on exception classes correctly (and i'll remove the unnecessary code that checks if the expected_exception was Exception). 

I will also be able to use `issubclass` to check that an exception class really was passed to `assertRaises` using `issubclass(expected_exception, BaseException)`.

s-cork:
this is now ready. I was able to update the code now that builtin inheritance works as expected.

PR #1049, opened 2020-02-24
Builtin Inheritance

### LINKED ISSUE #1046: builitin inheritance not recognised

s-cork:
This fixes #1046. 

I believe the algorithm for `setUpClass` was not quite right for builtins as it checked `child.tp$base` but the actual base class is located at `child.prototype.tp$base`

This effectively meant that all builtins had an attribute `__bases__ = ()` 

I added some failing tests. 

My first attempt failed because I unknowingly included `sk$abstract` classes in `bases`. The final commit corrected this, adding a check to ensure that a potential parent did not have `sk$abstract = true` before pushing it to bases. 

___
While I'm here I add a `.__name__` atribute to all builtins. 
And fix the `.__mro__` property
___
**Example fixes:**

```python
>>> issubclass(int, object)
True  # previously False
>>> int.__bases__
(<class 'object'>, ) # previously empty tuple
>>> int.__mro__
(<class 'int'>, <class 'object'>) #previously <class 'int'>
>>> int.__name__
'int' # previously AttributeError
>>> class F(int):
...  pass
...
>>> F.__bases__
(int,)
>>> issubclass(F, int)
True
>>> issubclass(F, object)
True # previously False

```

### DIFF
--- a/src/builtindict.js
+++ b/src/builtindict.js
@@ -29,6 +29,7 @@ Sk.builtins = {
     "any"       : new Sk.builtin.func(Sk.builtin.any),
     "all"       : new Sk.builtin.func(Sk.builtin.all),
 
+    "BaseException"      : Sk.builtin.BaseException, 
     "AttributeError"     : Sk.builtin.AttributeError,
     "ValueError"         : Sk.builtin.ValueError,
     "Exception"          : Sk.builtin.Exception,
--- a/src/import.js
+++ b/src/import.js
@@ -70,12 +70,14 @@ Sk.doOneTimeInitialization = function (canSuspend) {
     // Register a Python class with an internal dictionary, which allows it to
     // be subclassed
     var setUpClass = function (child) {
-        var parent = child.tp$base || child.prototype.tp$base;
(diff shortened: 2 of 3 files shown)


s-cork:
a slight issue with the approach in this code (and an issue before the changes)
all instances of the builtins get the elements from the child.$d... 

in Python all of these should raise an exception... 
```python
>>> x = list()
>>> x.__bases__
(<class 'object'>,) #should raise exception
>>> x.__mro__ 
(<class 'list'>, <class 'object'>) #should raise exception 
>>> x.__name__ #this is new... 
list #should raise exception
```
That said it's the same issue for standard classes

```python
>>> class F:
...     pass
...
>>> f = F()
>>> f.__bases__ 
(<class 'object'>,) #should raise exception
```

albertjan:
merging this I consider this a small change

PR #1053, opened 2020-02-27
Type checks for Python 3 in set operators.

rixner:
Also clean up set construction and iteration.


albertjan:
calling this a small change merging

PR #1054, opened 2020-02-28
Add sq$asarray method to lists and tuples

rixner:
The ```sq$asarray``` method returns a copy of the internal javascript array.  This simplifies and improves the performance of common operations where you internally need a copy of a list or tuple.  Instead of using a Python iterator to access them, this allows you to simply and easily get a javascript array.

I took this opportunity to somewhat regularize how lists/tuples/sets/frozensets are constructed.

As I was doing this, I did some additional list cleanup as I came across things.


s-cork:
hi @rixner I like this this a lot. I wonder if `sk$asarray` or even just `sk$array` is a better name for the flag since it's a skulpt only implementation and not pulled from Cpython. 

I'll merge the idea into what I'm working on in #1092 (you'll pleased to hear i'm getting some performance benefits there too!)

rixner:
@s-cork I was trying to capture that it was for sequences, but I'm fine with ```sk$asarray```...

Not sure what to do about overriden iter methods.  As you've already discovered, subclassing builtins is a little bit perilous to begin with in Skulpt.

s-cork:
I was thinking about adding a flag `klass.prototype.hp$type = true` for all dynamically create klasses. Which would match cpython. (and also give a nice check for whether the attribute `__class__` can be changed or not. currently using the flag on the `getset_descriptor` for `__class__` in #1092 ). 

then in this pr you could do something like
```
if (this.sk$asarray && this.hp$type === undefined) {...}
```

then you still the speed win for builtin types and
skulpt users that build their own native types in javascript can use the same flag

and dynamically created python classes that inherit from a builtin just use the `tp$iternext` architecture

s-cork:
what about this as a helper function.

```javascript
Sk.abstr.arrayFromIterable = function (iterable, canSuspend) {
    if (iterable == null) {
        // if we get an undefined or null value return an empty array
        return [];
    }
    if (iterable.hp$type === undefined && iterable.sk$asarray !== undefined) {
        return iterable.sk$asarray();
    }
    const L = [];
    const ret = Sk.misceval.iterFor(Sk.abstr.iter(iterable), (i) => {
        L.push(i);
    });
    if (ret === undefined) {
        return L;
    } else if (canSuspend) {
        return Sk.misceval.chain(ret, () => {
            return L;
        });
    } else {
        Sk.misceval.retryOptionalSuspensionOrThrow(ret);
        return L;
    }
};
```

This has the nice benefit of being suspendable and throwing the generic error message  

```javascript
"'" + Sk.abstr.typeName(obj) + "' object is not iterable"
``` 
from `Sk.abstr.iter`

and then the set constructor could do something like:

```javascript
if (!Array.isArray(S)) {
      S = Sk.abstr.arrayFromIterable(S)
}
len = S.length;
for (i = 0; i < len; i++) {
    this.v.mp$ass_subscript(S[i], true);
}
```

rixner:
I was trying not to build an intermediate array.  You would now be iterating over it twice.  Once to create the array, and then once to consume it.  If the sequence is large, this is going to waste time and be worse than the current situation.  Granted the second iteration is over a javascript array, so is much cheaper, but it's still unnecessary.

The point of that function was to return an array in situations where it already existed, so was cheap to do so, thereby eliminating an unnecessary iteration (and all the iteration machinery just to get the elements, which were already in a javascript array that could easily be iterated over).

I think it's better to only have the asarray function on true builtins and somehow know not to use it on subclasses.

s-cork:
good point. I guess `set` wouldn't be a good use case for such a helper function.

rixner:
@skulpt/committers what do I need to do in order to get this accepted?

albertjan:
I'm happy with this.

s-cork:
I agree - LGTM

rixner:
Before it gets pulled, should I change it to ```sk$asarray``` or leave it as ```sq$asarray```?  I think @s-cork 's point that it should have an ```sk$``` prefix is valid.

PR #1062, opened 2020-04-13
fix #928

### LINKED ISSUE #928: customize input() alert fail
Can't customize the input（）alert  by this method?https://github.com/skulpt/skulpt/issues/685
config:
```
 Sk.configure({
            output: handleOutput,
            read: builtinRead,
            __future__: Sk.python3,
            execLimit: 60000,
            killableWhile: true,
            killableFor: true,
            timeoutMsg: () => '运行超时',
            nonreadopen: true,
            filewrite,
            fileopen,
            inputfun: showInput,
            inputfunTakesPrompt: true,
        });
showInput:

function showInput(prompt) {
    return new Promise((resolve, reject) => {
        resolve('1000');
        // Prompt({
        //     title: prompt,
        //     onOk: value => {
        //         resolve(value);
        //     },
        // });
    });
}
when run ,it caused err: valueError unknow

### PR DESCRIPTION
**Pull request by viewerw**
I know why I can't customize input
Promise in my project achieve by promise-polyfill , so it's not an instance of window.Promise ,so add  typeof x.then==='function' can fix it.

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -1359,7 +1359,7 @@ Compiler.prototype.craise = function (s) {
         // for the Python version you're using.
 
         var instantiatedException = this.newBlock("exception now instantiated");
-        var isClass = this._gr("isclass", exc + " instanceof Sk.builtin.type || " + exc + ".prototype instanceof Sk.builtin.BaseException");
+        var isClass = this._gr("isclass", exc + ".prototype instanceof Sk.builtin.BaseException");
         this._jumpfalse(isClass, instantiatedException);
         //this._jumpfalse(instantiatedException, isClass);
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/file.js [hunk 1]
```diff
@@ -159,7 +159,7 @@ Sk.builtin.file.$readline = function (self, size, prompt) {
 
         x = Sk.inputfun(lprompt);
 
-        if (x instanceof Promise) {
+        if (x instanceof Promise || typeof x.then === 'function') {
             susp = new Sk.misceval.Suspension();
```
s-cork, line 163: can we fix the double quotes which is required to pass Travis

s-cork:
@viewerw are you able to update this pr?

src/file.js
s-cork: @viewerw - can we change this to `(x && typeof x.then === "function")`

PR #1070, opened 2020-04-15
fix dir/getattr/setattr for reservedNames

s-cork:
This pr aims to fix a few bugs with mangled words and names. 

Currently there are the following APIs for fixing `reservedWords_` and `reserervedNames_`
```
Sk.fixReservedWords 
Sk.fixReservedNames
Sk.unfixReserved
```
I've added a method `Sk.fixReserved` which does `fixReservedNames(fixReservedWords(name))`.
_(it seemed like a sensible API since unfixReserved already existed which does the opposite)_

I replace adjacent calls to `fixReservedWords` and `fixReservedNames` with the new API `fixReserved`.

There are some places in `compile.js` and `symtable.js` where only one of `fixReservedWords` or `fixReservedNames` is called (or they are not called adjacently). I leave these as they were. 

---

prior to this pull request `getattr` and `setattr` only called `fixReservedWords`
which is problematic for cases like this
```python
>>> class A:
...     pass
...
>>> a = A()
>>> a.name = 'foo'
>>> a.name
'foo'
>>> getattr(a, 'name')
'klass'
>>> a.length = 'bar
>>> a.length
'bar'
>>> getattr(a, 'length')
2 
>>> settattr(a, 'watch', 'foo')
>>> a.watch
AttributeError('A' object has no attribute 'watch_$rn$')
>>> getattr(a, 'watch')
'foo'
>>>  # reserved words like default are already mangled by getattr and setattr
... a.default = 'bar' 
>>> a.default
'bar'
>>> getattr(a, 'default')
'bar'
```

This pr also fixes the above `AttributeError` message by calling `unfixReserved` before throwing the `AttributeError` -  the above error message would now read
```python
AttributeError('A' object has no attribute 'watch')
```

---

The dir does not show any mangled `js_names` since the algorithm ignored any attribute with the `$` symbol. 
this pr calls `unfixReserved` for each `name` in the `dir`. 

prior to this pull request
```python
>>> 'name' in dir(a)
False   # :-(
```

---

`arguments`, `prototype` and `caller` were added to the `reservedNames_` fixing the following issue

```python
>>> class A:
...     pass
...
>>> A.arguments
ExternalError - Cannot read property '$isSuspension' of null
>>> A.prototype
<__main__.A object>
>>> A.caller
ExternalError - Cannot read property '$isSuspension' of null
```


s-cork:
@meredydd - this pr is ready for review

s-cork:
An interesting quirk for `reservedNames_` is that it could be reduced to the following (I think):
```
var reservedNames_ = {
    "apply": true,
    "arguments": true,
    "call": true,
    "caller": true, 
    "eval": true,
    "__noSuchMethod__": true,
    "prototype": true,
    "toSource": true,
    "unwatch": true,
    "watch": true,
    "length": true,
    "name": true,
};

```

since the `fixReservedNames` does:
```
function fixReservedNames (name) {
    if (reservedNames_[name]) {
        return name + "_$rn$";
    }
    return name;
}
```

So things like
```
fixReservedNames ("constructor") --> "constructor_$rn$"
fixReservedNames ("toString") --> "toString_$rn$"
```
already happen...

meredydd:
The only thing that makes me hesitate to declare this a "small change" and merge it immediately is that this change is visible in Python code (because of the `__getattribute__` stuff described earlier).

I still think it's worth it, though, so I'm going to leave it a day or two in case any of the other @skulpt/committers object.

rixner:
I feel like the reserved names handling has always been a weak point.  One of the major problems is that its very hard to test it.  As it has to continue to work across the diversity of Javascript implementations that people run Skulpt on.  Are we certain this is strictly an improvement across the board and doesn't create regressions on some Javascript version?

s-cork:
This commit makes no aim at changing the underlying approach (I hope)

From what I could tell 
- Skulpt already mangles `reservedWords_` and `reservedNames_` in compile code
- when a python string is passed to `Sk.builtin.getattr` and `Sk.builtin.setattr`, Skulpt already mangles `reservedWords_` 

This PR primary aim is to add mangling of `reserevedNames_` to `Sk.buitin.getattr` and `Sk.builtin.setattr` 

This was achieved in the first commit https://github.com/skulpt/skulpt/pull/1070/commits/62110302db799b98ac4760aeb532509cd61bfc71 in `Sk.builtin.js` with two lines of code at line 858 and 880.

---
The other changes are there to:
- unmangle a string before displaying an attribute error
- unmangle a string before checking if it should be added to the dir
- use the new API where Skulpt was mangling `reservedWords_` and `reservedNames_` in subsequent lines of code.
---

albertjan:
There aren’t any changes that would affect the supported Javascript versions. I also don’t see an obvious way to do this in another way. And it does solve a problem we all have run into. 

I think it makes sense to merge this.

PR #1073, opened 2020-04-17
Tokenize module

timmartin:
This is a rudimentary implementation of the Python `tokenize` module. It doesn't deal with the strings / bytes distinction, but other than that the API should match the Python API where it is implemented (I haven't tried to implement things like `untokenize`).


albertjan:
Thanks @timmartin 🎉

Issue #1079, opened 2020-04-22
tp$name on instances and klasses are equivalent...?


s-cork:
```
class A:
  pass

>>> A().foo
'A' object has no attribute 'foo'
>>> A.foo
'A' object has no attribute 'foo'
```

But `A` is not an instance of `'A' object` it is an instance of `type` and should read `type object 'A'`

I've been working on a fix for this so will do a pr later.

meredydd:
Yeah...line 485 of `type.js` looks Just Wrong to me.

```
Sk.builtin.type.makeIntoTypeObj = function (name, t) {
    Sk.asserts.assert(name !== undefined);
    Sk.asserts.assert(t !== undefined);
    t.ob$type = Sk.builtin.type;
    t.tp$name = name;
    (snip)
```
I think `t.tp$name` should be `"type"`.

s-cork:
I've done some looking around and here's what i've found

this is based on both the pypy implementation:
https://foss.heptapod.net/pypy/pypy/blob/branch/default/pypy/module/__builtin__/interp_classobj.py
`W_ClassObject`s get a `.name` attibute but `W_instanceObject`s only get `.w_class` attribute
#### example code
```python
        if exc:
            raise oefmt(space.w_AttributeError,
                        "%s instance has no attribute '%s'",
                        self.w_class.name, name)


```

and the cpython implementation:
https://docs.python.org/3.0/c-api/typeobj.html#tp_name

https://github.com/python/cpython/blob/master/Objects/typeobject.c
#### Example code
```
type -> tp$name
```

#### Another similar example:
```c
        PyErr_Format(PyExc_TypeError,
                     "first argument must be a string, not %.80s",
                     Py_TYPE(pystr)->tp_name);
```

#### and another...
```c
    if (!PyDict_Check(value)) {
        PyErr_Format(PyExc_TypeError,
                     "attrib must be dict, not %.200s",
                     value->ob_type->tp_name);
        return -1;
    }
```

---

#### My question is then - what is the paradigm?

Should it be that in Cpython when we do 
```c
PyType(value)->tp_name
/* and similarly */
value->ob_type->tp_name
```
This maps to Skulpt as
```
value.tp$name
/* and similarly */
value.ob$type.prototype.tp$name
// or equivalently
```

I see that this is true of functions... So it would be good to be consistent here.

@meredydd - Am I on the right track here. 
[...]

PR #1080, opened 2020-04-23
setting attributes of builtin types should fail

### LINKED ISSUE #1078: can set attributes of builtins in python 3
the algorithm to set an attribute of a builtin is broken in python 3. 

it checks the repr for the word `type`

but `class` is in the repr for python 3 which means I can do

```python
setattr(int, 'foo', 'bar')
```
without issue... 

Also `genericSetAttr` doesn't check for a builtin so it's fine to just do 
```python
int.foo = 'bar'
```

s-cork:
This pr aims to fix #1078 

The approach I took was to add the appropriate `tp$setattr` function to builtin type and type objects 

This negates the need to check whether an object is a builtin in the `builtin.setattr` function

### DIFF
--- a/src/import.js
+++ b/src/import.js
@@ -89,6 +89,9 @@ Sk.doOneTimeInitialization = function (canSuspend) {
         child["$d"].mp$ass_subscript(Sk.builtin.type.basesStr_, new Sk.builtin.tuple(bases));
         child["$d"].mp$ass_subscript(Sk.builtin.type.mroStr_, new Sk.builtin.tuple([child].concat(bases)));
         child["$d"].mp$ass_subscript(new Sk.builtin.str("__name__"), new Sk.builtin.str(child.prototype.tp$name));
+        child.tp$setattr = function(pyName, value, canSuspend) {
+            throw new Sk.builtin.TypeError("can't set attributes of built-in/extension type '" + this.tp$name + "'");
+        };
     };
 
     for (x in Sk.builtin) {
--- a/src/type.js
+++ b/src/type.js
@@ -517,6 +517,9 @@ Sk.builtin.type["$r"] = function () {
         return new Sk.builtin.str("<type 'type'>");
     }
 };
+Sk.builtin.type.tp$setattr = function(pyName, value, canSuspend) {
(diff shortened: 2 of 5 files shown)


src/import.js
```diff
@@ -96,6 +96,9 @@ Sk.doOneTimeInitialization = function (canSuspend) {
         if ((func.prototype instanceof Sk.builtin.object ||
              func === Sk.builtin.object) && !func.sk$abstract) {
             setUpClass(func);
+            func.tp$setattr = function(obj, pyName, value) {
```
meredydd, line 99: Move this to `setUpClass()`?

meredydd: Also, that's the wrong call signature for `tp$setattr`!
```
tp$setattr = function(pyName, data, canSuspend)
```

src/type.js
```diff
@@ -514,6 +514,9 @@ Sk.builtin.type["$r"] = function () {
         return new Sk.builtin.str("<type 'type'>");
     }
 };
+Sk.builtin.type.tp$setattr = function(obj, pyName, value) {
```
meredydd, line 517: Fix arg names/sig

src/type.js
```diff
@@ -514,6 +514,9 @@ Sk.builtin.type["$r"] = function () {
...
+    throw new Sk.builtin.TypeError("can't set attributes of built-in/extension type '" + obj.tp$name + "'");
```
meredydd, line 518: We're pretty confident we know what `obj.tp$name` is here, right? `obj` will be `Sk.builtin.type`!

However, this three-line function is useful in several places, hence the copy-and-paste. Perhaps it should be abstracted out? (Minor thing)

src/import.js
s-cork: moved and done

src/type.js
s-cork: done

src/type.js
s-cork: I've changed this to `this.tp$name`
This is currently correct but would need to adjust depending on how #1079  is resolved...

s-cork: I could change this to `this.prototype.tp$name`

s-cork:
@meredydd changes made - i wasn't sure what to do about the error message - for now i've left it as `this.tp$name`.

PR #1083, opened 2020-04-27
support dynamically created magic methods

s-cork:
Skulpt currently does not fully support dynamically created magic methods

```python
class A:
  pass
ax = A()
ay = A()
A.__add__ = lambda x, y: 5
print(ax+ay)
#TypeError: unsupported operand type(s) for Add: 'A' and 'A' 
```

During `klass` creation:
- attributes are added to the `klass` and `klass.prototype`
- dunder methods are added to slots where appropriate

however in `tp$setattr`:
- methods are ONLY added to the `klass`

this pr abstracts away the function for assigning dunder methods to skulpt slots and reuses the logic in `tp$setattr` as well as adding the attribute to both the `klass` and `klass.prototype`

this way instances have access to the dynamically created magic methods at the class level. 

_NB: it is consistent with python docs. A dynamically created magic methods on an instance has the expected effect. (i.e. `ax.__add__ = lambda x,y: 7` does not change at the class level). But a dynamically created method on the class should work._

---

No tests are broken. 

New test file added - test_magicmethods.py.


s-cork:
Also worth noting that the diff is really just an addition of the three lines in `tp$setattr`

```
    this.prototype[jsName] = value;
    if (jsName in Sk.dunderToSkulpt) {
        Sk.builtin.type.$assignDunder(this, jsName);
    }
```

where Sk.builtin.type.$assignDunder is the same code from `Sk.builtin.type` abstracted into its own function

src/type.js
```diff
@@ -736,3 +701,49 @@ Sk.builtin.type.prototype["__format__"] = function(self, format_spec) {
...
+                args[j] = arguments[i];
+                j += 1;
+            }
+        }
+
+        if (canSuspend) {
+            return Sk.misceval.callsimOrSuspendArray(magic_func, args);
+        } else {
+            return Sk.misceval.callsimArray(magic_func, args);
+        }
```
meredydd, line 731: This is a lot of code on a hot path (it runs every time the slot function is called). Most (in fact, nearly all) of these decisions can be hoisted up in to `$shortcutDunder`, so the branches only need to be taken once.

src/type.js
```diff
@@ -736,3 +701,49 @@ Sk.builtin.type.prototype["__format__"] = function(self, format_spec) {
...
+    };
+};
+
+
+Sk.builtin.type.$assignDunder = function(klass, dunder) {
```
meredydd, line 736: Extremely minor style nit: Is there a reason `$shortcutDunder` is a separate function from `$assignDunder` here?

s-cork: Only because shortcutDunder was a separate function originally. 
I can move into $assignDunder

src/type.js
s-cork: I'm not particularly confident on suspensions yet. But I separated them into two functions. 
one for magic methods that can suspend and one for magic methods that can't...

src/type.js
s-cork: Done and changed the name to $allocateSlot

s-cork:
@meredydd  changes made - there might be some more optimisation possible on the code you commented on. I separated it into two functions - 1 for suspensions and 1 for not. 

I also moved the logic into a single function `$allocateSlot`

src/type.js
```diff
@@ -564,6 +527,10 @@ Sk.builtin.type.prototype.tp$setattr = function (pyName, value) {
     // class attributes are direct properties of the object
     var jsName = pyName.$jsstr();
     this[jsName] = value;
+    this.prototype[jsName] = value;
```
meredydd, line 530: Should these get the reserved-word rewrites?

s-cork: Resolved in commit 272954faea8d465080a3c2fe069671d9617f49d4

PR #1088, opened 2020-04-29
move tp$name onto the prototype of the klass

### LINKED ISSUE #1079: tp$name on instances and klasses are equivalent...?

s-cork:
this pr is for the correctness and aims to fix #1079

It moves in the direction of `tp$_____` existing on the prototype of all type objects and not mixing and matching.

---

While working on this pr there were errors in the messages for methods because the logic was getting the `typeName` of the `im_class` which is now `type` 

This pr aims to improves the logic in `Sk.builtin.method`. It sets the `tp$name` of a method instance to the `im_class` it is bound to.

There is still work to do with methods since unbound methods are depreciated in python 3. 

But this pr only aims to fix using the correct type names.

---

Suggestions and suggested tests?

example of when the current set up fails, is when you try to do something with a klass that you shouldn't be able to... e.g. 
```python
class A: pass

A+A()
#TypeError: unsupported operand type(s) for Add: 'A' and 'A'
```
but it should be
```python
TypeError: unsupported operand type(s) for Add: 'type' and 'A'
```

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -1442,7 +1442,8 @@ Sk.builtin.delattr = function delattr (obj, attr) {
         }
         throw new Sk.builtin.AttributeError(Sk.abstr.typeName(obj) + " instance has no attribute '"+ attr.v+ "'");
     }
-    throw new Sk.builtin.TypeError("can't set attributes of built-in/extension type '" + obj.tp$name + "'");
+    // if we're here then we're a builtin type
+    throw new Sk.builtin.TypeError("can't set attributes of built-in/extension type '" + obj.prototype.tp$name + "'");
 };
 
 Sk.builtin.execfile = function execfile () {
--- a/src/type.js
+++ b/src/type.js
@@ -434,8 +434,6 @@ Sk.builtin.type.makeIntoTypeObj = function (name, t) {
     Sk.asserts.assert(name !== undefined);
     Sk.asserts.assert(t !== undefined);
     Object.setPrototypeOf(t, Sk.builtin.type.prototype);
-    t.tp$name = name;
-
     return t;
 };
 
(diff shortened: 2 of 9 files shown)


s-cork:
my bad - I misread cpython docs - dynamically created klasses don't have a `.` only statically created classes, which are not builtins.

e.g. `collections.deque` or `itertools.count`

will amend accordingly.

test/unit/test_descr_skulpt.py
```diff
@@ -19,7 +19,7 @@ def test(self):
         self.assertTrue(str(test).startswith('<function test'))
 
         bound_no_type = test.__get__(4)
-        self.assertEqual(str(bound_no_type), '<bound method ?.test of 4>')
```
albertjan, line 22: these are the python2 tests we should probably make these changes behind a flag

s-cork: Good point!

i've reverted these checks so that bound methods no type are `'?'` as before. (This will be the case in either `py3` or `py2`)

Which I figured is preferable to a `future` flag at the moment. Mostly because skulpt implementation of `methods` is entirely based on `py2` and to move to py3 will require more than a plaster over `$r` which is what I was initially trying to do with this.

s-cork: resolved

src/builtin.js
```diff
@@ -1440,7 +1440,7 @@ Sk.builtin.delattr = function delattr (obj, attr) {
         }
         throw new Sk.builtin.AttributeError(Sk.abstr.typeName(obj) + " instance has no attribute '"+ attr.v+ "'");
     }
-    throw new Sk.builtin.TypeError("can't set attributes of built-in/extension type '" + obj.tp$name + "'");
+    throw new Sk.builtin.TypeError("can't set attributes of built-in/extension type '" + obj.prototype.tp$name + "'");
```
meredydd, line 1443: Add a comment here explaining that if we get to this point, `obj` is a type object

src/type.js
```diff
@@ -484,7 +484,7 @@ Sk.builtin.type.makeIntoTypeObj = function (name, t) {
     Sk.asserts.assert(name !== undefined);
     Sk.asserts.assert(t !== undefined);
     t.ob$type = Sk.builtin.type;
-    t.tp$name = name;
+    t.tp$name = "type";
```
meredydd, line 487: This is going to conflict on rebase with #1131, but that's going to be so much nicer than this :)

src/builtin.js
s-cork: done 👍

PR #1091, opened 2020-05-03
change dict hash key func for performance and maintain insertion order

s-cork:
this pr aims to improve dictionary lookups. 

It changes the hash function to do the following:
```javascript
function kf(key) {
    let key_hash = key.$savedKeyHash_;
    if (key_hash !== undefined) {
        return key_hash;
    } else if (key.ob$type === Sk.builtin.str) {
        key_hash = key.$jsstr().replace(/^[0-9!#_]/, "!$&"); // avoid numbers and clashes 
        key.$savedKeyHash_ = key_hash;
        return key_hash;
    } else if (typeof key === "string") {
        // temporary while sysModules allows javascript strings as keys to python dicts
        return key.replace(/^[0-9!#_]/, "!$&");
    }
    key_hash = Sk.builtin.hash(key).v; // builtin.hash returns an int
    key.$savedKeyHash_ = key_hash;
    return key_hash;
};
```

---

# How it works
```python
d = {1:0, 1.1: 1, 1.2: 2, 'a' : 3, 'b': 4, 2: 5}
# the hash function in skulpt will see collisions for the first three keys
```

Is mapped to... 
```javascript
this: {//dict
  buckets:  // keep track of collisions 
    1: [{lhs: int_(1), rhs: int_(0)}, {lhs: float_(1.1), rhs: int_(1)}, {lhs: float_(1.2), rhs: int_(2)}] //keep track of collisions
    2: [{lhs: int_(2), rhs: int_(5)}] //keep track of collision
  entries:  // preserves insertion order and provide a fast path for str entries...
    #1_0: {lhs: int_(1),     rhs: int_(0)} // non string (int_(1)
    #1_1: {lhs: float_(1.1), rhs: int_(1)} // non string
    #1_2: {lhs: float_(1.2), rhs: int_(2)} // non string
     a:   {lhs: str('a'),    rhs: int_(3)} //string 
     b:   {lhs: str('b'),    rhs: int_(4)} // string 
    #2_0: {lhs: int_(2),     rhs: int_(5)} 
  size: 6
}
```

---

the rationale is two-fold:
- insertion order
  - python 3.6 preserves insertion order
  - javascript Objects preserve insertion order for strings
  - so using strings as keys in will also preserve insertion order
- iteration
  - over a dictionary is now as simple as

```javascript
for (let key_hash in this.entries) {
  item = this.entries[key_hash];
  key = item.lhs;
  val = item.rhs;
}
```

---
### tests

I edited print tests where order changed as the result of the new algorithm

---

### A note on sets
sets will now likely preserve insertion order because they use dict.js under the hood, 
I don't see this as a problem since the user should not expect sets to be ordered... 

```python
set([5,4,3,2,1,6,7])
# before {1, 2, 3, 4, 5, 6, 7}
# now {5, 4, 3, 2, 1, 6, 7}

```

---

### dealing with deleted items

replace the `item` in `bucket` with `undefined`

e.g. 
```python
d = {1:0, 1.1: 1, 1.2: 2}
del d[1]
```
the dict object will look like

```javascript
// deleted the zeroth item with hash value = 1
this: {//dict
  buckets:  // undefined signals a free slot
    1: [**undefined**, {lhs: float_(1.1), rhs: int_(1)}, {lhs: float_(1.2), rhs: int_(2)}] //keep track of collisions
  entries: 
    #1_1: {lhs: float_(1.1), rhs: int_(1)} // non string
    #1_2: {lhs: float_(1.2), rhs: int_(2)} // non string
  size: 2
```
The `undefined` value signifies there is a `free slot` and this will get filled upon the next collision with the hash = `1`.

e.g.
```python
d[1.3] = 3
# skulpt hash function maps 1.3 to the hash 1 i.e a collision
```

```javascript
this: {//dict
  buckets:  // inserted into free slot
    1: [{lhs: float_(1.3), rhs: int_(3)}, {lhs: float_(1.1), rhs: int_(1)}, {lhs: float_(1.2), rhs: int_(2)}] 
  entries: // preserves insertion order 
    #1_1: {lhs: float_(1.1), rhs: int_(1)} // non string
    #1_2: {lhs: float_(1.2), rhs: int_(2)} // non string
    #1_0: {lhs: float_(1.3), rhs: int_(3)} 
  size: 3
```

---
### does it improve speed?

Yes. 

```
fields.py 100 Ticks 10 loops in npm repl
this branch avg: 2.722899961471557
Master avg: 3.401200008392334
20% improvement
```

update (latest version of master approx 6-10%)


src/dict.js
```diff
@@ -158,20 +184,30 @@ Sk.builtin.dict.prototype.sq$contains = function (ob) {
...
+    }
+    this.$all_strings = false;
+
+    var bucket = this.buckets[k];
     var item;
 
     if (bucket === undefined) {
         // New bucket
-        bucket = {$hash: k, items: [
+        bucket = {$hash: new Sk.builtin.int_(parseInt(unhash_key(k))), items: [
```
meredydd, line 202: What's this change about?

rixner:
This is a reasonably invasive change for no performance improvement.  So, it appears that it comes down to whether we want to maintain insertion order?  Is this the "best" way to do that?

albertjan:
I ran this through field.py, and I saw a 3% perf improvement see numbers below: 

master:
```
external.        - field.py
Run time: 1.235s - time elapsed: 0.9880001544952393
Run time: 1.223s - time elapsed: 0.9809999465942383
Run time: 1.226s - time elapsed: 0.9739999771118164
Run time: 1.241s - time elapsed: 0.9850001335144043
Run time: 1.257s - time elapsed: 1.000999927520752
Run time: 1.249s - time elapsed: 1.000999927520752
Run time: 1.202s - time elapsed: 0.9590001106262207
Run time: 1.212s - time elapsed: 0.9590001106262207
Run time: 1.208s - time elapsed: 0.9620001316070557
Run time: 1.206s - time elapsed: 0.9560000896453857
avg.    : 1.2259 - avg         : 0.9766000509 
```

this branch
```
external.        - field.py
Run time: 1.18s  - time elapsed: 0.9430000782012939
Run time: 1.252s - time elapsed: 1.013999938964844
Run time: 1.186s - time elapsed: 0.9469997882843018
Run time: 1.18s  - time elapsed: 0.937000036239624
Run time: 1.181s - time elapsed: 0.9409999847412109
Run time: 1.193s - time elapsed: 0.9340000152587891
Run time: 1.182s - time elapsed: 0.9389998912811279
Run time: 1.192s - time elapsed: 0.9500000476837158
Run time: 1.175s - time elapsed: 0.937000036239624
Run time: 1.179s - time elapsed: 0.940000057220459
avg.    : 1.19   - avg         : 0.9481999874
```

s-cork:
@rixner  worth noting that my `$all_keys` function is really just the dict version of `sk$asarray` (#1054) which I extracted from the logic in creating the `dict_key_iterator`.  

---
side note and possible to add to this pr or to a separate one:

I think that it is worth considering preventing JS objects from being keys/values of a Py dict. 
which is that weird hack I have for the `hashkey function` :
```javascript
// because we sometimes let js strings be keys
else if (typeof key === "string") {
        return "_" + key;
    }
```

We allow this in `Sk.sysmodules = new Sk.builtin.dict;` where we pass in JS strings instead of Py strings. 

Which then causes issues since the dict module uses `Sk.misceval.richCompareBool` to compare keys to the bucket LHS items...  and the documentation for `Sk.misceval.richCompareBool` states that it should only be used with python objects. 😞 

---

is 3% a good (enough) speed improvement?

---

is it worth having a set of generic speed test programs that can be run? So that we can easily compare speeds using the same programs and the differences would be branches and systems.

albertjan:
> We allow this in Sk.sysmodules = new Sk.builtin.dict; where we pass in JS strings instead of Py strings.

that's weird and I agree we shouldn't allow js objects to live in our dict.

> is 3% a good (enough) speed improvement?

I think that's pretty good @meredydd will have a better view on this, but I think the really big savings have all been done so what remains are these kinds of improvements.

> is it worth having a set of generic speed test programs that can be run? So that we can easily compare speeds using the same programs and the differences would be branches and systems.

Yeah it would be good to formalise these, we've all been running @rixner's field.py script and I've also been running a slightly modified version of pystone.py. We used to have something like that, it would run a script 10 times and collect run time data, @mariechatfield made that. But that disappeared when we moved away from `skulpt.py` to `npm run *`.

[pystone-field.zip](https://github.com/skulpt/skulpt/files/4680807/pystone-field.zip)

s-cork:
i've added the `Sk.sysModules` change here and as it's own separate pr if this doesn't get pulled. #1098

s-cork:
Update: I came up with a significantly better hash function than the initial proposal in this pr. 

I've amended accordingly and see `20%` speed improvements on `fields.py` (figures in the pr description)

This new proposal also simplifies dictionary iteration significantly from a JS perspective.

albertjan:
This is dependent on the the sys modules PR isn't it? Did you push this new hash function? 😄

s-cork:
This PR includes the `Sys Modules` change - so yes totally dependent on the approval of that pr. 

re hash function - the `Sk.builtin.hash` remains the same. it's the `dict.js` `hash_function (var kf =)` which is different.  It's on line 75 in `dict.js`

s-cork:
I've reverted the `sysModules` change in this pr so that it is no longer a breaking change.

s-cork:
I've also added `sk$asarray` to `Sk.builtin.dict.prototype`

src/dict.js
```diff
@@ -69,68 +74,64 @@ Sk.builtin.dict.tp$call = function(args, kw) {
...
+        return;
+    }
+    for (let i=0; i<bucket.length; i++) {
+        item = bucket[i];
+        if (item === undefined) {
+            continue;
+        }
+        stored_key = item.lhs;
+        if (stored_key === key || Sk.misceval.richCompareBool(key, stored_key, "Eq")) {
+            return this.entries["#" + i + base_hash.slice(1,)];
```
meredydd, line 116: This method of construction confused me a little - it wasn't obvious that `base_hash.slice(1,)` will *always* start with an underscore, so I assumed there might be ambiguity here.
 
I think this will be resolved by documenting more about how this works in a code comment.

src/dict.js
```diff
@@ -69,68 +74,64 @@ Sk.builtin.dict.tp$call = function(args, kw) {
...
+        key_hash = "_" + key.$jsstr();
+        key.$savedKeyHash_ = key_hash;
+        return key_hash;
+    } else if (typeof key === "string") {
+        return "_" + key;
     }
-
-    return null;
+    key_hash = "#_" + Sk.builtin.hash(key).v;
+    key.$savedKeyHash_ = key_hash; // this is a base key hash
```
meredydd, line 90: Is it OK to attach this property to _any_ possible key?

src/dict.js
```diff
@@ -69,68 +74,64 @@ Sk.builtin.dict.tp$call = function(args, kw) {
...
-        eq = Sk.misceval.richCompareBool(item.lhs, key, "Eq");
-        if (eq) {
-            return item;
-        }
+    if (Sk.builtin.checkString(key)) {
+        key_hash = "_" + key.$jsstr();
+        key.$savedKeyHash_ = key_hash;
+        return key_hash;
+    } else if (typeof key === "string") {
+        return "_" + key;
```
LeszekSwirski, line 87: It might be better to return an unmodified string here, to creating a new string and possibly take advantage of previous string canonicalisation for the hash lookup; you could return an object literal (smth like `{key:key_hash}`) for the non-string case.

src/dict.js
```diff
@@ -16,9 +16,14 @@ Sk.builtin.dict = function dict (L) {
     }
 
     this.size = 0;
+    // entries will preserve insertion order of the key value pairs
+    this.entries = {};
```
LeszekSwirski, line 20: I would recommend `this.entries = { __proto__: null }` to create an object with a null prototype -- this will avoid issues with accesses to `Object.prototype`, and in V8 it hints that this object wants to be in dictionary mode.

src/dict.js
```diff
@@ -16,9 +16,14 @@ Sk.builtin.dict = function dict (L) {
...
+    // buckets will keep track of non string hashes and collisions;
+    // each base_hash will be of the form #_123: [{lhs: key, rhs: value}, undefined, {lhs: key, rhs: value}];
+    // undefined will be the result of a deleted item
     this.buckets = {};
```
LeszekSwirski, line 24: Could do the same thing here too actually.

src/dict.js
```diff
@@ -69,68 +74,64 @@ Sk.builtin.dict.tp$call = function(args, kw) {
...
-    var item;
-    var eq;
-    var i;
+Sk.builtin.dict.prototype.sk$asarray = function () {
+    const all_keys = [];
+    const entries = this.entries;
+    for (let hash in entries) {
+        all_keys.push(entries[hash].lhs);
+    }
+    return all_keys;
```
LeszekSwirski, line 100: Depending on desired browser support, this can be simplified to `return Object.values(entries).map(x=>x.lhs);`

src/dict.js
s-cork: yes since we already know it's `hashable` since `Sk.builtin.hash` deals with throwing the `unhashable` error here... 

Similarly the `Sk.builtin.hash` algorithm does the same thing...
```javascript
            value.$savedHash_ = value.tp$hash();
            return value.$savedHash_;
```

src/dict.js
```diff
@@ -157,47 +158,88 @@ Sk.builtin.dict.prototype.sq$contains = function (ob) {
...
-            {lhs: key, rhs: w}
-        ]};
-        this.buckets[k.v] = bucket;
-        this.size += 1;
+    const hash = kf(key);
+    if (hash[0] === "_") {
+        // we have a string so pass it to the dictionary 
+        if (this.entries[hash] === undefined) {
+            this.size += 1;
+            this.entries[hash] = {lhs: key};
```
LeszekSwirski, line 166: you should pre-initialize `rhs` to be `w` here too, otherwise the rhs property will go out-of-object (on V8 at least).

src/dict.js
s-cork: realised I could change this specific line to `return item`
but generally i've added some additional comments for clarity

src/dict.js
s-cork: Yes this is probably a good idea. I think there are only 3 values that could clash
`_proto__, _defineGetter__, _defineSetter__` 
but stranger things can happen!!!
will use `Object.create(null)`

src/dict.js
s-cork: do you mind providing a little more of a code snippet.

I think I follow but just in case I don't...

worth noting that I used the leading underscore since javascript objects only preserver insertion order for strings so if the user had a `{'1': 'foo'}` then without the leading underscore the insertion order would be lost...

and also 86-87 are only used by `sysModules` if #1098 gets merged these two lines will go... 
technically i could remove them already but it makes `sysModules` a little faster and avoids sending `strings` to `Sk.misceval.richCompareBool` which is only designed for `skulpt objects`

src/dict.js
s-cork: Nice!
@albertjan do we need to worry here - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values

src/dict.js
LeszekSwirski: I was imagining something like:

```js
if (key.ob$type === Sk.builtin.str) {
    return key.$jsstr();
}
return { key: Sk.builtin.hash(key).v };
```

which is used as

```js
key = kf(obj);
if (typeof key == "string") {
    return entries[key];
}
return entries[key.key]
```

Dunno how well a non-string saved key would work with the rest of the system.

The numbers thing is a good point -- I would suggest maybe special-casing prefixes? You could e.g. do:

```js
function sanitize_key(key) {
  return key
}

if (key.ob$type === Sk.builtin.str) {
    return key.$jsstr().replace(/^[0-9!#_]/, "!$&");
}
return "#" + Sk.builtin.hash(key).v;
```

which would make sure that only non-string hashes start with a `#`, and no hashes start with a number or a `_` (for avoiding conflicts with setting `__proto__`).

s-cork: I've had another play around. I think the `"_"` proceeding keys is the easiest solution... you never know when the replacing might conflict with some random string...? and it fixes the problem with ints in javascript not preserving insertion order...

The  `"_" + key.$jsstr()` should only happen the first time a string is entered into a dictionary. The next time that key is looked up `$savedKeyHash_` will return the appropriate value.

for the non string keys, I now just return the integer `hash value`. Which seems cleaner.

thoughts?

LeszekSwirski: Your call -- my regex approach didn't replace anything, but rather reduced the number of cases where it prepended `!` to only those where the string starts with one of `0-9!#_` (i.e., the "danger cases"). Always prepending something also works, albeit less efficiently. At least don't use `_` though, or you'll accidentally set the prototype on the object if your string is `_proto__`.

s-cork:
@meredydd - I did another speed test with `fields.py` - I think the sensitivity of the `Sk` object pre #1143 impacted the figures in the description. (Which were valid at the point i made them)
I'm now getting around 10% `2.79 seconds (master) vs 2.49 seconds (this branch)` (200 TICKS)

src/dict.js
s-cork: Oh i see now! my regex needs some revision! will have a think about this one tomorrow thanks!

meredydd: I like @LeszekSwirski's idea, and I can see how this could be a performance win, but it looks like we've got the potential for collisions here:

> ```js
> key = kf(obj);
> if (typeof key == "string") {
>     return entries[key];
> }
> return entries[key.key]
> ```

In particular, if we store the string `foo` and then a Python object that _hashes_ to `foo`, won't they overwrite each other in `entries`? This was the whole purpose of the prefixing...

meredydd: Re the other point - if we _are_ prefixing, it does seem like `_` is a poor choice of character, because of the existence of `__proto__`! Perhaps we should use `!` or something else that is definitely not going to be in a JS identifier...

LeszekSwirski: er, sorry, I think I meant "buckets" on the second return. Either way, I think adding a prefix to the "complex" hash and making sure that the "simple" hash doesn't start with that prefix is the better solution than a wrapper object. 

Regarding `Object.create(null)` vs `{__proto__: null}` -- there's no difference in the objects they produce :) setting __proto__ on _any_ object will set its prototype.

src/dict.js
albertjan: Lets be good citizens and use https://github.com/zloirock/core-js#ecmascript-object and use `Object.values` 😄

src/dict.js
s-cork: > In particular, if we store the string "foo" and a Python object that hashes to "foo", won't they overwrite each other in entries?

This won't happen... hash values from the python hash algorithm are always integers. 

So with the current version the hash value from `kf` will return a js string for python str ie the fast path. Or a JavaScript number if it's not a string (the .v of the return value from `builtin.hash`). 

I'll update this algorithm - on reflection I'll go with @LeszekSwirski solution of prepending based on the regex idea. Like you said it's efficient, avoids the `str(1)` not preserving insertion order as well as conflicts with the complex keys in entries (like `#2_0` - resulting from inserting an item with hash value 2 into the zeroth index of a bucket)

(Nb skulpt doesn't currently allow for user defined `__hash__` functions and it it did the builtin hash function would have to check it was a suitable return value (size and type)). 

Re the `__proto__` @albertjan I'll need you check if this affects interned!!

s-cork: updated! more suggestions welcome... I think it's there now...

src/dict.js
```diff
@@ -15,10 +15,40 @@ Sk.builtin.dict = function dict (L) {
...
+    /**
+     * the internal representation is a combination of a javascript hash map (this.entries)
+     * and a series of buckets (this.buckets) to account for collisions of the skulpt/python hash function
+     * 
+     * this.entries is used as a fast path for items where the key is a python str
+     * As well as to preserve insertion order for dict.items as per python 3.6
+     */
     this.size = 0;
-    this.buckets = {};
+    this.entries = Object.create(null); // avoid clashes with Object.prototype
```
LeszekSwirski, line 26: doesn't matter _massively_, but `{__proto__:null}` saves you two lookups and a call.

src/dict.js
```diff
@@ -69,68 +99,63 @@ Sk.builtin.dict.tp$call = function(args, kw) {
...
-        }
-        return null;
+function kf(key) {
+    // str => jsstr().replace(/^[0-9!#_]/, "!$&") avoids conflicts
+    // other => hash.v value from builtin.hash (javascript number)
+    let key_hash = key.$savedKeyHash_;
+    if (key_hash !== undefined) {
+        return key_hash;
+    } else if (key.ob$type === Sk.builtin.str) {
+        key_hash = key.$jsstr().replace(/^[0-9!#_]/, "!$&");
```
LeszekSwirski, line 109: You may also want to experiment will hoisting this regex out and pre-compiling it; might not matter, but I remember it making a performance difference in the parser.

src/dict.js
```diff
@@ -157,47 +182,94 @@ Sk.builtin.dict.prototype.sq$contains = function (ob) {
...
-        // New bucket
-        bucket = {$hash: k, items: [
-            {lhs: key, rhs: w}
-        ]};
-        this.buckets[k.v] = bucket;
-        this.size += 1;
-        return;
+        bucket = this.buckets[hash_value] = [];
+        key_hash = "#" + hash_value + "_" + 0; // this is the zeroth entry
+        bucket.push(item);
```
LeszekSwirski, line 215: may as well just have `bucket = this.buckets[hash_value] = [item]`?

src/dict.js
```diff
@@ -157,47 +182,94 @@ Sk.builtin.dict.prototype.sq$contains = function (ob) {
...
+        }
+        bucket_key = item.lhs;
+        if (bucket_key === key || Sk.misceval.richCompareBool(key, bucket_key, "Eq")) {
+            const key_hash = "#" + hash_value + "_" + i;
+            delete this.entries[key_hash];
+            bucket[i] = undefined; // undefined signals this slot is free for use in the event of reinsertion/collision
+            return item;
+        }
+    }
+    return;
```
LeszekSwirski, line 254: consider cleaning up `this.buckets[hash_value]` if `bucket.length == 0`

src/dict.js
s-cork: i got nervous about that because https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto

src/dict.js
s-cork: @LeszekSwirski  i'm not sure i follow...

src/dict.js
LeszekSwirski: That's reasonable, it's a legacy feature (albeit one that is unlikely to ever get removed) and `Object.create(null)` is close enough to just as good.

src/dict.js
LeszekSwirski: If `bucket.length == 0`, then afaict you can `delete this.buckets[hash_value]`. If I'm mistaken then please ignore me :)

s-cork: do you mean something like? 

```javascript
    const bucket = this.buckets[hash_value];
    let bucket_key, item;
    if (bucket === undefined || bucket.length === 0) {
        return;
    }
```
I still think I'm missing what you mean... sorry

s-cork: Oh I think i see now - are you saying if this was the only `item` in the `bucket` then we can `delete` the `bucket` it... rather than setting it to `undefined` (which we'd have to do if there were multiple items)??

so something like

```javascript
if (bucket.length === 1) {
    delete this.buckets[hash_value];
} else {
    bucket[i] = undefined; // undefined signals this slot is free for use in the event of reinsertion/collision
}
return item

```

LeszekSwirski: What I mean is, that if all the entries of bucket are undefined (I guess that's more complex than just bucket.length being set to zero, you may have to also do `delete bucket[i]` rather than setting undefined), then you can delete the entire bucket from the set of buckets, i.e. do the inverse of line 213. Basically just a memory cleanup for empty buckets. It's not a big deal probably :)

s-cork: ok I think i've got it... maybe

src/dict.js
```diff
@@ -157,47 +185,95 @@ Sk.builtin.dict.prototype.sq$contains = function (ob) {
...
-        return;
+        bucket = this.buckets[hash_value] = [item];
+        key_hash = "#" + hash_value + "_" + 0; // this is the zeroth entry
+    } else {
+        // we might have a freeslot from deleting an item
+        // so either insert into freeslot or push
+        const free_slot_idx = bucket.indexOf(undefined);
+        if (free_slot_idx !== -1) {
+            key_hash = "#" + hash_value + "_" + free_slot_idx;
+            bucket[free_slot_idx] = item;
```
s-cork, line 224: having deleted a slot rather than set it to `undefined` this check now becomes useless (I think).
```javascript
new Array(4).indexOf(undefined) // -1
```
cleaner then just to disregard freeslots here? - I can't think of a nice way to find an empty slot. 
So either 
A - skip this logic, 
B - revert deleting the element and set it to undefined, 
C - is there a nice way to find an empty slot.

src/dict.js
```diff
@@ -157,47 +185,95 @@ Sk.builtin.dict.prototype.sq$contains = function (ob) {
...
+            continue;
+        }
+        bucket_key = item.lhs;
+        if (bucket_key === key || Sk.misceval.richCompareBool(key, bucket_key, "Eq")) {
+            const key_hash = "#" + hash_value + "_" + i;
+            delete this.entries[key_hash];
+            delete bucket[i]; // create empty slot for use in the event of reinsertion/collision
+            if (bucket.every((item) => item === undefined)) {
+                delete this.buckets[hash_value];
+            }
```
s-cork, line 255: here's the approach I took to check if a bucket was an `Array` of `empty` slots...
better approach?
`delete this.buckets[hash_value]` or just `bucket.length = 0`

src/dict.js
```diff
@@ -69,68 +99,66 @@ Sk.builtin.dict.tp$call = function(args, kw) {
...
+    const bucket = this.buckets[hash_value];
+    let bucket_key, item;
+    if (bucket === undefined) {
+        return;
+    }
+    for (let i = 0; i < bucket.length; i++) {
+        item = bucket[i];
+        if (item === undefined) {
+            // free slot from having deleted an item.
+            continue;
```
s-cork, line 139: given the empty slots now...
If I instead do `for (let item of bucket)` it will just skip these slots. better?

src/dict.js
LeszekSwirski: You're right, I think B reverting deleting the element is your best option, it was a bad suggestion of mine.

src/dict.js
```diff
@@ -69,68 +99,66 @@ Sk.builtin.dict.tp$call = function(args, kw) {
...
+    return key_hash;
 };
 
-Sk.builtin.dict.prototype.key$pop = function (bucket, key) {
-    var item;
-    var eq;
-    var i;
+Sk.builtin.dict.prototype.sk$asarray = function () {
+    return Object.values(this.entries).map((x) => x.lhs);
+};
```
rixner, line 126: Am I understanding correctly that this will not necessarily return all of the keys?  Just the ones that are stored in entries, and not in buckets?  If true, doesn't that break the ```sk$asarray``` "contract"?  (Resulting in incorrect behavior when other functions use ```sk$asarray``` which assume that it will correctly convert a dictionary to an array of all keys.)  Or am I missing something?

s-cork: `this.entries` has all the keys. So it abides by the contract. (I think if it didn't some tests should have broken [i hope])

from the example in the description
```python
d = {1:0, 1.1: 1, 1.2: 2, 'a' : 3, 'b': 4, 2: 5}
```
```javascript
  this.entries: 
    #1_0: {lhs: int_(1),     rhs: int_(0)} // non string (int_(1)
    #1_1: {lhs: float_(1.1), rhs: int_(1)} // non string
    #1_2: {lhs: float_(1.2), rhs: int_(2)} // non string
     a:   {lhs: str('a'),    rhs: int_(3)} //string 
     b:   {lhs: str('b'),    rhs: int_(4)} // string 
    #2_0: {lhs: int_(2),     rhs: int_(5)} 
```

`entries` contains all the items
`buckets` deals with clashes - in other words - everything except strings

rixner: Rereading the comments, they do imply this, but the descriptions are a little unclear to me.  I feel like it would be helpful to be more explicit that all dictionary key/value pairs are in ```this.entries``` and they are **duplicated** in ```this.buckets``` if they are not string keys.  It would also be helpful, in my opinion, if the example dictionaries in the comments were the same dictionary in both cases.  In the ```entries``` description, you show the Python dictionary and the corresponding Skulpt representation.  In the ```buckets``` description, you use a different Python dictionary, don't show the Python dictionary, and then show the Skulpt representation of a single bucket without showing what was in entries.  If it were clearly the same dictionary, it would have been more obvious what was in entries, as that was already in the other comment.

Rereading it all, like I said, it does make sense.  I just think it could be much clearer.  Of course, I also could have just read more of the code, but I feel like the comments should speak for themselves....

meredydd (approved):
Green light from me - sounds like all issues have been resolved. Well done - this is a big and useful change.

PR #1098, opened 2020-05-26
put py strings in the sysmodules dict

s-cork:
`Sk.sysmodules` is a python dict in skulpt
but
`Sk.sysmodules` has js strings as keys rather than python strings. 

this seems odd to me. 

An example to highlight the oddness...

When doing print statements:  with every `print` statement we import `"sys"` and get the `"sys"` module from `Sk.sysmodules`. 

since `sys` is a jsstirng in a python dict there is no fast path. (the fast lookup is only for py strings)

the look up for `sys` takes the slow path and the bucket checking algorithm, getting passed onto `Sk.misceval.richCompareBool` which, according to documentation, is only supposed to be used with py objects. 😞  

since `Sk.misceval.richCompare` assumes py objects it then checks for all the possible richcompare shortcuts and reverse operations before finally giving up and checking `'sys'==='sys'`

(this happens with each print statements in a program)

related to #1091 

---

this pr changes all the `Sk.sysmodule` key assignments and lookups to be pyStrings. 
(NB: the values are already py modules)

### DIFF
--- a/src/import.js
+++ b/src/import.js
@@ -209,7 +209,7 @@ Sk.importModuleInternal_ = function (name, dumpJS, modname, suppliedPyBody, rela
 
         // if leaf is already in sys.modules, early out
         try {
-            prev = Sk.sysmodules.mp$subscript(modname);
+            prev = Sk.sysmodules.mp$subscript(new Sk.builtin.str(modname));
             // if we're a dotted module, return the top level, otherwise ourselves
             return topLevelModuleToReturn || prev;
         } catch (x) {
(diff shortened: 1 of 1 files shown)


albertjan:
@meredydd @rixner  I guess we're breaking the public api here. Are we ok with that?

acbart:
This would be very nice. My evolution of #789 that I've been using the past year was more difficult than I thought partially because of the weirdness of `sysmodules`.

s-cork:
closed via #1092

PR #1099, opened 2020-05-28
Pr/formatting

s-cork:
part discussion part pull request

I ran most of the core js files through prettier and es lint. 

prettier is very opinionated hence the huge number of changes. 
No code has been changed - only formatting.

```
I used a line length of 150
Object literals remove strings
Double quotes replace single quotes
Trailing commas are added at the end of arrays and literals
adds semi colons
tab space of 4
```
(settings can be adjust to preferences of others)

Having a fully formatted code base might make new pr diffs easier to read... 

This pr could be merged/adapted. A good time to merge might be when there are few pull requests in the pipe line. 

It can also easily revoked changed. I have no strong opinions.


albertjan:
I love it! Your name will be all over the codebase 😄 we'll all be absolved of the bugs. 

On a serious note I still love it we went through several stages of this in the past all very manual 😄 prettier is so nice. 

I assume you'll add the prettier config? 

I would like a pre-commit hook https://prettier.io/docs/en/precommit.html#option-1-lint-stagedhttpsgithubcomokonetlint-staged if we're going down this route and also like to run it on the ci too. Maybe we can use a GitHub action for that?

But before you go and do all this work I would like some input from @skulpt/committers

s-cork:
> I love it! Your name will be all over the codebase 😄 

motivation 1 😉 

> we'll all be absolved of the bugs.

haha I'm happy to take full responsibility for any bugs... 🕷️ 

---

> I assume you'll add the prettier config?

Yes for sure - will wait until some more wade in with opinions.

bnmnetp:
I'm also a fan of prettier for js and black for python.  But, I think we need to work on clearing the backlog of PR's in the queue or this will cause massive amounts of needless conflicts.

albertjan:
Yeah I agree we’ve made some good progress already but we should clear a few more esp. the bigger ones. 

@s-cork can you maybe have a look at the bytes class pr from @rixner and the str/bytes one from @meredydd and see if they are compatible and how you think we should get them in. 

@bnmnetp please add @s-cork as a contributor and to the commiters team. 😄

bnmnetp:
Also, https://github.com/psf/black has some notes on how we can configure blame to ignore certain revs, which is another reason why we would want to do the prettier mod in one commit.

bnmnetp:
@albertjan will do, re adding @s-cork to committers I had posted a question to the committers group asking about that very thing a few days ago, but got no response.  Maybe you didn't get notified.

albertjan:
@bnmnetp haha I didn't even know that was possible.

s-cork:
i've updated this pr so that it has a `pre-commit` hook with a `preiterrc.json` file. 

I've looked at the blame and there's also an open issue here https://gitlab.com/gitlab-org/gitlab/-/issues/31423 for ignoring revision on github

albertjan:
here's the ticket on the github community: https://github.community/t/support-ignore-revs-file-in-githubs-blame-view/3256

PR #1106, opened 2020-06-15
Upgrade this.py to Python 3

torenord:
Fixes `SyntaxError: bad input on line 28` when running with `__future__: Sk.python3` and still works for Python 2.


bnmnetp:
Thanks,  its always good to clean up the corners no matter how obscure.  😄

PR #1107, opened 2020-07-02
Another Bytes Implementation

### LINKED ISSUE #1032: str functions startswith() and endswith() take more than one arguments

s-cork:
Inevitably the work done in #989 and #983 would need adjusting if #1092 is approved. 
So here is the work that follows #1092 in order to implement `bytes`

Whilst looking carefully at both the above PRs I used bits of each. Neither was really compatible with the other without changing the approach of one them. 
#989 uses an `Array` type representation of a `bytes` object where #983 uses a `string` representation of a `bytes` object.

This PR uses the general approach of #983 whilst implementing missing features and incorporating all tests from #989 plus a few extras from CPython.

---

#### Overview of missing features

**missing from #989** 
- compilation of `bytes` like `b'\xf0\x9f\x8d\x95'`
- `nb$remainder` for `bytes`
- `$r` when `quote` character contained within the `bytes` object
- `is` for `bytes`

**missing from #983** 
- `1 in bytes([1, 2, 3])` and other variants
- `$r` emoji type characters representation for `bytes` e.g. 
```python
>>> bytes(b'\xf0\x9f\x8d\x95')
b'🍕'  # should be b'\xf0\x9f\x8d\x95
```
- `replace`, `ignore` as `errors` arguments for `encode`/`decode`
- `bytes` `nb$remainder` accounting for `%b`
- `fromhex`  edge cases
- `is` for `bytes`

---

while i'm here I also implement #1034 - close #1032

---

For the diff see https://github.com/s-cork/skulpt/pull/2


rixner:
@s-cork I appreciate all of the many contributions you are making to Skulpt, as I'm sure do many others.  However, I am a bit concerned about where you seem to be heading, so I feel compelled to speak out.  I mean this constructively, so I hope it comes across so, and that you take it as such.

In general, the Skulpt project is pretty friendly and inclusive. The contributors do not always agree, but we have generally been able to work things out so that the project happily moves forward.  Part of that inclusiveness is to provide constructive feedback on pull requests to allow the author to fix/update things.  In this case there are already **two** pull requests for bytes that have been around for a long time, both of which work, albeit with a few holes, as you have pointed out. It seems like the friendly thing to do here is to help the project pick which one makes more sense, merge it, perhaps allow the "losing" person to merge in the parts that were in their pull request but not the accepted one, and work from there.  Not to reimplement it and "steal" code from others.  Those pull requests have languished for a long time, but I am certain that @meredydd and I are fully capable of updating them or improving them as per the constructive comments of others.  This is not a case where either of us has withdrawn the pull request or said that we don't have time to update them based on others comments.  The issue is largely that I think @meredydd and I needed others to weigh in on which approach they wanted to take for the project.  Perhaps @meredydd sees the situation differently, though, and I don't mean to put words in his mouth.

I have seen projects where some give the impression that they know better than the other contributors, so instead of allowing them to improve/fix their pull requests, they just reimplement them.  I, for one, don't want to see Skulpt become that sort of project.

To me, the more friendly ordering would have been to accept #983 or #989 (with or without improvements) and then you could modify things in #1092 as needed.  Especially since #1092 may or may not get accepted (or may not get accepted any time soon).

Again, this is meant to be friendly and constructive and I hope you take it as such.  And again, thank you for your many contributions!

s-cork:
noted. I'm still learning. I did my best to attribute other's code correctly, you can see - for example - the work in `ast` on bytes will all be attributed to @meredydd via merge commit 
https://github.com/skulpt/skulpt/pull/1107/commits/079c72e7d68d9118cc8fa2160da7618576977c12

PR #1108, opened 2020-07-02
adjust checkInt function to fix #1082

### LINKED ISSUE #1082: random.randint raises ValueError when arguments are larger than 0x7ffffffe
Code:
```
import random
i = 0x7fffffff
print(type(i))
print(random.randint(0, i))
```

Result (from console):

```
<type 'int'>
ValueError: non-integer stop for randrange() on line 1
1>: print(random.randint(0, i))
```

Same code works in brython (for arbitrarily large numbers, as expected):

https://brython.info/tests/editor.html?lang=en&code=import%20random%0Ai%20%3D%200x7fffffff%0Aprint%28type%28i%29%29%0Aprint%28random.randint%280%2C%20i%29%29

s-cork:
A fix to close #1082 

the function `checkInt` checks if the argument is either `Sk.builtin.int_`, `Sk.builtin.float_` or a `JS number` which is an `integer`. The JS part is the bit that was broken for large JS integers. 

This PR replaces that part of test with `Number.isInteger` which we have used in some other places in the codebase. 

_(Worth noting that `Number.isInteger` is not supported in IE - do we need a `polyfill`? - I figured since it's already in use that's already been taken care of somewhere...)_

I added a test in `t441.py` to confirm that this fixes the issue.


albertjan:
Merging this as it's a small change with a test

Issue #47, opened 2013-01-27
Suggest a better default Sk.read() default implementation


csev:
It would seem to me that we should provide a decent SK.read() for the in-browser case.   It seems less than ideal to expect everyone who wants to use this in the browser to find and paste the code into their HTML pages:

```
function builtinRead(x)
{
    if (Sk.builtinFiles === undefined || Sk.builtinFiles["files"][x] === undefined)
        throw "File not found: '" + x + "'";
    return Sk.builtinFiles["files"][x];
}
```

And

```
Sk.configure({output:outf, read: builtinRead});
```

Just to get import statements working.

My recommendation is to put this code in the Sk.read() implementation in env.js - but only active is Sk.inBrowser is true.

The alternative (less attractive) is to do a bit more logic around the try/catch block in the Sk.importSearchPathForName implementation in import.js.

My preference is in env.js.  If the user wants a more sophisticated Sk.read() implementation, they can just override the default in the Sk.configure() call.

albertjan:
I think this is a good idea, I'm working on a fix to `import` that'll make `from bla import something` work for python packages.

bnmnetp:
@albertjan, I pushed a branch called cherry-picks that has such a fix in it.  It was mixed in with some other stuff for @eah13, so I did not make a PR, but if @eah13 has not incorporated that change into another PR you are welcome to look at it and use it.

@csv, I have long been an advocate of separating the builtinRead required for import from a more general read for file I/O simulation.   We certainly don't need to wait, but this is another reason behind my proposal this morning to look at a new more user friendly API for skulpt. See #446   Would be interested in your thoughts.

PR #1110, opened 2020-07-07
ipython example

s-cork:
I wanted to create a new example - potentially could be added to the website instead - but I thought i'd showcase it as a PR. 

it uses `ace` code editor. I'm sure I could adapt it to `codemirror` if preferred. 

here's the gif:

Bak9RsqQLN

___

The ipyhton repl adds keyboard bindings to scroll through previous entries
as well as a keyboard interrupt feature `ctrl + c`
`ctrl/cmd - enter` runs the command
`_`, `__`, `_1` work as per ipython
as do `In`, `Out`, `_i1`, `_ii` 

___

A couple of suggested changes with this pr - 
`Sk.retainGlobals` should match camelCase for `options["retainGlobals"]` (with backward compatibility)

Every example has the same `builtinRead` function should this become the default option?
Since it's the default option I remove this function from all examples. 

Both the above changes can be overridden - just ideas.

### DIFF
--- a/example/ipython.js
+++ b/example/ipython.js
@@ -95,7 +95,7 @@
         Sk.misceval.callsimArray(Sk.globals.In.append, [Sk.globals.In, codeAsPyStr]);
         this.inputs[this.inputs.length - 1] = code;
 
-        let compile_code = code.trimEnd();
+        let compile_code = code.trimEnd() || "None";
 
         const lines = compile_code.split("\n");
         let last_line = lines[lines.length - 1];
--- a/example/ipython.js
+++ b/example/ipython.js
@@ -1,0 +1,331 @@
+(function () {
+    var //finds import statements
+        importre = new RegExp("\\s*import"),
+        //finds defining statements
+        defre = new RegExp("def.*|class.*"),
+        //test for empty line.
+        comment = new RegExp("^#.*"),
+        //a regex to check if a line is an assignment
+        //this regex checks whether or not a line starts with
+        //an identifier followed with some whitspace and then an = and then some more white space.
+        //it also checks if the identifier is a tuple.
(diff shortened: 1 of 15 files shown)


src/env.js
```diff
@@ -261,7 +261,12 @@ Sk.output = function (x) {
  * todo; this should be an async api
  */
 Sk.read = function (x) {
-    throw "Sk.read has not been implemented";
+    if (Sk.builtinFiles === undefined) {
+        throw Sk.builtin.NotImplementedError("Sk.read has not been implemented");
```
albertjan, line 265: should we say that `-stdlib.js` wasn't loaded? instead of saying it's not implemented?

s-cork: Any preference on the error type?

albertjan: I don’t have a preference. 😁

albertjan:
awesome stuff

albertjan:
@skulpt/committers over to you.

alexwenbj:
Great!
But how to make code run?Enter, command + Enter, ctrl + Enter not work for my Chrome.

s-cork:
Which platform? I only tested on mac-os so far.
it should be 
```
                    win: "ctrl-enter",
                    mac: "command-enter",
```

alexwenbj:
@s-cork I ran it on Mac OS + Chrome. I clone all your ipython repo and now it can response for the command + enter,but it shows another errors as this image below:
image

alexwenbj:
@s-cork I ran it on Mac OS + Chrome. I clone all your ipython repo and now it can response for the command + enter,but it shows another errors as this image below:
image

src/env.js
```diff
@@ -257,11 +257,16 @@ Sk.output = function (x) {
...
  */
 Sk.read = function (x) {
-    throw "Sk.read has not been implemented";
+    if (Sk.builtinFiles === undefined) {
+        throw "skulpt-stdlib.js has not been loaded";
+    } else if (Sk.builtinFiles.files[x] === undefined) {
+        throw "File not found: '" + x + "'";
+    }
+    return Sk.builtinFiles.files[x];
 };
```
s-cork, line 270: @alexwenbj this change was added to this pr which might be the error you got from the sys module

alexwenbj: It works, thank you.

meredydd:
This is a thing of beauty. I want it on skulpt.org!

example/ipython.js
```diff
@@ -0,0 +1,333 @@
...
+        __name__: new Sk.builtin.str("__main__"),
+    };
+    Sk.globals.In = Sk.globals._ih;
+    Sk.globals.Out = Sk.globals._oh;
+
+    Sk.stopExecution = false;
+
+    const info = "# Python 3.7(ish) (Skulpt ipython, " + new Date() + ") \n" + "# [" + navigator.userAgent + "] on " + navigator.platform;
+
+    Sk.ipython = function (dom) {
```
meredydd, line 32: If this isn't an officially supported feature (just a cool example), it probably shouldn't be in the `Sk` namespace

meredydd (approved):
Lovely example! (And I agree with having a default `builtinRead` implementation rather than making everyone do their own, unless they need to)

It's an example, so I'm considering it a "small change" and will merge on my own as soon as GitHub lets me.

albertjan:
closes #47 

talk about long standing issues! 2013

albertjan:
@csev we do listen! 😄

PR #1111, opened 2020-07-08
improve __bulitins__.classmethod.classmethod

s-cork:
this pr aims to improve the hack with `classmethod` with an additional hack 🤷‍♂️ 

It essentially removes the `__module__` property from each of `property`, `classmethod` and `staticmethod`
It also removes these three objects from `Sk.globals` since they get added to `__builtins__` and shouldn't really be in `Sk.globals`

this way
```python
>>> classmethod
<class 'classmethod'>
# rather than <class '__builtins__.classmethod.classmethod'>
>>> globals()
{'__name__': '__main__', '__file__': 'repl.py', '__doc__': None, '__package__': None}
# rather than {..., 'classmethod': <class '__builtins__.classmethod.classmethod'>, ...}
```

a small change to `repl.js` which - since it retains globals carries over the `__name__` property `__builtins__.classmethod` 
By setting the `Sk.globals.__name__` property in advanced this remains the same after `classmethod` et al are injected

### DIFF
--- a/repl/repl.js
+++ b/repl/repl.js
@@ -41,6 +41,8 @@ Sk.configure({
     __future__: py3 ? Sk.python3 : Sk.python2,
 });
 
+Sk.globals = { __name__: new Sk.builtin.str("__main__")};
+
 var //finds lines starting with "print"
     re = new RegExp("\\s*print"),
     //finds import statements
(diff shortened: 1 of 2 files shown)


albertjan:
small change merging

PR #1112, opened 2020-07-08
iter and iternext

### LINKED ISSUE #1018: Sk.builtin.checkIterable
This test fails on the rare occasion that a class has an `__iter__` method with no `yield` and no `__next__` method

```python
class N:
    'Iterator missing __next__()'
    def __init__(self, seqn):
        self.seqn = seqn
        self.i = 0
    def __iter__(self):
        return self
```

```javascript
Sk.builtin.checkIterable(N(range(1000)) --> true
```
But I think this should be false since:

## python
```python
>>> iter(N(range(1000)) 
TypeError("iter() returned non-iterator of type 'N'")
```

## skulpt
```python
>>> iter(N(range(1000)) 
<__main__.N at 0x1114cd668>
```

When I call next I then get the TypeError but this is too late according to python
## skulpt
```python
>>> next(iter(N(range(1000)))
TypeError("iter() returned non-iterator of type 'N'")
```

s-cork:
this is a change that aims to fix #1018 
I found the bug when working on itertools #1020 and have uncommented the tests that were previously broken. 

prior to this pr `Sk.abstr.iter` was not throwing an error if `tp$iter/__iter__` existed but `tp$iternext/__next__` didn't exist... 

Additional to this pr I reduce the amount of options for `Sk.abstr.iter` which now only checks for `tp$iter` rather than both`tp$iter` and then `__iter__`
To support this change - `type.js` now only defines `tp$iter` and `tp$iternext` if the relevant dunders exist rather than defining them on all `klass` objects. 

Small hidden away bug fixes to some builtin iterators where `tp$iter` was not a function. 
(hidden since `Sk.abstr` previously defaulted to checking `__iter__` if `tp$iter` threw an error).

It should hopefully result in performance improvements for user defined iterators since we no longer do an attribute lookup of the `__next__` function with each and every iteration (as well as checking whether it's `next` or `__next__`). We do this once when the `klass` is created.


albertjan:
@skulpt/committers over to you.

rixner:
This looks good to me, but can we use this change as an excuse to get rid of ```Sk.__future__.dunder_next``` and just use ```Sk.__future__.python3```?

albertjan:
Yeah I'm ok with removing the dunder_next flag.

s-cork:
i've pushed a change that does the switch in `env.js` which is nicer. And means we only have one constant
`Sk.builtin.str.$next` which switches as necessary to `__next__` or `next` depending on `__future__`.

@albertjan / @rixner to remove the `dunder_next` flag from `__future__` should i delete lines `48` and `69` in `env.js` i.e. remove them from the `__future__` options entirely...

albertjan:
> to remove the dunder_next flag from __future__ should i also delete lines 48 and 69 in env.js i.e. remove them from the __future__ options entirely...

yep

PR #1114, opened 2020-07-09
compile pyBody if supplied string even if empty

s-cork:
A minor bug that I found when working on #1110 
If the supplied body is a string it should still be compiled - even if that string is empty. 
Otherwise the compile logic tries to find `.py` in `builtinFiles`

I didn't add tests but I did make the default code in `npm run brun` an empty string rather than a new line. 
A quick check of this branch vs master `npm run brun tok_test.py` will compile an empty line whereas before it would raise the error
```
ImportError: No module named <stdin>
```


albertjan:
small change - merged

PR #1116, opened 2020-07-09
improve performance of interned

s-cork:
This one improves performance for strings but only if doing a lot of str conversion like

```python
from time import time

for j in range(2):
    t = time()
    for i in range(3000000):
        str(i)
    print(time()-t)

# Master:     25.3270001411438,  12.41999983787537
# ThisBranch: 10.52700018882751, 10.72699999809265
```
the first loop sets strings into `interned` 
the second will get strings from `interned`

By using 
```javascript
Sk.builtin.interned = Object.create(null)
``` 
instead of 
```javascript
Sk.builtin.interned = {}
```
 there's no need for `"1"+ ret` since there won't be any name conflicts with `Object.prototype` since `Sk.builtin.interned` no longer inherits from `Object` and thus has no `__proto__` methods that would otherwise cause conflict.

### DIFF
--- a/src/str.js
+++ b/src/str.js
@@ -1,4 +1,12 @@
-Sk.builtin.interned = {};
+Sk.builtin.interned = Object.create(null);
+
+function getInterned (x) {
+    return Sk.builtin.interned[x];
+}
+
+function setInterned (x, pyStr) {
+    Sk.builtin.interned[x] = pyStr;
+}
 
 /**
  * @constructor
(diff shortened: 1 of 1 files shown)


s-cork:
When I was trying to work out inheritance I stumbled across it. Then, I guess a sudden wave of inspiration struck when I was playing with the constructors of type objects... 🧙‍♂️ I guess. 

I originally thought it would be useful in some other places when we had a use case of avoiding name mangling - but it doesn’t really help in most places. This though, seems a perfect use case. An object which has one job - to be looked up and doesn’t need any methods like toString or __defineGetter__ and actually those methods are detrimental for us here. 

It’s one of the small changes that I put on #1092 that doesn’t need to be there. This is part of that quest to try to separate out the small changes from that PR.

albertjan:
small change - merging

PR #1117, opened 2020-07-10
Pr/name mangling plus

### LINKED ISSUE #949: Using reserved names in a few places can cause unmangling issues
```python
def a_scope(length):
    def inner_scope():
        length+"anything"
    return inner_scope
a_scope("ada")()
```

Substitute `length` with any of the `reservedNames_` from `src/compile.js` (such as `name`, `toString`) and you can trigger the following error message in closures and certain attribute lookups.

```
NameError: Undefined variable in expression on line 3
```

The basic issue is that in some places Skulpt will mangle these variable names to avoid conflicting with the builtin JavaScript properties. However, there are circumstances where the names are not unmangled. Closures seem to be one such place: the parameter name is given via the `co_varnames` as unmangled string. I also got hit by this with manually setting attributes in a custom module (see #372).

I'm hoping I can play whack-a-mole with these unmanglings, but I wasn't clear if there's an underlying issue here.

s-cork:
this is building on top of #1070 and #1116 
close #949

I unify the handling of `reservedWords` and `reservedNames` so instead we have a single object and a single function `Sk.fixReserved`
- I remove passing of `_$rw$` to the client in `compile.js` - specifically for `functions` 
- I remove passing of `_$rw$` to the client in globals
- I implement @acbart fix outlined in issue #949 which is a related bug and seems sensible to include here

relevant tests added for each of the above that previously failed

---

The approach is to add `$mangled` onto each `str` instance. 
```javascript
Sk.builting.str = function str(x) {
...
    this.v = ret;
    this.$mangled = fixReserved(ret);
...
}
```

You'd think this might be slow but the same benchmark in #1116 shows identical speed
(below is a one time test)

```python
from time import time

for j in range(2):
    t = time()
    for i in range(3000000):
        str(i)
    print(time()-t)

# Master:      25.3270001411438,  12.41999983787537
# branch 1116: 10.52700018882751, 10.72699999809265
# this_branch: 10.25099992752075, 9.874000072479248
```

I think this is a strict improvement of the current approach. It simplifies the code in object lookups since we only deal with mangled names in one place - `typeLookup` (and the the odd internal cases when `$d` is an object)

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -44,7 +44,6 @@ Sk.builtin.object.prototype.GenericGetAttr = function (pyName, canSuspend) {
     var tp;
     var dict;
     var getf;
-    var jsName = pyName.$jsstr();
 
     tp = this.ob$type;
     Sk.asserts.assert(tp !== undefined, "object has no ob$type!");
@@ -59,11 +58,12 @@ Sk.builtin.object.prototype.GenericGetAttr = function (pyName, canSuspend) {
         } else if (dict.mp$subscript) {
             res = Sk.builtin._tryGetSubscript(dict, pyName);
         } else if (typeof dict === "object") {
-            res = dict[jsName];
+            const mangled = pyName.$mangled;
+            res = dict[mangled];
         }
         if (res !== undefined) {
             return res;
-        } else if (jsName == "__dict__" && dict instanceof Sk.builtin.dict) {
+        } else if (pyName.v == "__dict__" && dict instanceof Sk.builtin.dict) {
             return dict;
         }
     }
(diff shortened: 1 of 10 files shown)


src/symtable.js
```diff
@@ -422,8 +422,8 @@ SymbolTable.prototype.newTmpname = function (lineno) {
 SymbolTable.prototype.addDef = function (name, flag, lineno) {
     var fromGlobal;
     var val;
-    var mangled = Sk.mangleName(this.curClass, new Sk.builtin.str(name)).v;
-    mangled = Sk.fixReservedNames(mangled);
+    var mangled = Sk.mangleName(this.curClass, name).v;
```
s-cork, line 425: I’m fairly certain that the name object passed here is always a `Sk.builtin.str` object but someone who knows this part of the code base can conform -  @albertjan ? It seemed unnecessary then to str a str. I did put an assert statement when working on this pr and ran all the tests and there was no issue with this change...

albertjan: I mean the fact the tests pass mean that you're probably right. Would `mangleName` fail if it was called with a js string?

albertjan:
```
function mangleName (priv, ident) {
    var name = ident.v;
```
 
Yeah it would. So you're fine.

src/object.js
```diff
@@ -59,11 +58,12 @@ Sk.builtin.object.prototype.GenericGetAttr = function (pyName, canSuspend) {
...
             res = Sk.builtin._tryGetSubscript(dict, pyName);
         } else if (typeof dict === "object") {
-            res = dict[jsName];
+            const mangled = pyName.$mangled;
+            res = dict[mangled];
         }
         if (res !== undefined) {
             return res;
-        } else if (jsName == "__dict__" && dict instanceof Sk.builtin.dict) {
+        } else if (pyName.v == "__dict__" && dict instanceof Sk.builtin.dict) {
```
rixner, line 66: This should be ```pyName.$jsstr()``` for consistency.

s-cork: Good point. Will update next week.

meredydd:
This is a significant simplification, as well as strict correctness improvement, and I am happy to see that it doesn't cost in performance. Merging!

PR #1119, opened 2020-07-11
Min/max suspension aware

s-cork:
as found by @albertjan min/max isn't currently suspendible. 
I wrote code in #1065 that suggested suspension were allowed without realising. 

Here is a pr that makes these two functions suspension aware. 
I add `unit3/test_suspensions.py` for all future suspension tests and include tests that currently fail on master branch.


albertjan:
@meredydd about `unit3/test_suspensions.py` we can force skulpt to suspend every call right? would that have picked these up?

meredydd:
> @meredydd about `unit3/test_suspensions.py` we can force skulpt to suspend every call right? would that have picked these up?

Yes -- we should run the tests in debug mode, which suspends between every line of code, and that would catch it without you having to write explicit suspension tests. We used to have a mechanism for invoking the tests in debug mode, but I had a vague impression this had gone away with the (much improved!) new build system

PR #1120, opened 2020-07-12
use of keyword new

s-cork:
a selection of places where we should use `new` but don't
I also replaces `builtins` with `builtin` where appropriate


albertjan:
small change

PR #1121, opened 2020-07-12
Pr/math `new`, <native JS>, some formatting

s-cork:
4 types of changes to math.js
use of `new` keyword
make sure all js funcs have names to avoid the dreaded `<native JS>`
put it through prettier
a couple of var/cons/let adjustments

No change to logic in this pr. 

_edit: 
I've added a change to Babel minifier for stdlib to preserve func names so that things like `print(math.ceil)` work as expected for the user._

### DIFF
--- a/src/lib/math.js
+++ b/src/lib/math.js
@@ -239,7 +239,7 @@ var $builtinmodule = function (name) {
         }
     });
 
-    _isclose = function isclose(a, b, rel_tol, abs_tol) {
+    function isclose(a, b, rel_tol, abs_tol) {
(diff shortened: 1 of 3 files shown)


s-cork:
I forgot that babel minifier will just remove those function names 🤦 
I've added a commit to tell babel to preserve function names and some tests to test this...
I got the options from https://github.com/babel/minify/tree/master/packages/babel-preset-minify#options

the babel minified code goes from:
```
   l.isinf = new Sk.builtin.func(function (a) {
```
to...
```
    l.isinf = new Sk.builtin.func(function isinf(a) {
```

PR #1122, opened 2020-07-12
use of callsimArray

s-cork:
3 changes in places using `callsim(OrSuspend)Array` where this is our internally preferred callsignature
I also remove this in one place where it makes more sense to directly call the JS function.


albertjan:
small change

PR #1124, opened 2020-07-12
minor test adjustments

s-cork:
There are a few places I found where test are slow because skulpt can be slow

I change these tests with reduced iterations but maintain the integrity of the test.
I found these changes useful when developing and wanting to test all unit tests.  

---

It stops the lag in one or two places e.g. `test_math` which was previously 10+ seconds now runs quicker. 

There are 4 tests related to reducing iterations.

1. We run this test in two locations
```python
self.assertEqual(list(zip(SequenceWithoutALength(), range(2**23))), list(enumerate(range(5))))
# Cpython 0.01 sec | Skulpt 5 secs
# becomes
self.assertEqual(list(zip(SequenceWithoutALength(), range(2**10))), list(enumerate(range(5))))
# skulpt 0.01 secs
```
This test is making sure that zip only zips the shorter of the two iterators
Our range object is slow here because we create a list of `2**23` ints rather than generate each int on the fly
In Cpython `5` ints would be generated for this test, in Skulpt `2**23` ints are generated. 
I don't believe this change affects the integrity of this test. 

2. 
```python
self.assertEqual(str(factorial(8888)).count("8"), 2887)
# cpython 0.02 seconds | Skulpt 3.5 seconds
# becomes
self.assertEqual(str(factorial(888)).count("8"), 216)
```
Factorial hits big ints as soon as we go above `18!` 

3. 
```python
total = 1
for i in range(1, 1000): # becomes 100
    total *= i
    self.assertEqual(math.factorial(i), total)
    self.assertEqual(math.factorial(float(i)), total)

# Cpython 0.25 seconds | Skulpt 10 seconds
# again we hit bit ingeters as soon as we go above 18
```

4.

```python
from random import random, gauss, shuffle, seed
seed(0)
for j in range(200): # becomes 20
    vals = [7, 1e100, -7, -1e100, -9e-20, 8e-20] * 10
    s = 0
    for i in range(200):
        v = gauss(0, random()) ** 7 - s
        s += v
        vals.append(v)
    shuffle(vals)

    s = msum(vals)
    self.assertAlmostEqual(msum(vals), math.fsum(vals), 12)

# testing fsum matches a pure python implementation
# Cpython 0.2 | Skulpt 7.5
```


rixner:
I don't think the amount of time it takes to run the tests should be a factor. While these are similar, they are not the same tests.  And always/only writing tests that are "fast" might encourage us to leave larger cases untested.  So, I have mixed feelings about both this pull request and the general sentiment of using faster tests...

s-cork:
I agree it's not a good general sentiment. 

i've added comments about the 4 perspective tests that have reduced iterations in the description.

I simply found reducing these allowed for some quicker development. I found myself more likely to hit `npm run test` when I knew I wouldn't have to wait 100 seconds each time.

These 4 tests on my machine take around 35 seconds of that test time...

albertjan:
By reducing the amount of iterations of 4 tests we win 30% of the runtime of unit tests I like those numbers and I feel it makes sense. The test coverage doesn't significantly change, if at all.

test/unit/test_complex.py
```diff
@@ -114,7 +114,8 @@ def test_floordiv(self):
 
     def test_richcompare(self):
         self.assertIs(complex.__eq__(1+1j, 1 << 10000), False)
-        self.assertRaises(TypeError, complex.__lt__, 1+1j, None)
+        # in py3 this is returns NotImplemented so comment this out
```
meredydd, line 117: We're not in py3 in this file (although I imagine this change came across from somewhere where it's relevant!)

s-cork: yeah this was getting ahead of myself. I'll remove this test change until it makes sense to do so.

PR #1125, opened 2020-07-12
handling of bool in abstr.objectPositive/Negative

s-cork:
in `objectPositive/Negative` we handle `bools` as if they are a special case but since they are instances of `int` they are not special and so can be treated as per all other number types with an `nb$positive/negative` slot.

### DIFF
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -737,35 +737,18 @@ Sk.abstr.objectAdd = function (a, b) {
 
 // in Python 2.6, this behaviour seems to be defined for numbers and bools (converts bool to int)
 Sk.abstr.objectNegative = function (obj) {
-    var objtypename;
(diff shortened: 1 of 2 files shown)


src/abstract.js
```diff
@@ -737,35 +737,18 @@ Sk.abstr.objectAdd = function (a, b) {
...
-    var objtypename = Sk.abstr.typeName(obj);
-    var obj_asnum = Sk.builtin.asnum$(obj); // this will also convert bool type to int
-
-    if (obj instanceof Sk.builtin.bool) {
-        obj = new Sk.builtin.int_(obj_asnum);
-    }
-
-    if (obj.nb$negative) {
+    if (obj.nb$positive) {
         return obj.nb$positive();
```
s-cork, line 749: Note this is previously a bug I guess...

PR #1128, opened 2020-07-13
complex internal representation

s-cork:
Currently we use `Sk.builtin.float_` to represent complex number real and imag parts

this pr changes the internal representation to javascript numbers. 
It simplifies `complex.js` somewhat as we no longer do `this.real.v` but just `this.real`

### DIFF
--- a/src/complex.js
+++ b/src/complex.js
@@ -773,7 +773,7 @@ Sk.builtin.complex.prototype.__int__ = function (self) {
 };
 
 
-Sk.builtin.complex.prototype._internalGenericGetAttr = Sk.builtin.object.prototype.GenericGetAttr;
+Sk.builtin.complex.prototype.$internalGenericGetAttr = Sk.builtin.object.prototype.GenericGetAttr;
 
 /**
  * Custom getattr impl. to get the c.real and c.imag to work. Though we should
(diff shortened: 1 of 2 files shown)


src/complex.js
```diff
@@ -794,17 +787,8 @@ Sk.builtin.complex.prototype.tp$getattr = function (name) {
...
 Sk.builtin.complex.prototype.tp$setattr = function (name, value) {
-    if (name != null && (Sk.builtin.checkString(name) || typeof name === "string")) {
-        var _name = name;
-
-        // get javascript string
-        if (Sk.builtin.checkString(name)) {
-            _name = Sk.ffi.remapToJs(name);
-        }
-
-        if (_name === "real" || _name === "imag") {
```
albertjan, line 805: do we now support setting `real` or `imag`?

s-cork: only in the same we we did before e.g.
```
>>> complex(1,2).real
1
```
in #1092 we have full support for `real` and `imag` because they are just `getset_descriptors` and so no need to hack `tp$getattr` like we currently do in `complex.js`

in #1092 it's just
```
getsets = {
  real: {
    $get: function () { return new Sk.builtin.float_(this.real) }
  }
...
```
and similar `getsets_descriptors` for `int` and `float`

s-cork: Actually I've made a mistake on tp$setattr here. I'll fix and push.

s-cork: done with latest commit cae22b3

PR #1129, opened 2020-07-13
Pr/sum

s-cork:
This came up in @albertjan review of #1092 

It fixes a bug where `sum` doesn't work in certain cases like summing complex numbers. 
It also does not support suspensions for suspending iterators. 

I followed the general approach of CPython: 
https://github.com/python/cpython/blob/master/Python/bltinmodule.c#L2293

And use the `Sk.misceval.chain/iterFor` architecture

there are three cases: `fastSumInt`, `fastSumFloat`, `slowSum`
both `fastSumInt` and `fastSumFloat` call the slot `nb$add` directly whereas `slowSum` uses `Sk.abstr.numberBinOp`

I have yet to add suspension tests. I will add these to the same test file as in #1119 after that PR concludes. 

---

**suspensions**
```python
from time import sleep

def sleeping_gen():
    for i in range(5):
        sleep(.01)
        yield i

print(sum(sleeping_gen()))
print(sum(float(i) for i in sleeping_gen()))

```
tests akin to the above work on this branch and raise `SuspensionError`s on master


src/builtin.js
```diff
@@ -472,52 +472,76 @@ Sk.builtin.all = function all (iter) {
...
+    if (start === undefined || tot.constructor === Sk.builtin.int_) {
+        initValue = fastSumInt();
+    } else if (tot.constructor === Sk.builtin.float_) {
+        initValue = "float";
+    } else {
+        initValue = "slow";
     }
 
-    return tot;
+    return Sk.misceval.chain(
```
albertjan, line 528: could you elaborate on how this works with the `Break("slow")`

s-cork: `Sk.misceval.iterFor` will either complete and return undefined or return a break value. 
I needed an argument for the break value otherwise it would be equivalent to completing all the iterations. 
"slow" seemed like an appropriate value. It basically just needed to be any value that wasn't `undefined`

i.e. I need three states, `undefined` (loop completed)/ `"float"` use `fastSumFloat`/ not `"float"`/`undefined` use `slowSum`

I think I could write it without the use of "slow" and just enter the `slowSum` regardless... either the iterator has been consumed or it hasn't...

s-cork: here's the alt verstion with no `"slow"`

```javascript
    let initValue;
    if (start === undefined || tot.constructor === Sk.builtin.int_) {
        initValue = fastSumInt();
    } else if (tot.constructor === Sk.builtin.float_) {
        initValue = "float";
    } 

    return Sk.misceval.chain(
        initValue,
        (brValue) => {
            if (brValue === "float") {
                return fastSumFloat();;
            }
        },
        () => {
            return slowSum(); 
            // if the iterator has been consumed then this will enter slowSum and then exit
        },
        () => tot
    );
```
I'm not sure which I prefer...

s-cork:
variable name adjusted

PR #1131, opened 2020-07-14
Pr/js inheritance for py

s-cork:
This is the fundamental building block for #1092 presented here as the minimum required, non-breaking change

It leads to this prototypical inheritance chain for `type` and `object`
```javascript

type.__proto__             = type   (type instanceof type)      // isinstance(type, type)     => True
type.__proto__.__proto__   = object (type instanceof object)    // isinstance(type, object)   => True
type.prototype.__proto__   = object (type subclasssof object)   // issubclass(type, object)   => True
object.__proto__           = type   (object instanceof type)    // isinstance(object, type)   => True
object.__proto__.__proto__ = object (object instanceof object)  // isinstance(object, object) => True

```
Since function is no longer down the prototypical chain we add this back on to the `prototype` of `type` so that type objects remain callable... useful when they need to be constructors... 😉 
```
Sk.builtin.type.prototype.call = Function.prototype.call
Sk.builtin.type.prototype.apply = Function.prototype.apply
```

---

I would add the tests below:
```python
self.assertTrue(issubclass(type, object))
self.assertFalse(issubclass(object, type))
```
which are now fixed in this pr...
But `issubclass` returns a javascript `boolean` rather than a `Sk.builtin.bool`! 
That can be a separate pr... 

---
In the second commit... 

I remove `Sk.builtin.type.prototype.tp$richcompare` 
This is not a thing in python. 
What this function does is already taken care of in `misceval.richCompareBool`
To hack back in `type(int('1'*600)) == int` in py3 mode I make `Sk.builtin.lng.prototype.ob$type = Sk.builtin.int_`
in `env.js`

### DIFF
--- a/src/import.js
+++ b/src/import.js
@@ -60,8 +60,6 @@ Sk.importSearchPathForName = function (name, ext, searchPath) {
  * @return {undefined}
  */
 Sk.doOneTimeInitialization = function (canSuspend) {
-    var proto, name, i, x, type, typesWithFunctionsToWrap, builtin_type, j;
-
     // can't fill these out when making the type because tuple/dict aren't
     // defined yet.
     Sk.builtin.type.basesStr_ = new Sk.builtin.str("__bases__");
@@ -70,11 +68,10 @@ Sk.doOneTimeInitialization = function (canSuspend) {
     // Register a Python class with an internal dictionary, which allows it to
     // be subclassed
     var setUpClass = function (child) {
-        var parent = child.prototype.tp$base;
-        var bases = [];
-        var base;
+        const parent = child.prototype.tp$base;
+        const bases = [];
 
-        for (base = parent; base !== undefined; base = base.prototype.tp$base) {
+        for (let base = parent; base !== undefined; base = base.prototype.tp$base) {
             if (!base.sk$abstract && Sk.builtins[base.tp$name]) {
                 // check the base is not an abstract class and that it is in the builtins dict
                 bases.push(base);
@@ -82,7 +79,7 @@ Sk.doOneTimeInitialization = function (canSuspend) {
         }
 
         child.tp$mro = new Sk.builtin.tuple([child].concat(bases));
-        if (!child.tp$base){
+        if (!child.hasOwnProperty("tp$base")){
             child.tp$base = bases[0];
         }
         child["$d"] = new Sk.builtin.dict([]);
@@ -91,8 +88,8 @@ Sk.doOneTimeInitialization = function (canSuspend) {
         child["$d"].mp$ass_subscript(new Sk.builtin.str("__name__"), new Sk.builtin.str(child.prototype.tp$name));
     };
 
-    for (x in Sk.builtin) {
-        type = Sk.builtin[x];
+    for (let x in Sk.builtin) {
+        const type = Sk.builtin[x];
         if (type instanceof Sk.builtin.type && type.sk$abstract === undefined) {
             setUpClass(type);
         }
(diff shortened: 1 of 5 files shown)


src/import.js
```diff
@@ -89,16 +89,12 @@ Sk.doOneTimeInitialization = function (canSuspend) {
         child["$d"].mp$ass_subscript(Sk.builtin.type.basesStr_, new Sk.builtin.tuple(bases));
         child["$d"].mp$ass_subscript(Sk.builtin.type.mroStr_, new Sk.builtin.tuple([child].concat(bases)));
         child["$d"].mp$ass_subscript(new Sk.builtin.str("__name__"), new Sk.builtin.str(child.prototype.tp$name));
-        child.tp$setattr = function(pyName, value, canSuspend) {
-            throw new Sk.builtin.TypeError("can't set attributes of built-in/extension type '" + this.tp$name + "'");
-        };
     };
```
s-cork, line 92: this is now part of the `type.prototype.tp$setattr` implementation

src/import.js
```diff
@@ -89,16 +89,12 @@ Sk.doOneTimeInitialization = function (canSuspend) {
...
 
     for (x in Sk.builtin) {
-        func = Sk.builtin[x];
-        if ((func.prototype instanceof Sk.builtin.object ||
-             func === Sk.builtin.object) && !func.sk$abstract) {
-            setUpClass(func);
+        type = Sk.builtin[x];
+        if (type instanceof Sk.builtin.type && type.sk$abstract === undefined) {
+            setUpClass(type);
         }
```
s-cork, line 98: use instanceof because we can...

src/object.js
```diff
@@ -16,6 +16,50 @@ Sk.builtin.object = function () {
...
+ *
+ * @ignore
+ */
+(function setUpBaseInheritance () {
+    Object.setPrototypeOf(Sk.builtin.type.prototype, Sk.builtin.object.prototype);
+    Object.setPrototypeOf(Sk.builtin.type, Sk.builtin.type.prototype);
+    Object.setPrototypeOf(Sk.builtin.object, Sk.builtin.type.prototype);
+    Sk.builtin.type.prototype.tp$base = Sk.builtin.object;
+})();
+
```
s-cork, line 62: here's where the magic happens

albertjan:
@s-cork why did you close this? 😄

s-cork:
no reason apart from to give you some breathing room...!

test/unit3/test_inheritance.py
```diff
@@ -109,8 +109,8 @@ def test_skulpt_bugs(self):
         self.assertTrue(isinstance(type, type))
         self.assertTrue(isinstance(object, object))
         self.assertTrue(isinstance(object, type))
-        # self.assertTrue(issubclass(type, object)) # commenting out until issubclass returns a pyBool
-        # self.assertFalse(issubclass(object, type))
+        self.assertTrue(issubclass(type, object))
+        self.assertFalse(issubclass(object, type))
```
s-cork, line 114: is travis bugging out? this change works fine on my home pc and can't be the cause of all the run time errors that travis found... ?

s-cork: something to not compatible with a recent commit to master - will investigate and get back

src/import.js
```diff
@@ -70,47 +68,43 @@ Sk.doOneTimeInitialization = function (canSuspend) {
     // Register a Python class with an internal dictionary, which allows it to
     // be subclassed
     var setUpClass = function (child) {
-        var parent = child.prototype.tp$base;
-        var bases = [];
-        var base;
+        const parent = child.prototype.tp$base;
+        debugger;
```
albertjan, line 72:
```suggestion
```

😄

s-cork: 🙃

s-cork:
There are two things I'd like to add onto this pr. I'm not sure whether to add them here or as separate prs. 

I don't think that `tp$mro/tp$base` should be on both the `klass` and `klass.prototype` (similar to my pr with `tp$name` #1088) it should just be on the `prototype` imo. 

I also think that internally `tp$mro` should be an array and when the user asks for `__mro__` they get the tuple.

src/object.js
meredydd: No kidding. :exploding_head: 

Comment needs updating to (I think):
```
 * type.__proto__             = type.prototype   (type instanceof type)
 * type.__proto__.__proto__   = object.prototype (type instanceof object)
 * type.prototype.__proto__   = object.prototype (type subclassof object)
 * object.__proto__           = type.prototype   (object instanceof type)
 * object.__proto__.__proto__ = object.prototype (object instanceof object)
```

PR #1133, opened 2020-07-15
adjust richCompareBool

### LINKED ISSUE #530: Skulpt richcomparebool returns always True or False
The function returns always a single boolean value even if a class implements `__eq__` and wants to return a list of bools. This works in real Python but currently not in Skulpt as richcomparebool does call `Sk.misceval.true(x)` on every result from a dunder method.

s-cork:
this pr adjusts richCompareBool and fixes an issue where 
```python
>>> None < None
False # yes in py2 No in py3
``` 
in py3 mode this should be an Error. 

It simplifies some of the logic in the function. 
It is also fixes a couple of issues with number types and the `is` operator. 
e.g. 
```python
>>> True is 1 #should be false
True # in skulpt we treat is the same for instances of number but shouldn't
>>> class A(int): pass
>>> A(1) is 1
True # should be False - same issue with bool - A(1) is an instance of int...
```

Because the above was a hidden bug in skulpt it meant that we were not getting correct results from
```python
self.assertTrue(True & True) 
# in skulpt this is 1 but because the `is` operator treated True and 1 the same hence we didn't notice this was a bug
```

So I override `Bool` number slots [as per Cpython](https://github.com/python/cpython/blob/master/Objects/boolobject.c#L95). 

For this pr I also needed to add slots for `ob$eq` for `bool` because `bool` is an unusual builtin (in that it inherits from `int`) and because `richCompareBool` is still weird. 

`richCopareBool` is still weird because there are 3 ways we can do any comparison operation... the slot `ob$eq` the `tp$richcompare` or the dunder method (More fixes in a future pr see richcompare bool [here](https://github.com/s-cork/skulpt/blob/prototypical_getsets_mappingproxy/src/misceval.js#L394) in #1092 where there is 1 way it can be called)

### DIFF
--- a/src/misceval.js
+++ b/src/misceval.js
@@ -236,62 +236,42 @@ Sk.misceval.opSymbols = {
 Sk.misceval.richCompareBool = function (v, w, op, canSuspend) {
     // v and w must be Python objects. will return Javascript true or false for internal use only
     // if you want to return a value from richCompareBool to Python you must wrap as Sk.builtin.bool first
-    var wname,
-        vname,
-        ret,
-        swapped_method,
-        method,
+    var ret,
         swapped_shortcut,
-        shortcut,
-        v_has_shortcut,
-        w_has_shortcut,
-        op2shortcut,
-        vcmp,
-        wcmp,
-        w_seq_type,
-        w_num_type,
-        v_seq_type,
-        v_num_type,
-        sequence_types,
-        numeric_types,
-        w_type,
-        v_type;
+        shortcut;
 
     Sk.asserts.assert((v !== null) && (v !== undefined), "passed null or undefined parameter to Sk.misceval.richCompareBool");
     Sk.asserts.assert((w !== null) && (w !== undefined), "passed null or undefined parameter to Sk.misceval.richCompareBool");
 
-    v_type = v.ob$type;
-    w_type = w.ob$type;
+    const v_type = v.ob$type;
+    const w_type = w.ob$type;
 
     // Python 2 has specific rules when comparing two different builtin types
     // currently, this code will execute even if the objects are not builtin types
     // but will fall through and not return anything in this section
-    if (!Sk.__future__.python3 &&
-        (v_type !== w_type) &&
-        (op === "GtE" || op === "Gt" || op === "LtE" || op === "Lt")) {
+    if (!Sk.__future__.python3 && v_type !== w_type && (op === "GtE" || op === "Gt" || op === "LtE" || op === "Lt")) {
         // note: sets are omitted here because they can only be compared to other sets
-        numeric_types = [Sk.builtin.float_.prototype.ob$type,
-                         Sk.builtin.int_.prototype.ob$type,
-                         Sk.builtin.lng.prototype.ob$type,
-                         Sk.builtin.bool.prototype.ob$type];
-        sequence_types = [Sk.builtin.dict.prototype.ob$type,
-                          Sk.builtin.enumerate.prototype.ob$type,
-                          Sk.builtin.filter_.prototype.ob$type,
-                          Sk.builtin.list.prototype.ob$type,
-                          Sk.builtin.map_.prototype.ob$type,
-                          Sk.builtin.str.prototype.ob$type,
-                          Sk.builtin.tuple.prototype.ob$type,
-                          Sk.builtin.zip_.prototype.ob$type];
+        const numeric_types = [Sk.builtin.float_, Sk.builtin.int_, Sk.builtin.lng, Sk.builtin.bool];
+        const sequence_types = [
+            Sk.builtin.dict,
+            Sk.builtin.enumerate,
+            Sk.builtin.filter_,
+            Sk.builtin.list,
+            Sk.builtin.map_,
+            Sk.builtin.str,
+            Sk.builtin.tuple,
+            Sk.builtin.zip_,
+        ];
 
-        v_num_type = numeric_types.indexOf(v_type);
-        v_seq_type = sequence_types.indexOf(v_type);
-        w_num_type = numeric_types.indexOf(w_type);
-        w_seq_type = sequence_types.indexOf(w_type);
+        const v_num_type = numeric_types.indexOf(v_type);
+        const v_seq_type = sequence_types.indexOf(v_type);
+        const w_num_type = numeric_types.indexOf(w_type);
+        const w_seq_type = sequence_types.indexOf(w_type);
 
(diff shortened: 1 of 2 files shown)


src/misceval.js
```diff
@@ -360,63 +340,65 @@ Sk.misceval.richCompareBool = function (v, w, op, canSuspend) {
...
-            return v.numberCompare(w) === 0;
-        } else if (v instanceof Sk.builtin.float_ && w instanceof Sk.builtin.float_) {
-            return v.numberCompare(w) === 0;
-        } else if (v instanceof Sk.builtin.lng && w instanceof Sk.builtin.lng) {
-            return v.longCompare(w) === 0;
+        if (v_type === w_type) {
+            if (v === w) {
+                return true;
+            } else if (v_type === Sk.builtin.float_) {
+                return v.v - w.v === 0;
```
albertjan, line 349: this is weird 😁 why not `v.v === w.v`

s-cork: Oooh. Not sure. I think that's because v.numberCompare(w) did. Will change it

src/misceval.js
```diff
@@ -514,55 +495,43 @@ Sk.misceval.richCompareBool = function (v, w, op, canSuspend) {
...
+            }
         }
     }
 
-
     // handle equality comparisons for any remaining objects
     if (op === "Eq") {
-        if ((v instanceof Sk.builtin.str) && (w instanceof Sk.builtin.str)) {
-            return v.v === w.v;
-        }
```
rixner, line 554: Why is it the case that this is no longer necessary? I believe that it can be cut, but I don't know/remember why it was here in the first place. With string interning working correctly, I agree it should not be necessary, but I'm sure this was here for a reason at some point in time.

src/misceval.js
```diff
@@ -360,63 +339,65 @@ Sk.misceval.richCompareBool = function (v, w, op, canSuspend) {
...
         return Sk.misceval.chain(Sk.abstr.sequenceContains(w, v, canSuspend), Sk.misceval.isTrue);
     }
     if (op === "NotIn") {
-        return Sk.misceval.chain(
-            Sk.abstr.sequenceContains(w, v, canSuspend),
-            function(x) { return !Sk.misceval.isTrue(x); }
-        );
+        return Sk.misceval.chain(Sk.abstr.sequenceContains(w, v, canSuspend), function (x) {
+            return !Sk.misceval.isTrue(x);
+        });
```
rixner, line 377: I am not a fan of mixing gratuitous, unnecessary formatting changes with real content changes. This pull request does a lot of that.

src/misceval.js
```diff
@@ -360,63 +339,65 @@ Sk.misceval.richCompareBool = function (v, w, op, canSuspend) {
...
+        if (v_type === w_type) {
+            if (v === w) {
+                return true;
+            } else if (v_type === Sk.builtin.float_) {
+                return v.v === w.v;
+            } else if (v_type === Sk.builtin.int_) {
+                return v.v === w.v;
+            } else if (v_type === Sk.builtin.lng) {
+                return v.longCompare(w) === 0;
+            }
```
rixner, line 353: What is the argument for breaking the abstraction here and comparing the ".v" values instead of using numberCompare? (Same for "IsNot")

src/misceval.js
s-cork: NumberCompare checks if other instanceof int_ which is redundant since we know already. So I didn't see the abstraction as beneficial with float or int here.

src/misceval.js
s-cork: Would you prefer a separate pr for formatting?

src/misceval.js
s-cork: As you said. I figured it was before interned was implemented/working and I can't see this line as being necessary.

s-cork: NumberCompare checks if other instanceof int_ which is redundant since we know already. So I didn't see the abstraction as beneficial with float or int here.

src/misceval.js
s-cork: I've gone through the blame.
interned - [Jul 2010](https://github.com/skulpt/skulpt/commit/720410b5aecba0d54dc84a6883895b95e138895e)
this code [Aug 2010](https://github.com/skulpt/skulpt/commit/55639a1ffa9281bd60ed18f7e3b4e87f362c180b)

fixes to interned much later...

So I presume this code was written as a reaction to `interned` not working completely... since it was written 1 month after interned was implemented and way before interned accounted for keys on `Object`'s `__proto__`. I think therefore it's safe to remove it now...

rixner: Fair enough!

src/misceval.js
rixner: Makes sense.

src/misceval.js
rixner: The issue is that this makes it unnecessarily harder to use git blame when you need to figure out when/why something was done and it creates the potential for unnecessary conflicts when people have internal modifications (and even with other pending pull requests).  We have eslint set up for the project in an agreed upon way.  If eslint is okay with something, then I don't think it should be changed, especially when it's not relevant to the content of the PR.

If you don't like the agreed upon style, the community should have an independent conversation about that, and, in my opinion, if we change anything, the tools need to be set up to catch style violations.  That's what we did last time we cleaned up the code base.  Otherwise, new violations will be introduced and others will introduce "fixes" in subsequent pull requests.  It's a never ending rat hole for people to reformat code that is not to their liking.

s-cork: is this a deal breaker? 

I typically do a prettier format of the function I'm working on and so formatting changes here are just what prettier told me to do 😬 . 

But I get that it can take away from the content of the PR. And I try to make a judgement call about formatting going too far...

s-cork: how do we feel about const and let changes in this pr? is that the same or different to style changes?

s-cork: i've reverted some style changes and left the const changes.

rixner: The formatting changes in this particular PR are not deal breakers for me, I am just making a point that I find these things to muddy the waters unnecessarily. Given that the current style is not based on what prettier does (whether or not it should be), I'm not a fan of changing things just because it says to do so.  Again, though, if we all decide we want to go that way, that's fine, but we should stop, merge all the pending pull requests that we are going to merge, and just reformat all the code in one go.  But, there are a bunch of pending pull requests out there, and every such formatting change could cause trouble for them.

I agree the const/let changes are actual improvements to the code, not stylistic preferences.

PR #1135, opened 2020-07-15
remove StandardError from error.js

s-cork:
This pr takes the opinion that we are preferring `python 3` and making sure `python 2` doesn't break.

To keep StandardError in the game I make `StandardError` = `Exception` in the switch. 
I can't imagine this would break any code since using `StandardError` in code would be unusual, which I guess is why it's not in `python 3`...

But maybe there is a weird case where this will break some code somewhere that I haven't considered.


albertjan:
Would when in python2 mode this still work:
```
try:
     something()
except StandardError:
     yadda()
```
still work?

s-cork:
Yeah that would definitely work. 

What wouldn't work 

```python
class A(Exception): pass

try:
  raise random.choice([A(), StandardError()])
catch StandardError:
  print 'foo'
catch Exception:
  print 'bar'
```

But to me it seems very convoluted to come up with examples that break... and the advantage is that we don't get `StandardError` showing up in the MRO in py3 mode.

PR #1137, opened 2020-07-15
Exceptions repr and str methods as per python

s-cork:
our repr for exceptions is not correct

Skulpt
```python
>>> e = TypeError('foo')
>>> str(e)
TypeError: foo
>>> repr(e)
TypeError: foo
```
Python
```python
>>> e = TypeError('foo')
>>> str(e)
'foo'
>>> repr(e)
TypeError('foo')
```

I fix this. And keep our `toString` implementation the same so that raised error messages do not change.

We also have a strange thing going on with args in errors. Where we assume that if there are more than 3 args that the file and line number have been provided. But this is strange because technically a user can provide multiple arguments to an Exception and skulpt will just assume that these are file numbers and files
e.g.
```python
e = Exception('a', 'b', 'c', 'd') # skulpt assumes that 'b' is the file, and 'c' is the line number
```

So instead I check whether the first argument is javascript string before making this assumption...
I could be wrong here but internally the `filename` and `line number` are a javascript `string` and `number` respectively so I think the use of `v` was incorrect for `filename`...

As a result of these changes I change the print tests... and use the `repr(e)` to check the error printouts


test/run/t498.py.real.force
```diff
@@ -50,8 +50,8 @@ floating point and long integers
 0.0883883476483 <type 'float'>
 
 ERROR CHECKING:
-TypeError: unsupported operand type(s) for pow(): 'list' and 'str' on line 53
```
albertjan, line 53: do we still need this force file or do we now output the same as python?

albertjan:
We might want to keep the line number though. Because we don't have stack traces for educational purposes it's probably good to know where something went wrong. 😄

test/run/t902.py.real
```diff
@@ -1,7 +1,7 @@
...
-KeyError: 5 on line 20
+TypeError('first argument must be callable')
+TypeError('first argument must be callable')
+TypeError('object is not iterable')
+KeyError('5')
 None
-TypeError: __missing__() takes at most 1 arguments (2 given) on line 30
-KeyError: {1: 2} on line 35
+TypeError('__missing__() takes at most 1 arguments (2 given)')
+KeyError('{1: 2}')
```
s-cork, line 7: something like this still needs to be forced:
should be...
`__missing__() takes exactly one argument (2 given)`

test/run/t498.py.real.force
s-cork: how can I check? is it only travis that does this or if I delete them on my machine will I be able to tell? I think yes to a most of them...

s-cork:
You still get line numbers in `toString` so getting an error that you haven't caught will do the same thing as before. 
e.g. `t73.py` remains unchanged. 

or something like
```python
print(1/0) # ZeroDivisionError: integer division or modulo by zero on line 1
```
remains the same...

It's only when you catch an error and print it...

test/run/t902.py.real
s-cork: and this is a different error message to py3 🤔

test/run/t498.py.real.force
s-cork: i removed all the forced files associated with the tests I changed... travis likes it...

test/run/t902.py.real
albertjan: the only difference between py3 and py2 I see is:
py3: `TypeError('expected at most 1 arguments, got 2')`
py2: `TypeError('expected at most 1 arguments, got 2',)`

albertjan: we have this every so often we realise we should be running our tests agains python as well as against skulpt. 😄 we used to do that which is why we have these force files.

s-cork: > the only difference between py3 and py2 I see is:
> py3: TypeError('expected at most 1 arguments, got 2')
> py2: TypeError('expected at most 1 arguments, got 2',)

Ah yes - i meant to mention that. I chose the py3 version... and didn't see a reason to go with the `py2` version
So should I leave the force files in in that case?

PR #1138, opened 2020-07-15
only send ints to mp subscript not number

s-cork:
this pr is a suggestion - since we've made the conclusion that `tp$getattr` should only use `pyStrings` I think we should do something similar for `mp$subscript` which can currently accept `numbers`

this pr does not make that a requirement but simply picks up the places where internally we are using number as the index for `mp$subscript` instead of `Sk.builtin.int_`
(I will comment at each place) 

I put `random.js` through prettier - hence the semi colons. 
I also provided a fast case for `random.shuffle` which typically will receive a `list`

### DIFF
--- a/src/lib/random.js
+++ b/src/lib/random.js
@@ -84,7 +84,7 @@ var MersenneTwister = function (seed) {
     /* mti==N+1 means mt[N] is not initialized */
 
     this.init_genrand(seed);
-}
+};
 
 /* initializes mt[N] with a seed */
 MersenneTwister.prototype.init_genrand = function (s) {
@@ -100,7 +100,7 @@ MersenneTwister.prototype.init_genrand = function (s) {
         this.mt[this.mti] >>>= 0;
         /* for >32 bit machines */
     }
-}
+};
 
 /* initialize by an array with array-length */
 /* init_key is the array for initializing keys */
@@ -113,7 +113,7 @@ MersenneTwister.prototype.init_by_array = function (init_key, key_length) {
     j = 0;
     k = (this.N > key_length ? this.N : key_length);
     for (; k; k--) {
-        var s = this.mt[i - 1] ^ (this.mt[i - 1] >>> 30)
+        var s = this.mt[i - 1] ^ (this.mt[i - 1] >>> 30);
         this.mt[i] = (this.mt[i] ^ (((((s & 0xffff0000) >>> 16) * 1664525) << 16) + ((s & 0x0000ffff) * 1664525)))
             + init_key[j] + j;
         /* non linear */
@@ -145,7 +145,7 @@ MersenneTwister.prototype.init_by_array = function (init_key, key_length) {
 
     this.mt[0] = 0x80000000;
     /* MSB is 1; assuring non-zero initial array */
-}
+};
 
 /* generates a random number on [0,0xffffffff]-interval */
 MersenneTwister.prototype.genrand_int32 = function () {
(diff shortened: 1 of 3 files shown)


src/lib/random.js
```diff
@@ -439,7 +437,7 @@ var $builtinmodule = function (name) {
         Sk.builtin.pyCheckType("seq", "sequence", Sk.builtin.checkSequence(seq));
 
         if (seq.sq$length !== undefined) {
-            var r = toInt(myGenerator.genrand_res53() * seq.sq$length());
+            var r = new Sk.builtin.int_(toInt(myGenerator.genrand_res53() * seq.sq$length()));
             return seq.mp$subscript(r);
```
s-cork, line 441: place 1

src/lib/random.js
```diff
@@ -449,20 +447,27 @@ var $builtinmodule = function (name) {
...
+                L[r] = L[i];
+                L[i] = tmp;
+            }
+        } else if (x.sq$length !== undefined) {
             if (x.mp$ass_subscript !== undefined) {
                 for (var i = x.sq$length() - 1; i > 0; i -= 1) {
-                    var r = toInt(myGenerator.genrand_res53() * (i + 1));
+                    var r = new Sk.builtin.int_(toInt(myGenerator.genrand_res53() * (i + 1)));
+                    i = new Sk.builtin.int_(i);
                     var tmp = x.mp$subscript(r);
```
s-cork, line 464: place 2

src/range.js
```diff
@@ -36,7 +36,7 @@ Sk.builtin.range_.prototype.mp$subscript = function (index) {
     sub = this.v.mp$subscript(index);
     if (sub instanceof Sk.builtin.list) {
         if (Sk.builtin.checkNone(index.start)) {
-            start = this.v.mp$subscript(0).v;
+            start = this.v.mp$subscript(new Sk.builtin.int_(0)).v;
         } else {
```
s-cork, line 40: place 3

src/lib/string.js
```diff
@@ -66,10 +66,10 @@ var $builtinmodule = function (name) {
...
         }
 
-        var words = Sk.misceval.callsimArray(mod.split, [s, sep]);
+        var words = Sk.misceval.callsimArray(mod.split, [s, sep]).v;
         var capWords = [];
-        for (var i = 0; i < words.v.length; i++) {
-            var word = Sk.builtin.list.prototype['list_subscript_'].call(words, i);
+        for (var i = 0; i < words.length; i++) {
+            var word = words[i];
             var cap = Sk.misceval.callsimArray(mod.capitalize, [word]);
```
s-cork, line 73: place 4 - just access the js array directly here instead

src/lib/random.js
```diff
@@ -449,20 +447,27 @@ var $builtinmodule = function (name) {
...
                     x.mp$ass_subscript(r, x.mp$subscript(i));
                     x.mp$ass_subscript(i, tmp);
```
s-cork, line 466: Another place where we are assigning the index `i` to `tmp` but `i` here was previously a number...

albertjan:
would it make sense to log a warning in `.mp$subscript` is called with a number instead of an int?

s-cork:
maybe - but at present our 'builtin.checkInt`, `misceval.isIndex` and `misceval.asIndex` accept javascript numbers so I think that would warrant a separate pr as a breaking change...

albertjan:
small change

PR #1139, opened 2020-07-15
remove numOpAndPromote

s-cork:
This pr removes `numOpAndPromote` from `Sk.abstr.numberBinOp`. 
This function no longer seem to do anything of value and isn't part of the equivalent CPython algorithms.
Removing it improves the speed of `fields.py`

It did deal with converting `int/float` to `complex` which should have been handled by the `nb$reflected_*` slots that were not previously part of `complex.js` - I add these in.


meredydd:
Agreed - removing old, unhelpful code is a virtue!

PR #1140, opened 2020-07-15
change null to none

s-cork:
a niche bug

```python
def f():
   x = yield
   yield x

>>> gen = f()
>>> next(gen) is None
Error can't read ob$type of null
```

We send null rather than `Sk.builtin.none.none$`


albertjan:
small change

PR #1141, opened 2020-07-16
Pr/isSubType

s-cork:
this is a useful utility function and follows from #1131 

It lives down the prototypical chain of all type objects and can be used like
```javascript
if (obj.ob$type.$isSubType(other)) {
    // do a thing
}
```
or 
```javascript
if (Sk.builtin.checkClass(cls) && cls.$isSubType(other)) {
    // do a thing
}
```

These check are used throughout the code base in `descriptors`

The check typically replaces `issubclass` in the codebase which was returning a js boolean for convenience. 
I change this so that `issubclass` returns a `pyBool` for correctness and add a test that would have failed. 

collections types broke this change because they don't have an `mro` so i add these in.

I also change the `Sk.builtin.check*` functions because `arg !== null` is not sufficient if the `arg` is `undefined`
(I came upon this when I inadvertently sent an `undefined` value to `Sk.builtin.checkClass`)
I think `arg != null` is better because it checks both for `null` or `undefined`. If an `undefined` value snuck in the user would receive the error (`cannot read property x of undefined`) 
_(This can be reverted or made into it's own pr)_

### DIFF
--- a/src/lib/collections.js
+++ b/src/lib/collections.js
@@ -32,6 +32,7 @@ var $builtinmodule = function (name) {
         };
 
         Sk.abstr.setUpInheritance("defaultdict", mod.defaultdict, Sk.builtin.dict);
+        mod.defaultdict.tp$mro = new Sk.builtin.tuple([mod.defaultdict, Sk.builtin.dict, Sk.builtin.object]);
 
         mod.defaultdict.prototype['$r'] = function () {
             var def_str = Sk.misceval.objectRepr(this.default_factory).v;
@@ -113,6 +114,7 @@ var $builtinmodule = function (name) {
         };
 
         Sk.abstr.setUpInheritance("Counter", mod.Counter, Sk.builtin.dict);
+        mod.Counter.tp$mro = new Sk.builtin.tuple([mod.Counter, Sk.builtin.dict, Sk.builtin.object]);
 
         mod.Counter.prototype['$r'] = function () {
             var dict_str = this.size > 0 ? Sk.builtin.dict.prototype['$r'].call(this).v : '';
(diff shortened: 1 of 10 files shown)


src/lib/collections.js
```diff
@@ -276,6 +278,7 @@ var $builtinmodule = function (name) {
         }
 
         Sk.abstr.setUpInheritance("OrderedDict", mod.OrderedDict, Sk.builtin.dict);
+        mod.OrderedDict.tp$mro = new Sk.builtin.tuple([mod.OrderedDict, Sk.builtin.dict, Sk.builtin.object]);
```
meredydd, line 281: One question: Are `tp$mro`s now required for hand-crafted types like this? Because if so, this is probably going to be a breaking change for anyone who has taken these as examples for building their own classes.

(On the other hand, anyone who's copied this example is producing a purely prototypical class, so perhaps we're fine, and you're just setting these MROs for style points?)

s-cork: so I added these because in #1092's implementation of `$isSubType` , it assumed that every object had a `tp$mro`. In this branch when I used that approach it failed on something like:

```python
issubclass(OrderedDict, str) 
```

As a result I adjusted the `$isSubType` algorithm in this branch so that it wouldn't fail if `tp$mro` wasn't defined:
```javascript
    return this === other || 
           this.prototype instanceof other || 
          (this.hasOwnProperty("tp$mro") && this.tp$mro.v.includes(other));
```

So adding `tp$mro` to collections isn't necessary for this pr, but having noticed that the `tp$mro` didn't exist on `collection` classes I added them...

PR #1143, opened 2020-07-20
This shouldn't improve performance

s-cork:
I don't know why this seems to improve performance on my machine... 
It seems trivial and I was hoping someone could confirm/explain what's going on here!

it accounts for my claim that #1013 improved speed significantly. It's not that #1013 significantly improved speed but more that the #1013 branch was prior to the inclusion of `Sk.setupDictIterators` which seemed to slow down speed by approx 50% somehow...

It doesn't matter if I do `npm run build` or `npm run devbuild`
I was using `fields.py` `100 Ticks` `10 loops`
I got `3.4 seconds` vs `1.9 seconds` which seems ridiculous.

I hope there's a sensible explanation 🙏

### DIFF
--- a/src/dict.js
+++ b/src/dict.js
@@ -456,17 +456,7 @@ Sk.builtin.dict.prototype.py3$values = function (self) {
 
 Sk.builtin.dict.prototype["values"] = new Sk.builtin.func(Sk.builtin.dict.prototype.py2$values);
 
-Sk.setupDictIterators = function (python3) {
-    if (python3) {
-        Sk.builtin.dict.prototype["keys"] = new Sk.builtin.func(Sk.builtin.dict.prototype.py3$keys);
-        Sk.builtin.dict.prototype["values"] = new Sk.builtin.func(Sk.builtin.dict.prototype.py3$values);
-        Sk.builtin.dict.prototype["items"] = new Sk.builtin.func(Sk.builtin.dict.prototype.py3$items);
-    } else {
-        Sk.builtin.dict.prototype["keys"] = new Sk.builtin.func(Sk.builtin.dict.prototype.py2$keys);
-        Sk.builtin.dict.prototype["values"] = new Sk.builtin.func(Sk.builtin.dict.prototype.py2$values);
-        Sk.builtin.dict.prototype["items"] = new Sk.builtin.func(Sk.builtin.dict.prototype.py2$items);
-    }
-};
+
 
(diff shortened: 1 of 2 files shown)


albertjan:
@LeszekSwirski I rarely tag you in our PR's, but these moments I wonder what's going on and you might have an insight that could help. 😄 🙏

LeszekSwirski:
Yeah, that's pretty mysterious :/

My best guess from just looking at it is that the Sk object shape is being mutated after other functions have already accessed it, so in those functions it becomes a polymorphic lookup (of Sk-with-setupDictIterators, and Sk-without-setupDictIterators), plus any existing optimized code accessing Sk will deoptimize. But that's just a guess -- I can't find `fields.py` anywhere to profile it, do you have a link to a copy?

s-cork:
[field.py.zip](https://github.com/skulpt/skulpt/files/4965140/field.py.zip)

LeszekSwirski:
Well I'm definitely not seeing anything obvious, looking a profile with https://v8.github.io/tools/v8.4/profview/ everything just seems to take a little longer. Looking at ICs in https://v8.github.io/tools/v8.4/ic-explorer.html doesn't seem to show much difference between the runs.

Adding `Sk.setupDictIterators = 10` to `dict.js` regresses, so the root cause is almost certainly this `Sk` object shape change, but I can't quite pin down what that shape change is affecting.

LeszekSwirski:
Oh, I see what this is! Using https://v8.github.io/tools/v8.4/map-processor.html, I found that this one extra property on `Sk` makes it trigger a "TooManyFastProperties" on a transition later on, which transitions the `Sk` object to dictionary mode, making all subsequent `Sk` accesses be dictionary lookups rather than IC'ed direct field accesses.

The TooManyFastProperties limit is 128, and sure enough:

```sh
$ grep -r 'Sk\.[a-zA-Z_][a-zA-Z_0-9]\+' . --only-matching --no-filename | sort -u | wc -l
129
```

albertjan:
Wow @LeszekSwirski thank you for looking into this 😄

LeszekSwirski:
No problem, thanks for the nerd snipe :)

We've got an intern working on performance investigation tooling, I sent this to them as an example user journey so that hopefully in the future it's easier to notice issues like this.

meredydd:
I suggest merging this PR to solve the immediate problem, and then (re)organising the contents of the `Sk` object so we're not skating so close to this line!

PR #1149, opened 2020-08-03
decorators run in reverse order

### LINKED ISSUE #1147: Decorator wraps in wrong order
When chaining decorators into a function, they are executed in the opposite order.

Considering this example from stackoverflow:
[https://stackoverflow.com/questions/27342149/decorator-execution-order](https://stackoverflow.com/questions/27342149/decorator-execution-order)

The output of the code should be: "&lt;b&gt;&lt;i&gt;hello world&lt;/i&gt;&lt;/b&gt;"
But instead I get: "&lt;i&gt;&lt;b&gt;hello world&lt;/b&gt;&lt;/i&gt;"

s-cork:
fix #1147 

This seemed a sensible fix but perhaps there's a better way.


albertjan:
ugh we need to run our tests of python too! small change merging!

Issue #1148, opened 2020-08-01
re.findall() implementation not correct


alexwenbj:
For example:
```
import re
key = r"mat cat hat pat"
p1 = r"[^c|p]at"
pattern1 = re.compile(p1)
print(pattern1.findall(key))
```
In Python, it prints:
`['mat', 'hat']`
In Skulpt, it prints:
`[]`

PR #1151, opened 2020-08-04
:bug: fix #issue1148 for re.findall

Dustyposa:
fix `re.findall`  code 
```python
import re
key = r"mat cat hat pat"
p1 = r"[^c|p]at"
pattern1 = re.compile(p1)
print(pattern1.findall(key))
```
detail: #1148


src/lib/re.js
```diff
@@ -344,7 +344,7 @@ var $builtinmodule = function (name) {
             if (start == "^") {
                 start = str.indexOf("\n") + 1;
             }
-            if (end == Sk.builtin.none.none$) {
+            if (end === Sk.builtin.none.none$ || endpos === Sk.builtin.none.none$) {
                 end = str.length;
```
s-cork, line 348: Is it possible for `end` to be `Sk.builtin.none.none$` here? since it was remapped to Js...

Dustyposa: Oh, I cant's make sure it could be...
I do the least of change for the code..
if `end` not be `Sk.builtin.none.none$`, we can delete it...

s-cork: I think we can just delete `end === Sk.builtin.none.none$` since it was remapped to js a few lines earlier.

Dustyposa: Yes, directly return `null`, I changed the condition and committed it.

PR #1152, opened 2020-08-05
arrayFromIterable - (tuple suspension aware)

s-cork:
This Pr adds a helper function `Sk.misceval.arrayFromIterable`

it is suspension aware and uses the `sk$asarray` architecture from #1054  where possible.
The function will raise a `TypeError` if the object is not iterable.
As well as a suspension error if the iterator suspends but `canSuspend` is `undefined || false`

I adjust the constructors of `tuple` and `list` to use this function. 
This adjustment allows `tuple` to be suspension aware. (tests added in `test_suspensions.py`

I also identify a few places in `itertools` where this function is beneficial. 

`sk$asarray` is designed for builtins so I add the flag `hp$type` to all instance of `klass` objects and check this flag before using the `sk$asarray` architecture within `arrayFromIterable`.

### DIFF
--- a/src/lib/itertools.js
+++ b/src/lib/itertools.js
@@ -175,20 +175,15 @@ var $builtinmodule = function (name) {
 
     var _combinations = function (iterable, r) {
         Sk.builtin.pyCheckArgsLen("combinations", arguments.length, 2, 2);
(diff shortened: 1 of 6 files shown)


meredydd:
I'm deeming this small enough to merge myself; it's tidy and very helpful.

PR #1154, opened 2020-08-05
don't set __module__ unless globals.__name__ exists

s-cork:
this pr makes a minor adjustment to `Sk.misceval.buildClass`

there are some cases in `lib` `js` files when the global object passed to `Sk.misceval.buildClass` does not have a `__name__` property. 

This is true of the `re.js` module for example. 
This might become problematic because when `locals.__module__` is set to `undefined` it will later create a dictionary where the `__module__` value is also `undefined`.

This hasn't caused issues in the codebase but I think it would be good to eliminate setting a value of a dictionary to be undefined. 
In cpython doing the equivalent of `mp$ass_subscript(key, undefined)` would be the equivalent of deleting that item from the dictionary. 

the `klass` `__module__` will be set to `Sk.globals.__name__` later anyway if it was not set in `locals`.

### DIFF
--- a/src/lib/array.js
+++ b/src/lib/array.js
@@ -152,9 +152,10 @@ $builtinmodule = function (name) {
         });
     };
 
+    mod.__name__ = new Sk.builtin.str('array');
+
     mod.array = Sk.misceval.buildClass(mod, array, "array", []);
 
-    mod.__name__ = new Sk.builtin.str('array');
 
     return mod;
 };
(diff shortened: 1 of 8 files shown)


s-cork:
Good point

I've added a bunch of `__name__`s to modules that were calling `misceval.buildClass` and so previously had their `__module__` (eventually) set to `__main__` (having initially set it to undefined). This is now corrected... e.g.

```python
>>> from array import array
>>> array
<class 'array.array'> # previously <class '__main__.array'>
```

meredydd (approved):
Nice - thanks for the extra tidy-ups!

PR #1155, opened 2020-08-05
change object.prototype.$r to be more inline with python

s-cork:
this pr adjusts `object.prototype.$r`

It basically takes the `else` part of `klass.prototype.$r` and moves it to `object.prototype.$r`
which i think is more correct.

it adjusts a previous test
```python
>>> object()
<object object> # previously <object> in python this is <object object at 0x123...>
```

while i'm here I adjust `klass.prototype.$r` and `tp$str` to use `callsimArray`
As well as adding a constant `Sk.builtin.str.$module`

### DIFF
--- a/src/type.js
+++ b/src/type.js
@@ -256,26 +256,18 @@ Sk.builtin.type = function (name, bases, dict) {
         klass["__name__"] = name;
         klass.sk$klass = true;
         klass.prototype["$r"] = function () {
-            var cname;
-            var mod;
-            var reprf = this.tp$getattr(Sk.builtin.str.$repr);
-            if (reprf !== undefined && reprf.im_func !== Sk.builtin.object.prototype["__repr__"]) {
-                return Sk.misceval.apply(reprf, undefined, undefined, undefined, []);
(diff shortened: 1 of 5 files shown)


src/type.js
```diff
@@ -257,26 +256,18 @@ Sk.builtin.type = function (name, bases, dict) {
...
+            const reprf = Sk.abstr.lookupSpecial(this, Sk.builtin.str.$repr);
+            if (reprf !== undefined && reprf !== Sk.builtin.object.prototype["__repr__"]) {
+                return Sk.misceval.callsimArray(reprf, [this]);
             }
 
             if ((klass.prototype.tp$base !== undefined) &&
-                (klass.prototype.tp$base !== Sk.builtin.object) &&
                 (klass.prototype.tp$base.prototype["$r"] !== undefined)) {
-                // If subclass of a builtin which is not object, use that class' repr
+                // If subclass of a builtin, use that class' repr
```
meredydd, line 266: This comment isn't correct -- this path triggers if `tp$base` isn't a builtin! (And has caused us significant grief in Anvil.) Can I ask you to correct it while you're here?

PR #1156, opened 2020-08-05
adjust call signature and implementation of sssiter

s-cork:
this pr aims to simplify `sssiter$` and make the function more readable.

it now takes two arguments
- `len - number`  (instead of a `length` or an `object` whose `v` property has a `length`) 
- `function` which expects javascript `index` at each iteration.
  (rather than using an argument `wrt` at each iteration (previously defined as the outcome of `Sk.builtin.asnum$(wrt)`), it instead relies on closure in the same way that `Sk.misceval.chain` would use closure.)

I don't think this is breaking since it's fairly obscure function buried in the core. Used for `str`, `tuple` and `list`

In the functions that use `sssiter$` I also adjust `const/let/var`

### DIFF
--- a/src/list.js
+++ b/src/list.js
@@ -293,9 +293,8 @@ Sk.builtin.list.prototype.list_subscript_ = function (index) {
         }
     } else if (index instanceof Sk.builtin.slice) {
         const ret = [];
-        const lst = this.v;
-        index.sssiter$(lst.length, (i) => {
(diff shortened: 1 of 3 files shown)


meredydd (approved):
A much-needed cleanup. LGTM.

meredydd:
I'm declaring this a minor change and merging.

PR #1157, opened 2020-08-05
small change to asserts

s-cork:
remove reference to `Sk.asserts.AssertionError`
and fix small type with `Sk.asserts.fail`


src/builtin.js
```diff
@@ -987,16 +987,8 @@ Sk.builtin.raw_input = function (prompt) {
 Sk.builtin.input = Sk.builtin.raw_input;
 
 Sk.builtin.jseval = function jseval (evalcode) {
```
albertjan, line 989: we should kill this or at least make it a compile options this is downright dangerous

PR #1158, opened 2020-08-06
frozenset hash

s-cork:
**Builtin on top of #1091**

I built this ontop of #1091 since it relies on `Sk.builtin.dict` having an `sk$asarray` method which that pr adds. 

I also add the `sk$asarray` method to `frozenset` and `set`

One of the benefits of a `frozenset` is that it is `hashable`. But in order to be `keys` of a `dict` the `hash` function must be consistent. 

This pr takes the general approach of python 2.7 https://github.com/python/cpython/blob/8d21aa21f2cbc6d50aab3f420bb23be1d081dac4/Objects/setobject.c#L775

Mostly because it was easier to follow 🤷‍♂️ 

I add the relevant tests from `python3` that test the `frozenset` `hash` function being consistent...

For the diff see https://github.com/s-cork/skulpt/pull/6

### DIFF
--- a/src/frozenset.js
+++ b/src/frozenset.js
@@ -61,6 +61,9 @@ Sk.builtin.frozenset.prototype["$r"] = function () {
     }
 };
 
+Sk.builtin.frozenset.prototype.sk$asarray = function () {
+    return this.v.sk$asarray();
+};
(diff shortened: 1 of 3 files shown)


meredydd (approved):
Nice. Small, useful, tested.

PR #1162, opened 2020-08-09
combine #989 #983

s-cork:
This pr is the aim to marry the unicode work in #983 and the bytes work in #989 
The diff to #989 can be seen here s-cork#7 (diff fixed to be more readable with latest changes rebased)
The approx diff with #983 can be seen at s-cork#10
_(it's not totally accurate since it was after a merge commit of #983 on top of #989 but it's approx accurate. 
You can see - the minor change to ast.js, 
change in `compile.js`
removal of `builtin.staticfunc` and `misceval.iterator` (additions in #983 branch)
adjustment to `builtin.ord` 
cleanse of bytes from `str.js`)_

**AST** 
changes here are from #983 
I made no change of the #983 algorithm apart from making sure the ascii characters were `< 0x7f` rather than `0xff`
This means that `b'¥'` is now correctly thrown as a `SyntaxError` which was not the case in #983.
#983 removed `decodeUtf8` in favour of `decodeEscape`
I'm happy to adjust this algorithm as per discussion - might need some suggestions here.

The motivation: it seemed the simplest way to marry the branches and allowed for correct encoding of bytes objects like `b"\xf0\x9f\x8d\x95"` which were part of the tests in #983.

Since there are no longer any surprising errors from `decodeUtf8` I remove the `try` `catch` statements in `parsestrplus`. 
`decodeEscape` throws the appropriate `ast_error` where `decodUtf8` might have failed without an `ast_error`

---

**Compile.js**
In `compile.js` I adjusted the approach from #983 and #989. 
Since the byte literal has already been tested for correctness I convert each character to it's `charCode` and use the fast path passing an array to the bytes constructor.

---

**Bytes**
(the changes to constructor may be more easily seen directly in the file compared to reading the diff)
Since there are about 10 ways to call the bytes constructor:
```
undefined, Uint8Array, Array of js integers, String, js integer,
str, iterable of int like objects, int, bytes instance, object with __bytes__
throw TypeError
```
I decoupled the internal JS calls from the python calls
and then further decoupled the encodings into a separate function... and then `ascii` encoding into its own function. 
I implement `__bytes__` (since it's part of the #983 tests and some additional cpython tests added) 
I add suspension awareness for calls to bytes with an iterator or  `__bytes__` method. (a couple of tests added to `test_suspensions` - suspension awareness part of #983 bytes constructor)
a couple of changes to constructor enhance some simple subclassing of bytes. (some addition cpython tests added)

small adjustments as additional cpython tests were added...
small change to `indices` function  allow `start`/`end` to be `None`.
small change to `bytes.decode` text signature since encoding has a default value
```python
>>> bytes.decode.__text_signature__
"($self, /, encoding='utf-8', errors='strict')"
```

edit:
I add a $jsstr method and use this to implement comparisons
```javascript
// akin to
this.$jsstr() < other.$jsstr();
```

---

**Str.js**
merged changes from #983 and cleansed bytes from the file
adjusted `startswith`/`endswith` in light of `"astral awareness"` since pr #1033 added the `start/end` arguments.

---

**builtin.js**
`ascii` function merged from #983 
`ord` adapted to take a single bytes object and cpython tests added

---

**Other tests and beyond the scope of this pr**
`mod`: `%` symbol not yet supported
`index/count`: the algorithm should raise `ValueError` for ints outside of `range(0, 256)`

I've added these tests but commented them out. 

I remove one test in #983 which was incorrect
```python
# self.assertEqual(str(b'\xf0\x9f\x8d\x95'), '\U0001f355')
```

other quick observation - future pr might wish to address `bytes.fromhex` which is a `classmethod_descriptor` 
```python
>>> type(bytes.__dict__['fromhex'])
classmethod_descriptor
```
so could be called like `b'abc'.fromhex('ff')` in the same way that `dict.fromkeys` can be called...  

kwarg support might also be the consideration of a future pr...
```python
>>> str(encoding='utf-8', errors='strict', object=b'foo')
'foo'
>>> bytes('foo', errors='strict', encoding='utf-8')
b'foo'
```

### DIFF
--- a/src/builtindict.js
+++ b/src/builtindict.js
@@ -123,6 +123,7 @@ Sk.setupObjects = function (py3) {
         delete Sk.builtins["xrange"];
         delete Sk.builtins["StandardError"];
         delete Sk.builtins["unicode"];
+        delete Sk.builtins["basestring"];
         Sk.builtins["bytes"] = Sk.builtin.bytes;
         delete Sk.builtin.str.prototype.decode;
     } else {
@@ -133,6 +134,7 @@ Sk.setupObjects = function (py3) {
         Sk.builtins["xrange"] = new Sk.builtin.func(Sk.builtin.xrange);
         Sk.builtins["StandardError"] = Sk.builtin.Exception;
         Sk.builtins["unicode"] = Sk.builtin.str;
+        Sk.builtins["basestring"] = Sk.builtin.str;
         delete Sk.builtins["bytes"];
     }
 };
--- a/src/builtindict.js
+++ b/src/builtindict.js
@@ -125,7 +125,6 @@ Sk.setupObjects = function (py3) {
         delete Sk.builtins["unicode"];
         delete Sk.builtins["basestring"];
         Sk.builtins["bytes"] = Sk.builtin.bytes;
-        delete Sk.builtin.str.prototype.decode;
         Sk.builtins["ascii"] = new Sk.builtin.func(Sk.builtin.ascii);
     } else {
         Sk.builtins["filter"] = new Sk.builtin.func(Sk.builtin.filter);
@@ -137,7 +136,6 @@ Sk.setupObjects = function (py3) {
         Sk.builtins["unicode"] = Sk.builtin.str;
         Sk.builtins["basestring"] = Sk.builtin.str;
         delete Sk.builtins["bytes"];
-        Sk.builtin.str.prototype.decode = Sk.builtin.bytes.prototype.decode;
         delete Sk.builtins["ascii"];
     }
 };
--- a/src/builtindict.js
+++ b/src/builtindict.js
@@ -123,6 +123,8 @@ Sk.setupObjects = function (py3) {
         delete Sk.builtins["xrange"];
         delete Sk.builtins["StandardError"];
         delete Sk.builtins["unicode"];
+        Sk.builtins["bytes"] = Sk.builtin.bytes;
+        delete Sk.builtin.str.prototype.decode;
     } else {
         Sk.builtins["filter"] = new Sk.builtin.func(Sk.builtin.filter);
         Sk.builtins["map"] = new Sk.builtin.func(Sk.builtin.map);
@@ -131,9 +133,7 @@ Sk.setupObjects = function (py3) {
         Sk.builtins["xrange"] = new Sk.builtin.func(Sk.builtin.xrange);
         Sk.builtins["StandardError"] = Sk.builtin.Exception;
         Sk.builtins["unicode"] = Sk.builtin.str;
-        if (Sk.builtins["bytes"]) {
-            delete Sk.builtins["bytes"];
-        }
+        delete Sk.builtins["bytes"];
     }
 };
 Sk.exportSymbol("Sk.setupObjects", Sk.setupObjects);
--- a/src/builtindict.js
+++ b/src/builtindict.js
@@ -124,6 +124,7 @@ Sk.setupObjects = function (py3) {
         delete Sk.builtins["StandardError"];
         delete Sk.builtins["unicode"];
         delete Sk.builtins["basestring"];
+        delete Sk.builtin.str.decode;
         Sk.builtins["bytes"] = Sk.builtin.bytes;
         Sk.builtins["ascii"] = new Sk.builtin.func(Sk.builtin.ascii);
     } else {
@@ -135,6 +136,7 @@ Sk.setupObjects = function (py3) {
         Sk.builtins["StandardError"] = Sk.builtin.Exception;
         Sk.builtins["unicode"] = Sk.builtin.str;
         Sk.builtins["basestring"] = Sk.builtin.str;
+        Sk.builtin.str.decode = Sk.builtin.str.$py2decode;
         delete Sk.builtins["bytes"];
         delete Sk.builtins["ascii"];
     }
--- a/src/builtindict.js
+++ b/src/builtindict.js
@@ -126,6 +126,7 @@ Sk.setupObjects = function (py3) {
         delete Sk.builtins["basestring"];
         Sk.builtins["bytes"] = Sk.builtin.bytes;
         delete Sk.builtin.str.prototype.decode;
+        Sk.builtins["ascii"] = new Sk.builtin.func(Sk.builtin.ascii);
     } else {
         Sk.builtins["filter"] = new Sk.builtin.func(Sk.builtin.filter);
         Sk.builtins["map"] = new Sk.builtin.func(Sk.builtin.map);
@@ -136,6 +137,8 @@ Sk.setupObjects = function (py3) {
         Sk.builtins["unicode"] = Sk.builtin.str;
         Sk.builtins["basestring"] = Sk.builtin.str;
         delete Sk.builtins["bytes"];
+        Sk.builtin.str.prototype.decode = Sk.builtin.bytes.prototype.decode;
+        delete Sk.builtins["ascii"];
     }
 };
 Sk.exportSymbol("Sk.setupObjects", Sk.setupObjects);
--- a/src/bytes.js
+++ b/src/bytes.js
@@ -78,6 +78,9 @@ Sk.builtin.bytes.prototype.__class__ = Sk.builtin.bytes;
 function strEncode(pyStr, encoding, errors) {
     const source = pyStr.$jsstr();
     encoding = normalizeEncoding(encoding);
+    if (!(errors === "strict" || errors === "ignore" || errors === "replace")) {
+        throw new Sk.builtin.NotImplementedError("'" + errors + "' error handling not implemented in Skulpt");
+    }
     let uint8;
     if (encoding === "ascii") {
         uint8 = encodeAscii(source, errors);
--- a/src/bytes.js
+++ b/src/bytes.js
@@ -115,6 +115,8 @@ function tp$new(args) {
     if (args.length <= 1) {
         pySource = args[0];
     } else {
+        // either encoding is a py object or errors is a py object - currently kwargs not supported
+        // will fail if encoding is not a string || errors is not a string || pySource is not a string
         [pySource, encoding, errors] = args;
         encoding = encoding || null;
         errors = errors || null;
--- a/src/bytes.js
+++ b/src/bytes.js
@@ -64,7 +64,9 @@ Sk.builtin.bytes = function (source) {
     } else if (typeof source === "number") {
         this.v = new Uint8Array(source);
     } else {
-        Sk.asserts.fail("got a " + Sk.abstr.typeName(source));
+        // fall through case for subclassing called by Sk.abstr.superConstructor
+        const bytes_obj = tp$new([...arguments]);
+        this.v = bytes_obj.v;
     }
 };
 
--- a/src/bytes.js
+++ b/src/bytes.js
@@ -48,7 +48,7 @@ Sk.builtin.bytes = function (source) {
     if (!(this instanceof Sk.builtin.bytes)) {
         // called from python
         Sk.builtin.pyCheckArgsLen("bytes", arguments.length, 0, 3);
-        return tp$new([...arguments]);
+        return newBytesFromPy(...arguments);
     }
 
     // deal with internal calls
@@ -65,7 +65,7 @@ Sk.builtin.bytes = function (source) {
         this.v = new Uint8Array(source);
     } else {
         // fall through case for subclassing called by Sk.abstr.superConstructor
-        const bytes_obj = tp$new([...arguments]);
+        const bytes_obj = newBytesFromPy(...arguments);
         this.v = bytes_obj.v;
     }
 };
@@ -106,7 +106,7 @@ function encodeAscii(source, errors) {
     return Uint8ArrayFromArray(data);
 }
 
-function tp$new(args) {
+function newBytesFromPy(...args) {
     let source,
         pySource,
         dunderBytes,
@@ -242,8 +242,10 @@ Sk.builtin.bytes.prototype["$r"] = function () {
 };
 
 Sk.builtin.bytes.prototype.mp$subscript = function (index) {
+    var ret;
+    var i;
     if (Sk.misceval.isIndex(index)) {
-        let i = Sk.misceval.asIndex(index);
+        i = Sk.misceval.asIndex(index);
         if (i !== undefined) {
             if (i < 0) {
                 i = this.v.byteLength + i;
@@ -254,7 +256,7 @@ Sk.builtin.bytes.prototype.mp$subscript = function (index) {
             return new Sk.builtin.int_(this.v[i]);
         }
     } else if (index instanceof Sk.builtin.slice) {
-        const ret = [];
+        ret = [];
         index.sssiter$(this.v.byteLength, (i) => {
             ret.push(this.v[i]);
         });
--- a/src/bytes.js
+++ b/src/bytes.js
@@ -867,7 +867,7 @@ Sk.builtin.bytes.prototype["rpartition"] = new Sk.builtin.func(function (self, s
     if (val === -1) {
         final1 = new Sk.builtin.bytes(0);
         final2 = new Sk.builtin.bytes(0);
-        final3 = new Sk.builtin.bytes(self);
+        final3 = self;
         return new Sk.builtin.tuple([final1, final2, final3]);
 
     }
@@ -1568,7 +1568,7 @@ Sk.builtin.bytes.prototype["splitlines"] = new Sk.builtin.func(function (self, k
                 eol = i;
             }
 
-            final.push(Sk.builtin.bytes(self.v.subarray(sol, eol)));
+            final.push(new Sk.builtin.bytes(self.v.subarray(sol, eol)));
 
             sol = rn ? i + 2 : i + 1;
             i = sol;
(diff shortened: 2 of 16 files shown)


src/builtindict.js
```diff
@@ -127,7 +134,10 @@ Sk.setupObjects = function (py3) {
         Sk.builtins["xrange"] = new Sk.builtin.func(Sk.builtin.xrange);
         Sk.builtins["StandardError"] = Sk.builtin.Exception;
         Sk.builtins["unicode"] = Sk.builtin.str;
+        Sk.builtins["basestring"] = Sk.builtin.str;
+        delete Sk.builtins["bytes"];
+        delete Sk.builtins["ascii"];
```
meredydd, line 139: Python 2's `str`s need to support `decode()` as well as `encode()` (doesn't have to be fast, it just has to work!)
I'd suggest a pretty direct "call `$jsstr()`, barf if it's not a byte-string, turn it into a `Uint8Array`, pass that to the encoding machinery."

src/str.js
```diff
@@ -999,6 +1087,20 @@ Sk.builtin.str.prototype["istitle"] = new Sk.builtin.func(function (self) {
...
+    encoding = encoding || Sk.builtin.str.$utf8;
+    Sk.builtin.pyCheckType("encoding", "string", Sk.builtin.checkString(encoding));
+    encoding = encoding.v;
+    if (errors !== undefined) {
+        Sk.builtin.pyCheckType("errors", "string", Sk.builtin.checkString(errors));
+        errors = errors.v;
+    } else {
+        errors = "strict";
+    }
+    return Sk.builtin.bytes.$strEncode(self, encoding, errors);
```
meredydd, line 1101: In Python 2 mode, we need to return a string here (ie `new Sk.builtin.str(<value>.$jsstr())`)

src/bytes.js
```diff
@@ -0,0 +1,1497 @@
...
+        num = leading + "0" + num;
+    } else {
+        num = leading + num;
+    }
+    return num;
+};
+
+
+Sk.builtin.bytes.prototype.$jsstr = function () {
+    return Decoder.decode(this.v);
```
meredydd, line 202: This will throw an error if the bytes object isn't valid UTF-8. In the contexts in which this is used, I suggest it would be more helpful to return a JS binary string (a JS string with character codes in the range 0-255). This will also make life rather easier for those of us who have to accept bytestrings from Python 2 and 3!

src/bytes.js
```diff
@@ -0,0 +1,1497 @@
...
+        if (!(other instanceof Sk.builtin.bytes)) {
+            return Sk.builtin.NotImplemented.NotImplemented$;
+        }
+        return f(this.v, other.v);
+    };
+}
+Sk.builtin.bytes.prototype.ob$lt = slotCompareUint8((v, w) => new Sk.builtin.bool(v < w));
+Sk.builtin.bytes.prototype.ob$le = slotCompareUint8((v, w) => new Sk.builtin.bool(v <= w));
+Sk.builtin.bytes.prototype.ob$gt = slotCompareUint8((v, w) => new Sk.builtin.bool(v > w));
+Sk.builtin.bytes.prototype.ob$ge = slotCompareUint8((v, w) => new Sk.builtin.bool(v >= w));
```
meredydd, line 302: These cute comparisons don't work, because they rely on stringification, and if you compare strings then `'10' < '9'`. For a quick and dirty approach you could call `$jsstr()` and compare the results...

src/bytes.js
```diff
@@ -0,0 +1,1497 @@
...
+    const source = pyStr.$jsstr();
+    encoding = normalizeEncoding(encoding);
+    if (!(errors === "strict" || errors === "ignore" || errors === "replace")) {
+        throw new Sk.builtin.NotImplementedError("'" + errors + "' error handling not implemented in Skulpt");
+    }
+    let uint8;
+    if (encoding === "ascii") {
+        uint8 = encodeAscii(source, errors);
+    } else if (encoding === "utf-8") {
+        uint8 = Encoder.encode(source);
```
meredydd, line 90: Do we need to take account of `errors` here? Or will an encode always succeed into UTF-8, even if it's got weird codepoints in it? (There are some weird reserved ones like the UTF-16 surrogate range)

src/str.js
```diff
@@ -599,39 +719,7 @@ Sk.builtin.str.prototype["index"] = new Sk.builtin.func(function (self, tgt, sta
...
-    idx = self.v.lastIndexOf(tgt.v, end);
-    idx = (idx !== end) ? idx : self.v.lastIndexOf(tgt.v, end - 1);
-    idx = ((idx >= start) && (idx < end)) ? idx : -1;
-
-    return new Sk.builtin.int_(idx);
-});
+Sk.builtin.str.prototype["rfind"] = mkFind(true);
 
 Sk.builtin.str.prototype["rindex"] = new Sk.builtin.func(function (self, tgt, start, end) {
     var idx;
```
meredydd, line 725: *(A few lines down from here, but GH won't let me put the comment there...)*

`startswith()` and `endswith()` now need to be astral-aware too!

src/bytes.js
s-cork: adjusted to loop throw and return the `String.fromCharCode`

src/bytes.js
s-cork: quick and dirty approach utilized. Can adjust if desired.

src/bytes.js
s-cork: @rixner - i pass this to you - equivalent line in #989 
https://github.com/skulpt/skulpt/blob/eba781b4ea810ee7100f18e0a7ac91dbedcfcc0e/src/bytes.js#L131

src/builtindict.js
s-cork: i think i've fixed it:

```
Sk.builtin.str.$py2decode = new Sk.builtin.func(function (self, encoding, errors) {
    Sk.builtin.pyCheckArgsLen("decode", arguments.length, 1, 3);
    // use a fast javastring constructor path in bytes constructor this.v = new TextEncoder.encode(source); 
    const pyBytes = new Sk.builtin.bytes(self.v); 
    return Sk.builtin.bytes.$decode(pyBytes, encoding, errors);
});
```

src/str.js
s-cork: have resolved this this with a `Sk.__future__` check

src/str.js
s-cork: have done this and extracted the indices function used by `mkFind`, `startswith`, `endswith` you might want to check this in s-cork#10

src/bytes.js
rixner: @meredydd do you have an example string that you think might fail?  I am fairly certain that this is guaranteed to always encode into utf-8.  The input string is not being interpreted as utf-8.  The bytes constructor is encoding a sequence of codepoints U+0000 through U+00FF into utf-8.  So, it's not possible to have a "weird codepoint" in the input.

All of the errors will occur on decode.  I tried a bunch of potentially invalid utf sequences in both Python and TextEncoder and I get the same results from both every time. I couldn't produce an error in Python.  I tried invalid utf-8 bytes, utf-16 surrogate pairs, etc.  So I am pretty sure this can't fail.

rixner:
I am okay with this.  However, I am uncomfortable with the constructor organization.  It has a few problems, in my opinion:

1. It is inconsistent with the constructor for every other builtin. This creates unnecessary cognitive load on people looking at the code base to try and decipher why this one is different.  I really think that consistency in the code base is important.
2. Intuitively, it feels backward to use Sk.builtin.bytes as the internal fast path and an auxiliary function as the Python path. We could just call the auxiliary function directly internally.  Again, though, that's completely inconsistent with the rest of the code base, so I am not in favor of doing so, I'm just saying the approach "feels" backwards to me.
3. I am leery of relying on the presence or absence of "new" to change the behavior of the constructor in this way.  I know we do type checking when new is missing, but somehow that seems less problematic to me (though, I'm not sure that's really a defensible argument).

I feel like this is something we should either be doing for all of the constructors or none of them.  And since no others do this, I don't think this PR should include it.

src/bytes.js
rixner: Wait, I think I'm wrong.  Looking into this further.

meredydd:
So, I jump entirely the other way on the constructor organisation. My PoV:

* I find the way we currently do builtin constructors is _extremely_ difficult to understand. The `dict` and `list` constructors on their own have cost me significant numbers of brain-cells. (I have already tried and given up a couple of times attempting to make `dict()` suspension-aware).

   `bytes` is even more extreme, because there are so many ways to construct the darn things: If all the 10 (ten!) possible ways to call the `bytes` constructor were laid out end-on-end, that function would be really quite unwieldy - especially if it mixed up "calls used by internal JS code" and "calls from Python" willy-nilly. I find Stu's version much easier to comprehend.

* This arrangement also makes a neat conceptual distinction between "the bits that Python calls" and "the bits that JS calls". It's always seemed odd to me that "construct a builtin from inside our JS implementation" hits the same code paths as "call the `bytes()` function from Python". If you're going to divide up a long constructor, this seems like the natural way to do it.

   * If these constructors are to be suspension-aware (which they do need to be - `dict()` not being suspendable is a regular source of headaches in Anvil), this distinction is particularly acute. There need to be two ways of calling this constructor:
       * `new bytes(x)`, which is called from JS, cannot suspend, and must be fast; and
       * `Sk.misceval.callsimOrSuspend(bytes, x)`, which is called from Python, must be able to suspend if `x`'s iterator suspends, and must be correct.


* This is actually a very minor shift in control flow. Stu has taken a (very) long if/else chain, and extracts a semantically coherent subset of cases into a separate function.

* The only "unusual" thing I can see is that Stu is branching based on whether we're calling with `new` or not.
  a. We already need this distinction (because it tells us whether we can suspend); and
  b. For performance reasons, we should already always use `new` whenever we build these objects from JS. (I can't remember whether you taught me that or I had to learn it the hard way with the profiler!) This code fails loudly if you do that, rather than causing quiet performance issues.

I think this boils down to an argument that "this is something we should either be doing for all of the [builtin] constructors" :)
I don't see a huge problem with starting with this one, because:

* We're implementing this one from scratch (so we don't have the "what if external users are already calling the constructor as a function?" question to contend with)
* The control-flow change is minor (factor out a few cases from a long if/else block)
* It makes it very clear which code paths can and cannot suspend

src/builtindict.js
s-cork: I think this needs a once over or some python2 tests to check it behaves as expected...

src/str.js
s-cork: as above. some py2 tests here might be desirable... i'm not sure the `__future__` flag is sufficient or if the `$jsstr()` does what it's supposed to...

s-cork:
the alternative approach seemed to be #983 suspension awareness gambits... 
but I didn't think would be the best approach here...

I mostly took inspiration from the `list` constructor which makes a similar distinction between internal calls and calls from py in order to determine if it is suspendible or not

s-cork:
@rixner regarding @meredydd request for `title, istitle, splitlines, (rsplit), expandtabs` let me know how you want to proceed with those.

src/bytes.js
rixner: Okay, I was wrong about how it was working, but I still believe it's not possible for this to fail. A javascript string can only contain valid unicode code points.  So, this just takes each code point and converts it into utf-8 bytes.  That should always work, unless there is an invalid code point in the input (I believe this to be impossible) or there is a code point that utf-8 can't represent (I believe that also cannot happen).

rixner:
@meredydd I won't argue about what you've lost time working on.  But, I will continue to argue that consistency is important, lest we **all** lose time.  If you want to argue that we should move in that direction, I'm not against it (I still say it's backwards in my mind, but whatever, I'm not going to fight about that - as long as they are all consistent, it's fine). But, I really don't like the idea of one constructor being organized completely differently than the others, regardless of how "complex" it is or is not.

rixner:
@s-cork it doesn't matter to me if you or I implement those methods.  I can do it tonight, or you can take care of it if you have the time.

src/bytes.js
rixner: This also seems how CPython is working.  So, I think they are compatible.  Anyone think I'm missing something?

s-cork:
re the constructor - here's some further thoughts and related battles i've had with `skulpt` constructors generally...

Hacking in Skulpt seems to have 4 main aims :
1. consistency with python
2. consistency with cpython
3. consistency with itself
4. keeping its user code working

- with regard to 1. Constructors in skulpt are not consistent with python. Yes they work, but if you dig they can't be consistent because they don't have `__new__`/`__init__` methods and so make no distinction between mutable and immutable types. 
Take `list` for example - it's `__new__` function does nothing but return an empty list in python... so there's no way skulpt can be consistent with python here. In python i can also (in theory) do `[1, 2, 3].__init__()` which will clear the list... but won't in skulpt.

- In light of 1 - skulpt constructors are also not consistent with cpython. cpython typically has a C constructor and then a `tp_new` and a `tp_init` slot...

- since all cpython constructors have a(n) internal C way(s) of creating an object and a py way...
  `newBytesFromPy` is effectively the py way of calling to `bytes` (in cpython
  https://github.com/python/cpython/blob/490c5426b1b23831d83d0c6b269858fb98450889/Objects/bytesobject.c#L2596)
  And then the constructor is a combination of things like `PyBytesFromSize` 
  (some other Cpython implementations have one py way and one C way)
  So I don't see this approach as less consistent with Cpython...

- with 3 and 4 the challenge here is to keep bytes suspension aware as per the requirements in #983 while keeping it consistent with the codebase. I'd argue that the way this was done in #983 is not consistent with the codebase. And would be more challenging for someone to look at and understand see file... 
  https://github.com/skulpt/skulpt/blob/17b6fe31a823161a639de7ee8a79bda83d67de1c/src/str.js#L11

- The way list (and now tuple) handle this is arguably cleaner by adding an argument to the constructor `canSuspend`...
yes `list` is consistent with other functions in skulpt that have a `canSuspend` argument but then it's not consistent with 
other constructors...
  And the problem with this `canSuspend` argument is that it's no longer consistent with python.
  If I subclass from a list, in skulpt I can pass as many arguments as I want.
```python
>>> class A(list): pass
>>> A([1,2,3,4], 5, 6, 7, 8, None)
[1,2,3,4] # in python this raises an exception
```

- I do think that because `list` distinguishes between python calls and javascript calls that the proposal in this pr is closer to consistency than #983. And while #989 is consistent it arguably breaks 4.

- With most things I'm happy to move in different directions here. I've also spent many brain cells on this - and I think what I'm proposing is a fairly reasonable way to do things, and more consistent than first impression 
  _(particularly when looked at through the frame of `list` which is the only other constructor that is suspension aware)_
  _(I could pass the extra `canSuspend` argument - but I didn't like how this breaks subclassing)_

- As a potential advantage for someone reading the code - it's now clear that internal constructor calls should really take the following... 
```
Uint8Array, Array of ints, JS String, Js Number, undefined
```
to do anything else is still possible, but would need to be considered more carefully.

rixner:
As I said in #989, I'm done messing with that PR, if everything thinks it's "done".   I remain convinced that the encoding can not ever have errors in it (when I initially wrote how it worked, I was working on multiple things at once, I should have known better than to think about unicode issues with half my attention).  I have given it much further thought and looked at how things work.  I am pretty confident that strings should always have valid code points in them and such code points can always be converted to utf-8, so there is no possibility for an error.

As for the constructor, I have given my concerns.  The primary one is that I don't like internal inconsistency.  For me, that makes it significantly harder to keep a mental model of what's going on within the code.  So, I think the issue is not that I fundamentally disagree with @s-cork's approach here (though I have the minor concerns that I raised), but more that this now makes bytes diverge from everything else. @s-cork, your recent arguments make total sense to me, so I am not going to strenuously object to moving in this direction (ideally, with *all* the constructors).

Ultimately, I want these pull requests both merged more than I want to debate whether we should implement the constructor like this now, wait for a future PR that modifies all of the constuctors, or enter a prolonged debate about what the constructors should look like. 9 months is more than long enough for these things to stagnate.  Let's just merge them and move forward.

s-cork:
i just did a quick rebase and tidy... removed a couple of unused variables, refactored `startswith`, `endswith` having had to do something similar in str with astral codepoints. (negating the use for the `negend` function)

shouting a bit louder if `pySource` is not a skulpt object i.e. we passed a javascript object and so most likely forgot to use `new` _(I feel like you were testing me @rixner by missing out `new` in the implementation of `splitlines` 😄 )_ - i add an extra message by checking the flag `sk$object` on `pySource` which will be on the prototypical chain of all `py3 objects` since everything (including `type objects`) inherits from `object` since #1131 


I would like @meredydd to check over the py2 compatibility requests as I'm not convinced by what i've done and would like some tests that would be expected to work in py2 mode so as not to break code for his end users...

meredydd:
This is now looking good to go!

meredydd:
...ah, wait, do we have the full set of `bytes` methods in this PR now? If not, it would be great to add them. If so, let's get this sucker merged!

rixner:
@meredydd I implemented them in #989.  So, even if they aren't here yet, they will be after #989 is merged.  So, you and/or @s-cork can merge #989, update #1162, then merge #1162 if you're good to go.

meredydd:
It's in! :sunglasses:

albertjan:
NOICE!

rixner:
Woo hoo!!

PR #1163, opened 2020-08-09
fix deque errortype

alexwenbj:
2 small fix.


src/lib/collections.js
```diff
@@ -891,7 +891,7 @@ var $builtinmodule = function (name) {
             var ret;
             n = Sk.builtin.asnum$(num);
             if(!Number.isInteger(n)){
-                throw new Sk.builtin.OverflowError("can't multiply sequence by non-int of type '" + Sk.abstr.typeName(num) + "'");
+                throw new Sk.builtin.TypeError("can't multiply sequence by non-int of type '" + Sk.abstr.typeName(num) + "'");
             }
```
s-cork, line 895: I wonder if there is also a case for the overflow error that needs checking/testing?

```python
>>> d = deque('abc')
>>> d.rotate(sys.maxsize + 1) # Number.MAX_SAFE_INTEGER
OverflowError: Python int too large ...
```

alexwenbj: Yeah,You are right.This remind me why I used the OverflowError before.:-) I will modify it later.

alexwenbj: @s-cork I think about it again and I guess this should be an issue for the Sk.builtin.int_ but not an issue for this rotate() method.Otherwise all the methods with int argument should make this check.

s-cork: To me it's more that because python has an `OverflowError`, we should too... the `OverflowError` checks makes sense in operations with indexing arrays... 

In terms of whether all methods should... again I see it as the case that if `python` raises one, then we should probably do too - and if we don't then we should have a good reason not to.

Within the scope of this pr I would say it's sufficient to re-include `OverflowErrors` in the two places you've removed them....

alexwenbj: Modified it.

PR #1166, opened 2020-08-12
mp_ass_subscript

s-cork:
I built this pr on top of #1091 and #1154 
#1154 is a prerequisite to this pr since without it - it meant that undefined was being set in a dictionary before the item existed which was rightly raising a `KeyError` with respect to this pr.
#1091 is not a pre-requisite and if that pr is rejected this pr can be adapted

---

this pr aims to move in the direction of cpython when it comes to `mp_ass_subscript` as well as moving away from python 2 when it comes to `sq_ass_slice`. 

`mp_ass_subscript`
in cpython the call signature has two options - `mp_ass_subscript(key)` or `mp_ass_subscript(key, value)`
if it's called without a value (or rather a value of `NULL`) then it means the item should be deleted. Here's a nice code snippet.
https://github.com/python/cpython/blob/0dfee33dfe6b4683016dc920ee934d3a2d7323a1/Objects/dictobject.c#L2182

in skulpt we've made up the slot - `mp$del_subscript` this is a skulptism... the primary aim of this pr is to find out whether removing this slot makes sense and decide whether to revert that skulptism in favour of the cpython version.

_nb:_
_cpython uses `NULL` to determine whether an item is to be deleted._
_skulpt has the choice between `undefined` or `null`... I chose `undefined`._

---

**moving away from python 2**
`sq$slice`, `sq$ass_slice` were removed in python 3. We are still referring to these slots in the core when we no longer need to.
instead we should (I believe) be doing `mp$ass_subscript(slice_obj, value)` indeed this is what cpython does.
https://github.com/python/cpython/blob/0dfee33dfe6b4683016dc920ee934d3a2d7323a1/Objects/abstract.c#L1817
vs the python 2.7 version
https://github.com/python/cpython/blob/8d21aa21f2cbc6d50aab3f420bb23be1d081dac4/Objects/abstract.c#L2099

I remove those two slots from the core. The end user has no access to these functions anyway since we don't seem to have ever included the relevant `dunder` methods for these to be accessed. 

with this in mind I adapt some of the `Sk.abstr.sequence_____` and related functions.

---

side note - @meredydd  - I don't see why we need the two functions:
(referencing you since you wrote them in originally...)
`Sk.misceval.assignSlice`
`Sk.misceval.applySlice`
They're not used in the core at all... I've not removed these but I think they possible should just be removed.

---

in `list.js` I do perhaps too much. I adjust the names of functions so that functions related to `mp$ass_subscript` have `$` symbols rather than a trailing underscore. I do this because in theory a trailing underscore function can be overridden by a subclass of a builtin where a `$` cannot. Of course this will probably never happen - but seems to me a possibility that shouldn't be open in the first place...

---

of course if the skulpt `mp$del_subscript` is the preferred way to do things I can adapt this pr as necessary and focus it only on removing `sq$slice` slots (or just close it). I do think, however, that we should aim to align ourselves with cpython where possible...

This pr will be much easier to read if/after #1091 and #1154 are merged


src/abstract.js
```diff
@@ -385,35 +374,15 @@ Sk.abstr.sequenceGetCountOf = function (seq, ob) {
...
 
 Sk.abstr.sequenceGetItem = function (seq, i, canSuspend) {
-    var seqtypename;
-    if (seq.mp$subscript) {
-        return seq.mp$subscript(i);
-    }
-
-    seqtypename = Sk.abstr.typeName(seq);
-    throw new Sk.builtin.TypeError("'" + seqtypename + "' object is unsubscriptable");
+    return Sk.abstr.objectGetItem(seq, i, canSuspend);
```
meredydd, line 377: If we're going properly break-y, we should just remove `Sk.abstr.sequenceGetItem` and friends!

(And good riddance.)

src/misceval.js
```diff
@@ -117,45 +117,19 @@ Sk.misceval.asIndex = function (o) {
  * return u[v:w]
  */
 Sk.misceval.applySlice = function (u, v, w, canSuspend) {
```
meredydd, line 119: Ditto - if we make this a breaking change, we can blow this one away too.

src/misceval.js
```diff
@@ -117,45 +117,19 @@ Sk.misceval.asIndex = function (o) {
...
-    }
-    return Sk.abstr.objectGetItem(u, new Sk.builtin.slice(v, w, null), canSuspend);
+    return Sk.abstr.objectGetItem(u, new Sk.builtin.slice(v, w), canSuspend);
 };
 Sk.exportSymbol("Sk.misceval.applySlice", Sk.misceval.applySlice);
 
 /**
  * u[v:w] = x
  */
 Sk.misceval.assignSlice = function (u, v, w, x, canSuspend) {
```
meredydd, line 127: Likewise.

meredydd (approved):
I don't have strong opinions over "assign undefined" vs "delete", so I'm happy to match how CPython does it. It does feel a little odd that these are different operations at the syntactic (`del` vs `x[y]=`) and dunder (`__delitem__` vs `__setitem__`) levels, and yet the same slot underneath...but if that's how CPython does it, it would be far from the worst weirdness we've ever imported from them.

I don't think you "changed too much" in `list.js` - I think we should standardise on using `$`s.

Massive cleanup of all the `sq$`... stuff too - very welcome! (This led me to wonder why we have both `tp$setitem` and `mp$ass_subscript`, but we don't have to clear up everything at once :) )

However: **This is probably a breaking change** for people who manipulate `dict` objects - I know it's likely to break Anvil for example, so should be called out in a changelog at least. (And possibly get held until after a "ready to break things" release from @bnmnetp, although my current feeling is that if that's going to happen it should happen _soon_.)

The upside is that, if it's a breaking change, we can clean up even more (eg from the `misceval` namespace!)

s-cork:
closed via #1092

PR #1167, opened 2020-08-13
bytes mod hash

s-cork:
extension of the bytes pr

this pr adds support for bytes `nb$remainder`
Rather than rewrite `nb$remainder` it piggybacks off `str.prototype.nb$remainder`

To fully support `nb$remainder` bytes objects must have the ability to be keys of dictionaries. 
The current `hash` function is not suitable since two `bytes objects` that are the same in python have different in `hash` values in skulpt.
```python
>>> hash(b'x') == hash(b'x')
False # should be True
```
Skulpt/Python `hash` function for `str` is consistent because of `interned`
in Python, the `hash` function for `bytes` is the same as for its `str` counterpart
```python
>>> hash('x')
-3753845217207833703
>>> hash(b'x')
-3753845217207833703
```
so for the `bytes` `hash` function it make sense to take the same value that is returned by the `hash` function when applied to its `str` counterpart.

```javascript
Sk.builtin.bytes.prototype.tp$hash = function () {
    return Sk.builtin.hash(new Sk.builtin.str(this.$jsstr()))
}
```

---

In order to piggy back off `str` I add `$baseType` onto the prototype of `bytes` and `str` - i'm not wedded to this notation. 

I think it's preferable to using `__class__` or `ob$type` because the `nb$remainder` should still work for instances of `str` and `bytes` and `__class__` and `ob$type` are overridden by subclasses where `this.$baseType` will not be overriden.

Other suggestions: `this.sk$base_type`, `this.sk$base`, `this.sk$builtin_base`, `this.sk$bltn_base`, `this.base$type`, `this.sk$bltn_subclass`.

---

I adjust the fast path for the constructor of `bytes` when given a `javascript string`.
This now must take a [javascript binary string](https://developer.mozilla.org/en-US/docs/Web/API/DOMString/Binary)
which means that each `charCode` must be between (0, 255) i.e. equivalent to each value for the resulting `Uint8Array`.

I think this is a sensible restriction for the fast path. 
It means that `new Sk.builtin.bytes(this.$jsstr())` has the same `Uint8Array` values as `this` 
I also think this is sensible change now since this `string` fast path is currently not used in the code base and so does not affect existing code.

### DIFF
--- a/src/bytes.js
+++ b/src/bytes.js
@@ -59,7 +59,18 @@ Sk.builtin.bytes = function (source, encoding, errors) {
         Sk.asserts.assert(source.every((x) => x >= 0 && x < 256), "bad internal call to bytes with array");
         this.v = Uint8ArrayFromArray(source);
     } else if (typeof source === "string") {
-        this.v = Encoder.encode(source);
+        // fast path must be binary string https://developer.mozilla.org/en-US/docs/Web/API/DOMString/Binary
+        // i.e. the reverse of this.$jsstr();
+        let cc;
+        const arr = [];
+        for (let i in source) {
+            cc = source.charCodeAt(i);
+            if (cc > 0xff) {
+                throw new Sk.builtin.UnicodeDecodeError("invalid string (possibly contains a unicode character)");
+            }
+            arr.push(cc);
+        }
+        this.v = Uint8ArrayFromArray(arr);
     } else if (typeof source === "number") {
         this.v = new Uint8Array(source);
     } else {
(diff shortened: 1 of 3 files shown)


meredydd:
This looks good! I agree with you that `$baseType` sounds like it's a generic piece of Skulpt rather than a way of telling which builtin class a user-supplied object might be subtyping - `sk$builtinBase` or something like that (plus a comment explaining what it's for) would be ideal.

s-cork:
changed to `sk$builtinBase`

meredydd (approved):
I like this. Especially for a feature that's relatively disfavoured in Python 3, I think that "turn it into a binary JS string, then use `str`'s implementation" is an appropriate level of effort/extra code. (Plus, it's a useful feature!)

PR #1170, opened 2020-08-16
Replace CRLF with LF

hjiawei:
Replease CRLF with LF for the following source to be consisitent with
other sources. No real code changes.

```
example/webglq2/q2bspload.py
src/lib/pythonds/basic/deque.py
src/lib/pythonds/basic/queue.py
src/lib/pythonds/basic/stack.py
```


rixner:
These files drive me nuts because of this.  Happily merging!

PR #1172, opened 2020-08-18
swapped ops as per cpython

s-cork:
This pr is for correctness
While mathematically the opposite of `<` is probably `>=`
the `swappedOp_` is to check an `op` from the `other`'s perpsective

```
A < B
# swapped op
B > A
```

This can be seen in the description above `swappedOps_` in `misceval.js`
https://github.com/skulpt/skulpt/blob/000ad6ca865538e5be938e33aa7b78420661a1fe/src/misceval.js#L202
 and in the cpython source code
https://github.com/python/cpython/blob/c3dd7e45cc5d36bbe2295c2840faabb5c75d83e4/Objects/object.c#L650

I found this bug while implementing the remaining cpython `unit_tests` for `set` on another branch
https://github.com/python/cpython/blob/c3dd7e45cc5d36bbe2295c2840faabb5c75d83e4/Lib/test/test_set.py#L620

I've also removed the unused swapped ops since they're not called and don't seem like they should be there. 
```python
a in b # doesn't have a swapped op
```


s-cork:
Which tests? 

I guess in skulpt it's never tested. Particularly since it would be pretty rare to hit that code and get a return other than `NotImplemented`. Most builtins would have already handled this before that case was hit I guess... 

The tests I've linked would fail if they were in skulpt but they're not part of our test suite (yet).

s-cork:
p.s. @albertjan i've amended the description since my explanation was poor. 
p.p.s. I would just implement the tests I linked to and put them in this pr... but doing so then exposes some more bugs surrounding `set` rich comparisons... I can make that within the scope of this pr if preferred...

albertjan:
I'm happy with this I was just surprised you were able to change these without running into a problem.

albertjan (approved):

meredydd (approved):
This is obviously correct. A test would be nice, but this is a strict improvement!

PR #1173, opened 2020-08-18
LookupError

s-cork:
`LookupError` was recently introduced with `bytes`
this pr tidies a couple of related loose ends

```python
>>> KeyError.__bases__ 
(LookupError,) 
```

When a `KeyError` is thrown from a dict lookup we should throw the `KeyError` with the actual `key` and not the `repr`

```python
d = {}
x = 1
try: 
  d[x]
except KeyError as e:
  print(e.args[0] is x) # currently in skulpt the dict key lookups return the `repr` and so this would fail
```


s-cork:
added some set tests from cpython. Since we use dict under the hood the tests are relevant and fail on master.

PR #1174, opened 2020-08-18
raise error with non Exception should raise an error

s-cork:
```python
>> raise 1
TypeError: exceptions must derive from BaseException
# skulpt undefined undefined
```

This fixes a small issue that the user should only be able to raise `Exceptions` if they are instances of `BaseException` or classes that derive from `BaseException`


albertjan (approved):

albertjan:
would be great to have a test 😬

albertjan:
Small change merging.

PR #1175, opened 2020-08-19
fix assertRaises related bug

s-cork:
working on implementing these tests from `dict_tests.py` led to a couple of small bugs relating to `assertRaises`

https://github.com/python/cpython/blob/c3dd7e45cc5d36bbe2295c2840faabb5c75d83e4/Lib/test/test_dict.py#L767

https://github.com/python/cpython/blob/c3dd7e45cc5d36bbe2295c2840faabb5c75d83e4/Lib/test/test_dict.py#L201


albertjan (approved):
small change merging

PR #1183, opened 2020-09-09
Allow underscores in numeric literals and constructors

### LINKED ISSUE #1058: Underscore in Numeric Literals not Supported
[PEP515](https://www.python.org/dev/peps/pep-0515/) defines a syntax for underscores in numeric literals to e.g. work as a thousands separator.

`value = 12_000_000`
should be equivalent to
`value = 12000000`

As far as I can tell, this is not supported in Skulpt yet. Is that correct?

shreyasminocha:
~~Addresses part of #1058 (only `int`s, not all numeric literals).~~

Resolves #1058

Do I add unit tests to `test_underscores` in `test/unit3/test_int.py`? Something like:

```py
self.assertEqual(int('10_000_000'), 10_000_000)
self.assertEqual(int('0xCAFE_F00D', 16), 0xCAFE_F00D)
[...]
```


s-cork:
yes please! 

Best to take tests from Cpython where possible. 
https://github.com/python/cpython/blob/3ff6975e2c0af0399467f234b2e307cc76efcfa9/Lib/test/test_grammar.py#L183
and/or
https://github.com/python/cpython/blob/3ff6975e2c0af0399467f234b2e307cc76efcfa9/Lib/lib2to3/tests/data/py3_test_grammar.py#L75

adapting the tests where skulpt doesn't support `eval` 

Since skulpt doesn't have a `unit3/test_grammar.py` you might decide:
1. creating this file and include the relevant tests there, or 
2. add the tests to `unit3/test_int_literal.py` with a comment about where the tests are from. 

if you go with 1. then for bonus points you can add the non-int underscore literal tests and comment them out for future prs.

s-cork:
p.s. if you want to support the object literal as well (`1_000_000`) the code to look at would be 

https://github.com/skulpt/skulpt/blob/f81be786f7c32d3d40ff991559f02ee6a0e30ae7/src/ast.js#L2676

shreyasminocha:
While editing `src/ast.js`, I realized I'd need to handle floats, longs and complex numbers too (to prevent things like `3e+_3` from working). So I've expanded the scope of this PR to include floats, longs and complex numbers too.

I've created `unit3/test_grammar.py` taking cues from the CPython tests. The tests for the constructors `int`, `long`, `float`, and `complex` are complete[1], but I've commented out the tests for literals in `test_grammar.py` since those depend on the unsupported `eval`. Is there another way to test the literals? Testing `INVALID_UNDERSCORE_LITERALS` is especially tricky. Maybe that could wait until `eval` is implemented?

I've commented out:

```py
    # '0_7', # allowed as parameter to `int()`
```

... in `INVALID_UNDERSCORE_LITERALS` because, as the comment notes, it's allowed as a parameter to `int()`. We'll want to make sure that `0_7` is an invalid literal once we have a way to test those (see note above).

I can only hope I haven't missed any edge cases (there are many! lol), but it seems that all of the tests from CPython are passing, so that's a good sign.

[1]: Except I've commented-out the `eval` lines: `# self.assertEqual(float(lit), eval(lit))`

src/ast.js
```diff
@@ -2679,6 +2679,16 @@ function parsenumber (c, s, lineno) {
...
+    if (/_[eE]|[eE]_|\._|j_/.test(s)) {
+        throw new Sk.builtin.SyntaxError("invalid syntax", c.c_filename, lineno);
+    }
+
+    if (/_\.|[+-]_|^0_\D|_j/.test(s)) {
+        throw new Sk.builtin.SyntaxError("invalid decimal literal", c.c_filename, lineno);
+    }
+    
+    s = s.replace(/_(?=[^_])/g, "");
+
```
s-cork, line 2691: worth hositing the regex in this pr - i.e. defining them outside the scope of the function so that they are only compiled once. This can improve performance.

test/unit3/test_grammar.py
```diff
@@ -0,0 +1,92 @@
...
+]
+
+# class TokenTests(unittest.TestCase):
+    # def test_underscore_literals(self):
+    #     for lit in VALID_UNDERSCORE_LITERALS:
+    #         self.assertEqual(eval(lit), eval(lit.replace('_', '')))
+    #     for lit in INVALID_UNDERSCORE_LITERALS:
+    #         self.assertRaises(SyntaxError, eval, lit)
+    #     # Sanity check: no literal begins with an underscore
+    #     self.assertRaises(NameError, eval, "_0")
```
s-cork, line 89: for the eval problem - you could do something obscure like
```python
eval_alt = "Sk.importMainWithBody('test_literals', false, {0})"

for lit in INVALID_UNDERSCORE_LITERALS:
  try:
    jseval(eval_alt.format(lit))
  except Exception as e: # these will be ExternalErrors
    self.assertIn("SyntaxError", repr(e))
  else:
    self.fail(f"SyntaxError not raised for {lit}")

```

but it's pretty obscure! and only really works for testing that these Errors get raised (ish)

test/unit3/test_int.py
```diff
@@ -214,9 +215,20 @@ def test_basic(self):
...
+            if any(ch in lit for ch in '.eEjJ'):
+                continue
+            self.assertRaises(ValueError, int, lit, 0)
+
         # Additional test cases with bases != 0, only for the constructor:
-        # self.assertEqual(int("1_00", 3), 9)
-        # self.assertEqual(int("0_100"), 100)  # not valid as a literal!
+        self.assertEqual(int("1_00", 3), 9)
+        self.assertEqual(int("0_100"), 100)  # not valid as a literal!
+        # self.assertEqual(int(b"1_00"), 100)  # byte underscore
```
s-cork, line 231: happy to keep bytes outside of the scope of this pr

if you want it inside the scope of this pr - then...

in the constructors for `float/int` adjust the test for string e.g. 

https://github.com/skulpt/skulpt/blob/master/src/float.js#L32

https://github.com/skulpt/skulpt/blob/master/src/int.js#L60

becomes 
```
if (Sk.builtin.checkString(x) || Sk.builtin.checkBytes(x))
```

then also 
https://github.com/skulpt/skulpt/blob/f81be786f7c32d3d40ff991559f02ee6a0e30ae7/src/int.js#L66
needs to be
```javascript
 val = Sk.str2number(x.$jsstr())
```

(though an issue for this approach is the error messages won't be accurate for invalid byte strings)

test/unit3/test_grammar.py
```diff
@@ -0,0 +1,113 @@
...
+        self.assertEqual(0b_0, 0b0)
+        self.assertEqual(0x_f, 0xf)
+        self.assertEqual(0o_5, 0o5)
+        self.assertEqual(1_00_00j, 10000j)
+        self.assertEqual(1_00_00.5j, 10000.5j)
+        self.assertEqual(1_00_00e5_1j, 10000e51j)
+        self.assertEqual(.1_4j, .14j)
+        self.assertEqual((1_2.5+3_3j), (12.5+33j))
+        
+        eval_alt = "Sk.importMainWithBody('test_literals', false, {0})"
```
s-cork, line 93: Sorry in my original suggestion for this idea I missed the quotes around `'{0}'`

test/unit3/test_grammar.py
```diff
@@ -0,0 +1,113 @@
...
+
+        for lit in VALID_UNDERSCORE_LITERALS:
+            try:
+                jseval(eval_alt.format(lit))
+            except Exception as e: # these will be ExternalErrors
```
s-cork, line 98: If these are valid then they won't raise exceptions? So maybe if an exception is raised then this test should be something like?:
```
except Exception:
    self.fail(f'{lit} was not compiled')
```

But for these valid literals you might skip this niche eval test and just put them as literals inside a function since if they are valid they should just compile. 🤔

```python
def test_valid_literals(self):
  1_0
  # etc
```

s-cork (commented):
sorry about the eval suggestion.

test/unit3/test_grammar.py
s-cork: and turns out these won't be ExternalErrors - they were only external Errors because I forgot these quotes when I was thinking about it/testing it 🙄 🤦 

so you later you can do:
```python
for lit in INVALID_UNDERSCORE_LITERALS:
    self.assertRaises(SyntaxError, jseval, eval_alt.format(lit))
```

I've used a similar approach in #1185:
https://github.com/skulpt/skulpt/blob/47088ed7142b2908abddf3e3daf196b89333a0d5/test/unit3/test_unpack.py#L182-L212

s-cork (commented):
two things i might consider in this pr for performance:
- To accommodate underscores we now do regex tests for every string - I wonder if it's better to make this a slow path:
```javascript
if (s.inclues("_") {
  // take the slow path and do the the underscore regex tests/replaces
}
// otherwise take the fast path (no underscores)
```

- might be worth hoisting the regex in `int.js/float.js/complex.js` as well - and same slow path idea as above.

shreyasminocha:
Done :+1:

> might be worth hoisting the regex in int.js/float.js/complex.js as well

I've hoisted the regexes I added. Let me know if you want me to hoist the old ones (`/^-inf$/i`, `/^[+]?inf$/i`, `/^[-+]?nan$/i`) as well.

s-cork:
Great. I think you can add that same slow path to ast and complex. 

Re hoisting regex that was already there - from what I understand it can be an easy performance win - but no one will complain if you leave what's there as is. So up to you. 

I had another thought about the `test_underscore_literals`. It would be worth doing `self.assertEqual(1_00, 100)` etc for all the literals that are now compiled so as to check the values are also correct.

src/ast.js
```diff
@@ -2673,12 +2673,27 @@ function parsestrplus (c, n) {
     }
 }
 
+const invalidSyntax = /_[eE]|[eE]_|\._|j_/;
+const invalidDecimalLiteral = /_\.|[+-]_|^0_\D|_j/;
```
s-cork, line 2677: what am i missing here? - are these two regex tests not part of the slow path too? They're checking that underscores don't immediately proceed/follow something?

src/ast.js
```diff
@@ -2673,12 +2673,27 @@ function parsestrplus (c, n) {
...
     var tmp;
     var end = s.charAt(s.length - 1);
 
+    if (invalidSyntax.test(s)) {
+        throw new Sk.builtin.SyntaxError("invalid syntax", c.c_filename, lineno);
+    }
+
+    if (invalidDecimalLiteral.test(s)) {
+        throw new Sk.builtin.SyntaxError("invalid decimal literal", c.c_filename, lineno);
+    }
```
s-cork, line 2691: part of the slow path?

src/complex.js
```diff
@@ -271,6 +273,10 @@ Sk.builtin.complex.complex_subtype_from_string = function (val) {
     if (val.indexOf("\0") !== -1 || val.length === 0 || val === "") {
         throw new Sk.builtin.ValueError("complex() arg is a malformed string");
     }
+    
+    if (invalidUnderscores.test(val)) {
+        throw new Sk.builtin.ValueError("could not convert string to complex: '" + val + "'");
+    }
```
s-cork, line 279: part of the slow path?

test/unit3/test_grammar.py
```diff
@@ -0,0 +1,116 @@
...
+        self.assertEqual(.1_4e1, .14e1)
+        self.assertEqual(0b_0, 0b0)
+        self.assertEqual(0x_f, 0xf)
+        self.assertEqual(0o_5, 0o5)
+        self.assertEqual(1_00_00j, 10000j)
+        self.assertEqual(1_00_00.5j, 10000.5j)
+        self.assertEqual(1_00_00e5_1j, 10000e51j)
+        self.assertEqual(.1_4j, .14j)
+        self.assertEqual((1_2.5+3_3j), (12.5+33j))
+        self.assertEqual((.5_6j), (.56j))
```
s-cork, line 106: 👍

s-cork (commented):
thanks for the changes - let me know if I've missed something here.

src/ast.js
shreyasminocha: Oops lol, I missed that. Sorry, I've been tired haha

s-cork (approved):
supports numeric literals closes #1058, has tests, lgtm @skulpt/committers

src/ast.js
```diff
@@ -2673,12 +2673,27 @@ function parsestrplus (c, n) {
...
+const invalidDecimalLiteral = /_\.|[+-]_|^0_\D|_j/;
+const validUnderscores = /_(?=[^_])/g;
 function parsenumber (c, s, lineno) {
     var neg;
     var val;
     var tmp;
     var end = s.charAt(s.length - 1);
-
+    
+    if (s.includes("_")) {
```
meredydd, line 2685: Do we have a string polyfill for `includes()`? (I think we're still aiming to support IE10, aren't we?)

meredydd (changes requested):
One change - I think this will break IE10 (we still have some IE users!)

Other than that, looks good to merge!

src/ast.js
shreyasminocha: We could check this with [`indexOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf#Checking_occurrences) instead, sure.

meredydd (approved):
Nice work. In it goes!

PR #1184, opened 2020-09-09
:bug: fix string divmod error #1182

### LINKED ISSUE #1182: a bug for old-style format
For example:
```
X=1234
print("05d" % X)  #<-Yes, this is an incorrect statement.
```
It should raise an error:`TypeError: not all arguments converted during string formatting`,but Skulpt print out `05d`

Dustyposa:
fix #1182 
Reference implementation by [this](https://github.com/python/cpython/blob/master/Objects/unicodeobject.c#L15450)


src/str.js
```diff
@@ -1363,6 +1363,9 @@ Sk.builtin.str.prototype.nb$remainder = function (rhs) {
             return "%";
         }
     };
+    if (index < rhs.sq$length()) {
+        throw Sk.builtin.TypeError("not all arguments converted during string formattin");
+    }
```
s-cork, line 1368: can we fix this message (`'formatting'`) and use `'new'`
Also make sure that this passes the closure compiler linting checks, which seems to have failed somewhere

Dustyposa: Oh, when tests use python2 grammar, has some error like:
```shell
FAILED: (test/run/t166.py)
-----
print "%x" % 10
print "%x" % 10L
print "%x" % 100000000000L
print "%o" % 10
print "%o" % 10L
print "%o" % 100000000000L
print "%d" % 10
print "%d" % 10L
print "%d" % 100000000000L

-----
GOT:
-----
EXCEPTION: TypeError: not all arguments converted during string formattin on line 1

-----
WANTED:
-----
a
a
174876e800
12
12
1351035564000
10
10
100000000000


```
I need to check it.

s-cork:
I think this breaks on mappings now: 
```python
>>> '%(a)s' % {'a': 'foo', 'b': 'bar'}
TypeError: not all arguments converted during string formatting
# should be 'foo'
```

Dustyposa:
> I think this breaks on mappings now:
> 
> ```python
> >>> '%(a)s' % {'a': 'foo', 'b': 'bar'}
> TypeError: not all arguments converted during string formatting
> # should be 'foo'
> ```

got it. It's look need a flag to distinguish between `dict` and `tuple`

src/str.js
```diff
@@ -1364,7 +1364,7 @@ Sk.builtin.str.prototype.nb$remainder = function (rhs) {
         }
     };
     ret = this.$jsstr().replace(regex, replFunc);
-    if (index < Sk.builtin.asnum$(Sk.builtin.len(rhs))) {
+    if (rhs instanceof Sk.builtin.tuple && index < Sk.builtin.asnum$(Sk.builtin.len(rhs))) {
```
s-cork, line 1367: I think since we know this is a tuple we can use `rhs.sq$length()` which will return a js number.

Dustyposa: Oh, yes! I for get it :(

src/str.js
```diff
@@ -1107,7 +1107,7 @@ Sk.builtin.str.prototype.nb$remainder = function (rhs) {
         fieldWidth = Sk.builtin.asnum$(fieldWidth);
         precision = Sk.builtin.asnum$(precision);
 
-        if ((mappingKey === undefined || mappingKey === "") && conversionType != "%") {
+        if ((mappingKey === undefined || mappingKey !== "") && conversionType != "%") {
```
s-cork, line 1110: what's the reason for this change?

Dustyposa: oh. my miss. That's work for `dict`, now don't need it.....

s-cork (approved):
LGTM

albertjan:
Thanks @Dustyposa

PR #1185, opened 2020-09-10
starred assignment unpacking

### LINKED ISSUE #1063: extended iterable unpacking
https://www.python.org/dev/peps/pep-3132/

currently not supported in skulpt

```
>>> a, *b, c = range(5)
>>> a
0
>>> c
4
>>> b
[1, 2, 3]
```

s-cork:
This pr aims to add support for starred assignment - close #1063 

```python
>>> a, *b, c = 1, 2, 3, 4, 5
>>> a
1
>>> b
[2, 3, 4]
>>> c
5
```

general approaches were taken from Cpython's ast and compile files

It even works on weird nested assignment! who knew that was a thing...
```python
>>> a, *[b, *[c, *d]], e = range(6)
>>> a, b, c, d, e
(0, 1, 2, [3, 4] 5)
```

---

**TODO**
- [x] add tests

### DIFF
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -482,6 +482,7 @@ Sk.abstr.sequenceSetSlice = function (seq, i1, i2, x) {
  * // Sk.abstr.sequenceUncpack(seq, 3, 3, false)
  * // return [int_(1), int_(2), int_(3)]
  * 
+ * 
  * a, *b, c = 1,2,3,4 
  * // Sk.abstr.sequenceUncpack(seq, 1, 2, true)
  * // return [int_(1), list(int_(2), int_(3)), int_(4)]
@@ -508,7 +509,7 @@ Sk.abstr.sequenceUnpack = function (seq, breakIdx, numvals, hasStar) {
     }
     const starred = [];
     for (let i = it.tp$iternext(); i !== undefined; i = it.tp$iternext()) {
-        starred.push[i];
+        starred.push(i);
     }
     const end = starred.length + breakIdx - numvals;
     if (end < 0) {
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -469,29 +469,53 @@ Sk.abstr.sequenceSetSlice = function (seq, i1, i2, x) {
     }
 };
 
-// seq - Python object to unpack
-// n   - JavaScript number of items to unpack
-Sk.abstr.sequenceUnpack = function (seq, n) {
-    var res = [];
-    var it, i;
-
-    if (!Sk.builtin.checkIterable(seq)) {
-        throw new Sk.builtin.TypeError("'" + Sk.abstr.typeName(seq) + "' object is not iterable");
+/**
+ * 
+ * @param {*} seq the iterable to unpack
+ * @param {*} breakIdx either the starred index or the number of elements to unpack if no star
+ * @param {*} numvals the total number of un-starred indices
+ * @param {*} hasStar is there a starred index
+ * 
+ * this function is used in compile code to unpack a sequence to an assignment statement
+ * e.g.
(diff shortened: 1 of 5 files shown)


src/ast.js
```diff
@@ -160,6 +160,10 @@ function setContext (c, e, ctx, n) {
             }
             e.ctx = ctx;
             break;
+        case Sk.astnodes.Starred:
+            e.ctx = ctx;
+            setContext(c, e.value, ctx, n);
+            break;
```
s-cork, line 166: see:
https://github.com/python/cpython/blob/3.8/Python/ast.c#L1169

src/compile.js
```diff
@@ -917,7 +934,14 @@ Compiler.prototype.vexpr = function (e, data, augvar, augsubs) {
...
         case Sk.astnodes.Starred:
-            break;
+            switch (e.ctx) {
+                case Sk.astnodes.Store:
+                    /* In all legitimate cases, the Starred node was already replaced
+                     * by compiler_list/compiler_tuple. XXX: is that okay? */
+                    throw new Sk.builtin.SyntaxError("starred assignment target must be in a list or tuple", this.filename, e.lineno);
+                default:
+                    throw new Sk.builtin.SyntaxError("can't use starred expression here", this.filename, e.lineno);
+            }
```
s-cork, line 944: see
https://github.com/python/cpython/blob/3.8/Python/compile.c#L4977-L4987

albertjan (approved):
awesome

src/abstract.js
```diff
@@ -469,29 +469,57 @@ Sk.abstr.sequenceSetSlice = function (seq, i1, i2, x) {
...
+ */
+Sk.abstr.sequenceUnpack = function (seq, breakIdx, numvals, hasStar) {
     if (!Sk.builtin.checkIterable(seq)) {
-        throw new Sk.builtin.TypeError("'" + Sk.abstr.typeName(seq) + "' object is not iterable");
+        throw new Sk.builtin.TypeError("cannot unpack non-iterable " + Sk.abstr.typeName(seq) + " object");
+    }
+    const it = Sk.abstr.iter(seq);
+    const res = [];
+    for (let i = 0; i < breakIdx; i++) {
+        const nxt = it.tp$iternext();
```
meredydd, line 498: Could I be cheeky and suggest that, if/while we're doing this, we should make this function suspension-aware?
(All of the Anvil uses I'm most excited about are suspendable...)

test/unit3/test_unpack.py
```diff
@@ -106,6 +119,100 @@ def raise_bad_error2():
...
+            'a, *b, c, *d, e = range(10)' : (SyntaxError, 'multiple starred expressions in assignment'),
+            '[*b, *c] = range(10)': (SyntaxError, 'multiple starred expressions in assignment'),
+            'a,*b,*c,*d = range(4)': (SyntaxError, 'multiple starred expressions in assignment'),
+            '*a = range(10)': (SyntaxError, 'starred assignment target must be in a list or tuple'),
+            '*a' : (SyntaxError, 'can\'t use starred expression here'),
+            '*1': (SyntaxError, 'can\'t use starred expression here'),
+            'x = *a': (SyntaxError, 'can\'t use starred expression here'),
+
+        }
+        jseval("Sk.retainGlobals = true") # use globals from this module
```
meredydd, line 203: We have a `jseval` function? Yuck! (But also, wow.)

meredydd (approved):
This looks nice! I do have one request - how would you feel about making `Sk.abstr.sequenceUnpack` suspension-aware?

src/abstract.js
s-cork: done!

PR #1186, opened 2020-09-12
Implement the 'webbrowser' module

archibate:
Hi! This is my first PR for this cool project, looking forward for more chances!

## How to implement:

- Like `antigravity.js` does, using `window.open(url, "_blank")` for `webbrowser.open(url)`.
- A `TypeError` will be raised if `url` is not `str`.

## Questions / TODOs:

- [ ] How can I write a test for this? Currently this module only works in browser and our tests seems to be running on `node`?
- [ ] Will we implement `webbrowser.get`, `webbrowser.open_new`, `webbrowser.open_new_tab` too? And how?
- [ ] Is that possible to open a browser in Node.js?
- [x] Import and use `webbrowser` for `antigravity.py`.

Please let me know if I'm making things correct, thanks in advance!

### DIFF
--- a/src/lib/webbrowser.js
+++ b/src/lib/webbrowser.js
@@ -20,19 +20,42 @@ var $builtinmodule = function(name){
     });
 
     mod.open_new = new Sk.builtin.func(function open_new(url) {
-        Sk.builtin.pyCheckArgsLen("open_new", arguments.length + 1, 1, 1);
+        Sk.builtin.pyCheckArgsLen("open_new", arguments.length, 1, 1);
         return open_tab(url);
     });
 
     mod.open_new_tab = new Sk.builtin.func(function open_new_tab(url) {
-        Sk.builtin.pyCheckArgsLen("open_new_tab", arguments.length + 1, 1, 1);
+        Sk.builtin.pyCheckArgsLen("open_new_tab", arguments.length, 1, 1);
         return open_tab(url);
     });
 
+    function dflbrowser($gbl, $loc) {
+        $loc.__init__ = new Sk.builtin.func(function __init__(self) {
+            return Sk.builtin.none.none$;
+        });
+
+        $loc.open = new Sk.builtin.func(function open(self, url) {
+            Sk.builtin.pyCheckArgsLen("open", arguments.length, 2, 4);
+            return open_tab(url);
+        });
+
+        $loc.open_new = new Sk.builtin.func(function open_new(self, url) {
+            Sk.builtin.pyCheckArgsLen("open_new", arguments.length, 2, 2);
+            return open_tab(url);
+        });
+
+        $loc.open_new_tab = new Sk.builtin.func(function open_new_tab(self, url) {
+            Sk.builtin.pyCheckArgsLen("open_new_tab", arguments.length, 2, 2);
+            return open_tab(url);
+        });
+    }
+
+    mod.DefaultBrowser = Sk.misceval.buildClass(mod, dflbrowser, "DefaultBrowser", []);
+
     mod.get = new Sk.builtin.func(function get() {
         Sk.builtin.pyCheckArgsLen("get", arguments.length, 0, 1);
-        return $builtinmodule;
-    }
+        return Sk.misceval.callsimArray(mod.DefaultBrowser, []);
+    });
 
     return mod;
 };
--- a/src/lib/webbrowser.js
+++ b/src/lib/webbrowser.js
@@ -14,6 +14,8 @@ var $builtinmodule = function(name){
         return Sk.builtin.bool.true$;
     }
 
+    mod.__name__ = new Sk.builtin.str("webbrowser");
+
     mod.open = new Sk.builtin.func(function open(url) {
         Sk.builtin.pyCheckArgsLen("open", arguments.length + 1, 1, 3);
         return open_tab(url);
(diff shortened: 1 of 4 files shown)


src/lib/webbrowser.js
```diff
@@ -0,0 +1,19 @@
...
+        if (!(url instanceof Sk.builtin.str))
+            throw new Sk.builtin.TypeError("webbrowser.open expects 'str' for 'url'");
+
+        if (!inBrowser)
+            return false;
+
+        url = url.$jsstr();
+        window.open(url, "_blank");
+        return true;
+    });
```
s-cork, line 17: a coupe of things here that might be useful. 

Make sure you're returning javascript booleans inside a python function.
These should be `Sk.builtin.bool.false$` and `Sk.builtin.bool.true$` respectively

Worth giving the anonymous function a name. 
```javascript
mod.open = new Sk.builtin.func(function open() {
```
This way when the user uses introspection they will correctly see the name of the function. 

Style preferences for `if` statements with curly braces `{}`

You can use the the call signature 

```javascript
new Sk.builtin.func(function open(url) {
```

s-cork (changes requested):
I guess it would be nice to have `open` and `open_new` and `open_new_tab` but I don't see their functionality as being particularly important and can have the same behaviour as `open`. 

When I run the same 3 functions on my laptop they do exactly the same thing so I see no reason that skulpt shouldn't too...

perhaps you can do something like
```javascript
function open_tab(url) {
   // do the url type check and the window.open(url, "_blank")
}

mod.open = new Sk.builtin.funct(function open(url) {
    // check the args Len
    return open_tab(url);
};
mod.open_new = new Sk.builtin.funct(function open_new(url) {
    // check the args Len
    return open_tab(url);
};
mod.open_new_tab = new Sk.builtin.funct(function open_new_tab(url) {
    // check the args Len
    return open_tab(url);
};
```

You're call!

re the tests - like you said it's tricky to test this one in node 
there is precedence here - `document` isn't really tested that I can tell. 

nice PR 😄

src/lib/webbrowser.js
```diff
@@ -12,6 +12,22 @@ var $builtinmodule = function(name){
...
+        return open_tab(url);
+    });
+
+    mod.open_new = new Sk.builtin.func(function open_new(url) {
+        Sk.builtin.pyCheckArgsLen("open_new", arguments.length + 1, 1, 3);
+        return open_tab(url);
+    });
+
+    mod.open_new_tab = new Sk.builtin.func(function open_new_tab(url) {
+        Sk.builtin.pyCheckArgsLen("open_new_tab", arguments.length + 1, 1, 3);
```
s-cork, line 28: for `open` I think it should be 
```javascript
Sk.builtin.pyCheckArgsLen("open", arguments.length, 1, 3); 
// min 1 arg up to 3 args (even though we ignore the other 2)
```
and for the `open_new/open_new_tab` functions - looking at python's call signature - it should be
```javascript
Sk.builtin.pyCheckArgsLen("open_new", arguments.length, 1, 1); 
// exactly 1 arg
```

src/lib/webbrowser.js
```diff
@@ -0,0 +1,38 @@
...
+    });
+
+    mod.open_new_tab = new Sk.builtin.func(function open_new_tab(url) {
+        Sk.builtin.pyCheckArgsLen("open_new_tab", arguments.length + 1, 1, 1);
+        return open_tab(url);
+    });
+
+    mod.get = new Sk.builtin.func(function get() {
+        Sk.builtin.pyCheckArgsLen("get", arguments.length, 0, 1);
+        return $builtinmodule;
```
archibate, line 34: It seems that `webbrowser.get(using=None)` returns a browser object (seems they support multiple browsers?) which also have `open/open_new/open_new_tab` as methods.
However we only have one choice for browser JS.. so I'd like to return the module it self, am I doing correct by `return $builtinmodule;`?

s-cork: I don't think we need to implement this. 

If you want to then I would probably create a class object with the same methods as the module
here's the BaseBrowser class from Cpython
https://github.com/python/cpython/blob/c75330605d4795850ec74fdc4d69aa5d92f76c00/Lib/webbrowser.py#L139

For a simple class object written in JS I'd checkout the `re.js` file which has a couple.

archibate: Done, thank for the instructions!

BTW, I also noticed that many functions in `re.js` are not following the rules we're now using in `webbrowser.js`:
1. `function some_name(...)`
2. `Sk.builtin.pyCheckArgsLen`
2. `Sk.builtin.pyCheckType`

Which may result in confusing error message when users using them wrongly. Is this intentional for simplicity?

s-cork: Yes - there are two approaches to define a function. 

The approach in this module or the approach in re.js - that approach might do the equivalent of

```javascript
function open_new(url) {
  // only check type 
}
open_new.co_varnames = ["url"]
mod.open_new = new Sk.builtin.func(open_new)
```

the above tells the compiler that there is only one are called url. 

It has the benefit that the end user can do `url='foo'` as they can in python. But it's not the most intuitive interface.

It gets more complex the more complicated your call signature is.

s-cork (approved):
LGTM 👍

s-cork:
@archibate could you add 

```
mod.__name__ = new Sk.builtin.str("webbrowser")
```

Since you're passing this namespace to the buildClass function it means that `DefaultBrowser` will then get the correct `__module__` and resulting `repr` - (`< class webbrowser.DefaultBrowser>`).

Related to #1154

meredydd (approved):
Delightful!

PR #1189, opened 2020-09-23
remove String.fromcodepoint and Math.hypot polyfills - es5 build includes these already

s-cork:
since #889 we have different builds available. 
Anyone wishing to support older browsers that are not ES6 compliant should use the ES5 build option. 

You can see that webpack already polyfills `fromCodePoint` in our es5 build so we don't need to include this in the core - this might be a breaking change for someone somewhere.

https://github.com/skulpt/skulpt-dist/blob/6d321a5465e8877937a76d701b00ecddadd945e4/skulpt.es5.min.js#L13


s-cork:
close via #1092

Issue #695, opened 2017-07-06
Finish Processing.py


eah13:
Based on a large volume of user requests, we're considering a push to give processing.py much greater coverage of Processing's [python mode](http://py.processing.org/reference/).

As a review: Processing is a Java-based visual programming toolset for making animations, designed for teaching/learning.  C-style syntax sucks, so there's a Python mode that is essentially a Python to Java transpiler that exposes the base language's capabilities via Python syntax.  None of this works on the web but thanks to Atwood's law there's a JS project that attempts to replicate it.  Brad put it into Skulpt because it's a great teaching tool.

Brad's original implementation takes the right approach, I think, of wrapping [processing.js](http://processingjs.org/) in Skulpt (which is of course the analog to what the original Python mode does anyways).  If we completed this work, coverage would then theoretically be equivalent to processing.js's, which is [everything not listed in red here I think](http://processingjs.org/reference/). The supported processing.js version could also bump from 1.4.4 to 1.4.8.

Aside:  There's a newer JS based project, [p5.js](https://p5js.org/), but it's probably a worse choice since its goal is not to replicate Processing's capabilities and there's no standard Python mode.  

So, the basic plan is to finish what Brad started with a goal of full coverage of whatever the Venn diagram of processing.js and Python's Processing mode is.  

For discussion:

* Who's interested in using this if we built it?  Ditto for contributing.
* Any better ideas than the preliminary plan above?
* This could stay a part of the stdlib for backwards compatibility, but could also make sense as a third party lib if there are no objections to that.  The benefit would be smaller package size as the full module might add several hundred lines (at least).  The drawback would be that anyone who's got processing-based user could would need to configure the 3rd party lib to update skulpt.  I'm OK either way.
* What to do with the existing `processing` package?  Seems like we could keep it and build on it but if there are any things @bnmnetp would have done differently we've got a chance to do that now.  If it seems best we could even consider a full rewrite but there would probably need to be a really good reason to do so.
* Any other considerations?

Various processing issues that may be addressed directly or rendered moot with a rewrite (from a simple search of open issues - I didn't check to see if they've been fixed but not closed):
#564
#531 
#378 
#375 
#172

swirly:
Hello
I don't know if this issue is still revelant. processing.py was made exclusively for python 2 and will not be ported to python 3, due to it's underlying conception with jython.

[...]

PR #1191, opened 2020-09-23
ideas for new ffi api

s-cork:
related to ideas from #144 

I'm working on a branch that allows for easier interaction between skulpt objects and javascript objects
namely: 
- if it's skulpt iterable then it's javascript iterable
- doing `String(skulpt_object)` will do the obvious thing
- doing `Number(skulpt object)` will also do the obvious thing

e.g. 

Screen Shot 2020-09-23 at 22 42 31

---
To get these functions to work I override `toString` and `valueOf` for the obvious types.
Then implement [`Symbol.iterator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/iterator) on skulpt iterators. 
(it's only about 20 lines of code!)

There are probably too many functions and it could be slimmed down 
I'm curious about what others think of the ideas. 
Which are too cute, which are worth pursuing, which should be dropped, which I've removed that have value.


s-cork:
I've added the code here for the `proxy` object idea that allows arbitrary js objects to be passed to skulpt
(#695)

s-cork:
I've rebased and adjusted this api proposal with some example modules that implement it.

s-cork:
@tabreturn @nickmcintyre i'm wondering if you could have a look at the p5 implementation in this branch.

I've amended it today based on your examples from glitch [2D](https://glitch.com/edit/#!/invincible-voltaic-kitchen) and [3D](https://glitch.com/edit/#!/caramel-fixed-pony). And it seems to work pretty well.
(Remove `environment.` from 3D. Replace `PVector` with `createVector` in 2D. Use `Sk.canvas` in place of `Sk.p5Sketch`)

The advantage here is that it's 70 lines of code vs #1278  1700 lines of code.

nickmcintyre:
@s-cork this is great! Just opened a pull request to help complete the p5 module.

Your ffi API saves a huge amount of effort hand-rolling a wrapper. Any idea when it will be merged?

asbl:
Hello,

i love the idea of the ffi-api, but I have problems with it: I tried integrating ffi/p5 into an ide but I can't stop p5 programms because there are no suspensions which I can use, yieldLimit is ignored

s-cork:
interesting - thanks for the feedback. that's probably a problem with the p5 implementation rather than the ffi api. 
I'll take a look when I get a chance and see if there's an obvious way to do it from the p5 api

s-cork:
@asbl  - i've added an api for working with the p5 instances.
If you create `Sk.p5 = {node: HTMLElement for the container}` this will be used in place of Sk.canvas;

After the p5 module is imported you'll get:
- `Sk.p5.instance` - the current instance running
- `Sk.p5.kill` - kill the process for the current instance

asbl:
Thanks. It works!

Another problem are runtime errors which occur in the external lib. I was not able to catch error messages from p5 so far (but I didn't had enough time yet, so maybe there is an way but I didn't found it yet - Is there a way to catch the uncaught errors that are displayed in the console?). It would be great if you can get meaningful error message from the external library.

s-cork:
Two ways. 
1. Make sure you've set the uncaughtException hook in your Sk.configure call. (This pr's p5 implementation calls that when possible). 
2. Use `window.onerror` and possibly `window.onunhandledrejection` which you can override. Check MDN for the docs on those.

Hopefully that'll allow you to take care of all the exceptions.

asbl:
Thanx. I added this function in Sk.configure as handler for Sk.uncaughtException:

```
    handle_exception(error) {
        var Sk = this;
        var error_message = error.args.v[0].$mangled
        var lineno = error.traceback[0].lineno
        var colno = error.traceback[0].colno
        Sk.output("Error: " + error_message + " on line " + lineno)
    }
```

s-cork:
I think you can also do:
```javascript
    handle_exception(error) {
        Sk.output(error.toString());
    }
```
Which should be the same thing.

meredydd:
Hi all,

So, this PR has languished for a while, but:

* All the feedback on the new capabilities is positive
* We've shipped it in Anvil for some time, to rave reviews

I approve this change in principle, and propose shipping it once @s-cork updates the branch -- with the one proviso that the `p5` module is really out of scope (it's very cool, but it's more of a demo of what's possible with the new FFI than something we want to carry as part of our standard library).

Any @skulpt/committers who wish to object, speak now or forever hold your peace...

nickmcintyre:
@meredydd As a teacher, having turtle graphics and a sketchbook available by default would be *really* nice to have. One thought is to simply swap the legacy/incomplete Processing.js module for p5.js. It seems like #695 has also languished for a while and I never got around to completing [p5.sk](https://github.com/nickmcintyre/p5.sk) 😅

meredydd (approved):
I _really_ enjoyed reading this code. It is remarkably elegant.

Also battle-tested in production; let's ship it.

PR #1193, opened 2020-09-25
fix #1192

### LINKED ISSUE #1192: A bug in re module introduced by new Skulpt version
In this new version, the re.I/re.M will raise an exception,for example:
`matchObj = re.match( r'(.*) are (.*?) .*', line, re.I|re.M)`
raise an error:
`unsupported operand type(s) for BitOr: '<invalid type>' and '<invalid type>'`

alexwenbj:
Change the re.M and re.I to Skulpt type.


alexwenbj:
It is strange this changed the behavior of match().

alexwenbj:
fix #1192

PR #1194, opened 2020-09-26
fix #1192 re constants type

### LINKED ISSUE #1192: A bug in re module introduced by new Skulpt version

alexwenbj:
fix re module constants type


alexwenbj:
fix #1192

s-cork (approved):
lgtm - a failing test would be 👍 .

alexwenbj:
Not needed anymore because of a better implementation #1206 .

PR #1196, opened 2020-09-28
functools implementation with tests

s-cork:
this is a branch off of #1092 - separate branch so as not to continue to pollute that pr. 
It will fail on Travis at present since it heavily relies upon #1092 but passes all tests on my local machine.

it introduce a partial implementation of functools. 
- `partial`
- `partialmethod`
- `update_wrapper`
- `wraps`
- `cmp_to_key`
- `total_ordering`
- `reduce` (suspension aware)

All the above are written as native skulpt functions/classes 
This is different to Cpython, which mixes between C implementations and python implementations.

All the above use the full set of cpython tests (minus `pickling` and other unsupported tests).

some additions were made to the core and the diff with #1092 can be found here https://github.com/s-cork/skulpt/pull/13
most of those changes are relatively minor.

### DIFF
--- a/src/lib/functools.js
+++ b/src/lib/functools.js
@@ -27,7 +27,7 @@ function $builtinmodule() {
         WRAPPER_UPDATES: new Sk.builtin.tuple([new Sk.builtin.str("__dict__")]),
 
         /**@todo */
-        lru_cache: proxyFail("lru_cace"),
+        lru_cache: proxyFail("lru_cache"),
         TopologicalSorter: proxyFail("TopologicalSorter"),
         CycleError: proxyFail("CycleError"),
         singledispatch: proxyFail("singledispatch"),
--- a/src/lib/functools.js
+++ b/src/lib/functools.js
@@ -1,0 +1,459 @@
+function $builtinmodule() {
+    const functools = {
+        __name__: new Sk.builtin.str("functools"),
+        __doc__: new Sk.builtin.str("Tools for working with functions and callable objects"),
+        __all__: new Sk.builtin.list(
+            [
+                "update_wrapper",
+                "wraps",
+                "WRAPPER_ASSIGNMENTS",
+                "WRAPPER_UPDATES",
+                "total_ordering",
+                "cmp_to_key",
+                "lru_cache" /**@todo lru_cache */,
+                "reduce",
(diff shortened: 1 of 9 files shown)


s-cork:
rebased and added features required for this implementation of functools.

2 helper functions added
`Sk.abstr.keywordArrayToPyDict` and `Sk.abstr.keywordArrayFromPyDict`

**Some bug fixes were addressed that failed the functools tests**

`__name__` and `__module__` (no change to the api but adjusted for correctness)
`tp$name` should include a `.` for builtin type objects if they come from a module (see [cpython docs](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_name))
`hp$name` was added - also used in cpython - for `klass` objects.
related functions were adjusted accordingly:
- `Sk.abstr.typeName` 
- `Sk.abstr.buildNativeClass`
- `Sk.builtin.type.prototype.__name__/__module__`

`__dict__` getset descriptor was adjusted so that it couldn't be deleted for builtins (whereas it can for `klass` objects)

`__doc__`/`__module__` were changed in `function.js` so that they could be set. Required for functools.

src/lib/functools.js
```diff
@@ -0,0 +1,459 @@
...
+                "cached_property" /**@todo cached_property */,
+            ].map((x) => new Sk.builtin.str(x))
+        ),
+        WRAPPER_ASSIGNMENTS: new Sk.builtin.tuple(
+            ["__module__", "__name__", "__qualname__", "__doc__" /*"__annotations__"*/].map((x) => new Sk.builtin.str(x))
+        ),
+        WRAPPER_UPDATES: new Sk.builtin.tuple([new Sk.builtin.str("__dict__")]),
+
+        /**@todo */
+        lru_cache: proxyFail("lru_cace"),
```
bnmnetp, line 30: typo?? should be `lru_cache` ??

PS - would be very cool to have this functionality!

bnmnetp (approved):

src/lib/functools.js
s-cork: yes it would! I've never used those tools personally so it wasn't something I needed - when this gets merged I'll report an issue for the missing features for someone (or me) to have a go at.  will fix the typo

PR #1206, opened 2020-10-05
an update regex module

s-cork:
This is an attempt to enhance the features in skulpt's regex implementation.
- `re.RegexFlag` implemented - with a metatype!
  `list(re.RegexFlag) # [re.ASCII, re.UNICODE, ...]`
- `re.RegexFlag` is a subtype of `int`

missing methods implemented - `sub`, `subn`, `escape`, `findall`, `finditer`
fixes `re.split` for capture groups

support for named capture groups
```python
>>> re.compile('(?P<foo>)').groups
1
>>> re.compile('(?P<foo>)').groupindex
{foo: 1}
```

tests taken from cpython and as many tests as possible uncommented.
There are still unsupported features since python and javascript regex is different!
There also may be some edge cases where compilation between python and javascript is different.
If anyone has an extensive regex implementation in their skulpt - it would be great to see if there are any problems in the wild that are not covered by the skulpt test suite that i've missed. 

I've avoided using `negative look behinds` in the module but do allow users to use negative lookbehinds since most browsers now support those.

Of the 98 tests added to this pr - master currently passes 19 of them.

`bytes` is not supported as part of the `re` module but it shouldn't be too difficult to include here.

builds on top of #1092 with some small changes to that branch.
Only re changes included here which won't pass master since it relies upon features in that branch.

### DIFF
--- a/src/lib/re.js
+++ b/src/lib/re.js
@@ -1,7 +1,36 @@
 function $builtinmodule(name) {
+
+    const {
+        builtin: {
+            dict: pyDict,
+            str: pyStr,
+            list: pyList,
+            int_: pyInt,
+            type: pyType,
+            tuple: pyTuple,
+            mappingproxy: pyMappingProxy,
+            slice: pySlice,
+            none: {none$: pyNone},
+            NotImplemented: { NotImplemented$: pyNotImplemented },
+            Exception,
+            OverflowError,
+            IndexError,
+            TypeError,
+            ValueError,
+            checkInt,
+            checkString,
+            checkCallable,
+            hex,
+        },
+        abstr: { buildNativeClass, typeName, checkOneArg, numberBinOp, copyKeywordToNamedArgs, setUpModuleMethods },
+        misceval: { iterator: pyIterator, objectRepr, asIndexSized, isIndex, callsimArray: pyCall },
+    } = Sk;
+
+
+
     const re = {
-        __name__: new Sk.builtin.str("re"),
-        __all__: new Sk.builtin.list(
+        __name__: new pyStr("re"),
+        __all__: new pyList(
             [
                 "match",
                 "fullmatch",
@@ -32,25 +61,25 @@ function $builtinmodule(name) {
                 "DOTALL",
                 "VERBOSE",
                 "UNICODE",
-            ].map((x) => new Sk.builtin.str(x))
+            ].map((x) => new pyStr(x))
         ),
     };
 
     // cached flags
     const _value2member = {};
 
-    const RegexFlagMeta = Sk.abstr.buildNativeClass("RegexFlagMeta", {
+    const RegexFlagMeta = buildNativeClass("RegexFlagMeta", {
         constructor: function RegexFlagMeta() {},
-        base: Sk.builtin.type,
+        base: pyType,
         slots: {
             tp$iter() {
                 const members = Object.values(_members)[Symbol.iterator]();
-                return new Sk.misceval.iterator(() => members.next().value);
+                return new pyIterator(() => members.next().value);
             },
             sq$contains(flag) {
                 if (!(flag instanceof this)) {
-                    throw new Sk.builtin.TypeError(
-                        "unsupported operand type(s) for 'in': '" + Sk.abstr.typeName(flag) + "' and '" + Sk.abstr.typeName(this) + "'"
+                    throw new TypeError(
+                        "unsupported operand type(s) for 'in': '" + typeName(flag) + "' and '" + typeName(this) + "'"
                     );
                 }
                 return Object.values(_members).includes(flag);
@@ -58,9 +87,9 @@ function $builtinmodule(name) {
         },
     });
 
-    re.RegexFlag = Sk.abstr.buildNativeClass("RegexFlag", {
+    re.RegexFlag = buildNativeClass("RegexFlag", {
         meta: RegexFlagMeta,
-        base: Sk.builtin.int_,
+        base: pyInt,
         constructor: function RegexFlag(value) {
             const member = _value2member[value];
             if (member) {
@@ -72,10 +101,10 @@ function $builtinmodule(name) {
 
         slots: {
             tp$new(args, kwargs) {
-                Sk.abstr.checkOneArg("RegexFlag", args, kwargs);
+                checkOneArg("RegexFlag", args, kwargs);
                 const value = args[0].valueOf();
-                if (!Sk.builtin.checkInt(value)) {
-                    throw new Sk.builtin.ValueError(Sk.abstr.objectRepr(value) + " is not a valid RegexFlag");
+                if (!checkInt(value)) {
+                    throw new ValueError(objectRepr(value) + " is not a valid RegexFlag");
                 }
                 return new re.RegexFlag(value);
             },
@@ -93,18 +122,18 @@ function $builtinmodule(name) {
                     }
                 });
                 if (value) {
-                    members.push(Sk.builtin.hex(value).toString());
+                    members.push(hex(value).toString());
                 }
                 let res = members.join("|");
 
                 if (neg) {
                     res = members.length > 1 ? "~(" + res + ")" : "~" + res;
                 }
-                return new Sk.builtin.str(res);
+                return new pyStr(res);
             },
             sq$contains(flag) {
                 if (!(flag instanceof re.RegexFlag)) {
-                    throw new Sk.builtin.TypeError("'in' requires a RegexFlag not " + Sk.abstr.typeName(flag));
+                    throw new TypeError("'in' requires a RegexFlag not " + typeName(flag));
                 }
                 return this.nb$and(flag) === flag;
             },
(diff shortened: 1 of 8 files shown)


src/lib/re.js
```diff
@@ -1,435 +1,1128 @@
...
-
-        if (pat.match(/\$/)) {
-            var newline_at_end = new RegExp(/\n$/);
-            if (str.match(newline_at_end)) {
-                str = str.slice(0, -1);
+    const flagFails = Object.entries({
+        "cannot use LOCALE flag with a str pattern": re.L,
+        "ASCII and UNICODE flags are incompatible": new re.RegexFlag(re.A.valueOf() | re.U.valueOf()),
+    });
+    const inline_regex = /\(\?([i|s|a|m|u|x]+)\)/g;
```
alexwenbj, line 207: `[i|s|a|m|u|x]` may be `[isamux]`?

s-cork: Yeah. Good spot! Thanks.

alexwenbj:
@s-cork ,which PR compatible with this PR? I cloned your s-cork:prototypical_getsets_mappingproxy PR and took this re implementation, but not worked, many test cases in this PR failed(in my editor, not in the build process).Thank you.

s-cork:
here's the branch with the additions of the `re` module and `functools` module ontop of #1092 
https://github.com/s-cork/skulpt/tree/pr/re_update

and here's the diff with #1092 
https://github.com/s-cork/skulpt/pull/14

alexwenbj:
OK, I will have a try.

alexwenbj:
I have a question: in this PR, the pattern not permit to use bytes, but in current Skulpt version, we have made it support bytes. for example,this case:
`self.assertTypedEqual(re.sub(b'y', b'a', b'xyz'), b'xaz')`
failed.

s-cork:
to support bytes here for the `re` module wouldn't be too challenging. But it does require some extra checks since you can't mix between bytes patterns and bytes search/tests/sub. 

Also the return value should be the same as the input.

So for `re.sub(b'y', b'a', b'xyz')` we would need to first check that all the arguments were `bytes` objects and then convert them to `jsstrings` and then make sure the return value was also a `bytes` object. 

It's certainly on my todo - I'll get round to that at some point and add it to this pr.

alexwenbj:
To make these test cases pass, maybe you can use the solution I now using:
```
# Conditional group reference.
        self.assertIsNone(re.match(r'(?:(a)|(x))b(?<=(?(2)x|c))c', 'abc'))
        # self.assertIsNone(re.match(r'(?:(a)|(x))b(?<=(?(2)b|x))c', 'abc')) #-->
        self.assertTrue(re.match(r'(?:(a)|(x))b(?<=(?(2)x|b))c', 'abc'))
        self.assertIsNone(re.match(r'(?:(a)|(x))b(?<=(?(1)c|x))c', 'abc'))
        self.assertTrue(re.match(r'(?:(a)|(x))b(?<=(?(1)b|x))c', 'abc'))
#         # Group used before defined.
        # self.assertRaises(re.error, re.compile, r'(a)b(?<=(?(2)b|x))(c)')
        self.assertIsNone(re.match(r'(a)b(?<=(?(1)c|x))(c)', 'abc'))
        self.assertTrue(re.match(r'(a)b(?<=(?(1)b|x))(c)', 'abc'))
```
Here is my solution, cause the JS limitation, it is not perfect, but somehow accomplish the target:
```
var convert_yes_no_pattern = function(pat){
        // convert python yes_pattern/no_pattern to JS format
        pat = pat.replace(/\(\?\((?<num>\d+)\)(?<str>.+)(\))/g,function(...params){
            var p = params.pop();
            return "(?:(\\"+p.num+")?"+p.str+")";
        });
        pat = pat.replace(/\(\?\((?=(?<back_name>\w+))\k<back_name>\)/g,function(g0,bn,offset,input,{back_name}){
            return "(?:(\\k<"+back_name+">)?";
        });
        return pat;
    }
```
FYI.

s-cork:
@alexwenbj - i'm not sure that quite works e.g.
```
re.compile('(?:(a)|(c))b(?(1)a|c)') # should only match 'aba' or 'cbc'
```
but I think the suggested regex matches all of `'abc'`, `'aba'`, `'cba'` and `'cbc'`

alexwenbj:
Yes, cause the javascript limitation.

s-cork:
update on this pr. I'm using `String.prototype.matchAll` which is an `es2020` feature. So it needs polyfilling for Travis as it's not available in node < 12.

I updated google closure compiler to the latest version since the version we are using only goes up to es2019 polyfills.

After updating google closure compiler, `String.prototype.matchAll` got polyfilled 🏆  BUT the polyfill is incorrectly implemented from closure 🤦 . 
 I've raised an issue on their github - for now this pr should wait in the wings until either - closure fix their polyfill/I remove `matchAll` from the module.

https://github.com/google/closure-compiler/issues/3776

meredydd (approved):
This looks good - and is well overdue to be rebased and merged!

A few notes, lines are in `regex.js`:

238 does this match flags in all of the regex? Is that right? Turns out the answer is yes, until 3.11 - add comment to this line

304 add comment "use negative lookbehinds *in py_to_js_regex*"

355 Why is this handled differently in unicode mode? Is it OK to silently fall back like this or should we be failing/logging/something?

716 incorrect comment, we *keep* all flags except "g"

1008 why only two digit match numbers? The version after \g is more general. I assume you know what you're doing here (I'd buy that this is shorthand notation), just need to check.

s-cork:
1008: yeah - if you search \number in the python regex docs https://docs.python.org/3/library/re.html#regular-expression-syntax:
This special sequence can only be used to match one of the first 99 groups...

meredydd:
> 1008: yeah - if you search \number in the python regex docs https://docs.python.org/3/library/re.html#regular-expression-syntax: This special sequence can only be used to match one of the first 99 groups...

Oh, Python!

bnmnetp:
If @meredydd gives this a thumbs up, I'm happy to merge based on my quick look at the code.

bnmnetp:
I see the approval now!

Can we tag or use this thread to make a list of some of the other PRs that we should get merged.  It seems that our process is keeping a bunch of PRs lingering in an unmerged state.

If tests are passing why are we not just merging some of these? Is there controversy or are we just suffering from having enough people who feel qualified to review? I barely feel qualified to review most of what is being done these days, but I would like to get this functionality in there and out to my users!!

meredydd:
I think you're right, and our process is causing a bunch of lingering PRs - which is in turn making others more reluctant to contribute, because the prospect of seeing their changes get merged is so distance. I would suggest lowering the review bar for changes that are well tested, and lowering it further when it's filling in a glaring gap or fixing a bug, and especially where the code changes are well-contained (ie don't risk messing up other code). This change:

* Replaced a somewhat janky hack
* Was well tested, against the CPython tests
* Was very self-contained, in both functionality and codebase (only really affecting the re Javascript module, which is implemented in a single source file), limiting the impact even if it had produced

I propose that however we set the bar, we set it such that @s-cork could have merged this on his own recognizance, or at very most had another committer cast an eye over it and say "yep, seems reasonable".

bnmnetp:
I agree @meredydd 

I just looked at the fraction PR and I think that if @s-cork is ready to merge it, I say go ahead.

I can take the requirement for a review off, or, I'm also happy to have the submitter review their own work.  At least for those who have the ability to merge.  Which is me, @meredydd , @s-cork , @rixner, @albertjan, @ebertmi , @LeszekSwirski 

It may be that I should remove a couple names from that list.  People who have made wonderful contributions in the past but haven't been around for a few years?

Brad

meredydd:
I like the idea of a self-approval, which at least provides a place for the maintainer to say why they're self-approving. Perhaps we could set out criteria that permit one, then one could just cite the criteria in the self-approval?

Off the top of my head, things that push the balance towards self-approval might include:

* Tests (either very thorough coverage or, ideally, adding and passing the relevant CPython tests) 
* Very localised impact (eg "just this one library module")
* Relative improvement (eg if you're turning a NotImplemented stdlib module that clearly belongs in Skulpt into a working impl that passes CPython tests, you shouldn't need too much review!)

Does "any one of those being true" work for you?

bnmnetp:
Yes, any one of those 3 would qualify for self-approval in my mind.

Compilers and language implementation have never been my area of expertise, I took on the maintainer role for skulpt because I wanted it to live and and I wanted to use it in Runestone Academy.  So for me my criteria early on, up until we added this current set of rules was always pretty simple.  

1. Does it move us forward in some way?
2. It doesn't break stuff.

We have so many more unit tests now than we did before that I'm convinced that if it passes our current set of unit tests it is not going to break current core functionality.  I'm definitely in favor of a more permissive PR merge policy, and if we run into a situation that really causes trouble we can always revert the merge, add unit tests to detect the thing it broke, fix the PR and move on.

rixner:
I agree.  I think the approval process served us well in the past, but has now become an impediment to progress.

I would be in favor of requiring the self-approver to explain why they are self approving and then just do so.  But, maybe there should be a few day "waiting period" to allow anyone to review (if they have the time or inclination)?

I know that I haven't had the time to devote to the project that I once did, but I still look at things, even if I don't have time to do a careful enough review to approve...

PR #1207, opened 2020-10-09
make locals and globals work(ish)

s-cork:
This pr aims to fix globals and locals (ish)

---

**The current problem with globals:**

Modules can't use `globals` in skulpt 
from python docs `globals`:
> Return a dictionary representing the current global symbol table. This is always the dictionary of the current module (inside a function or method, this is the module where it is defined, not the module from which it is called).

but in skulpt we use whatever the current reference is to `Sk.globals`
this will depend on the module that is `__main__` and so `globals()` will only work if you're calling it from `__main__`

---

**How this pr fixes it:**
At each block I add a state object `Sk._state = {globals: $gbl, locals.$loc}`
and thus will be correct for any call to globals/locals

---

**Use of `mappingproxy`**
This is why I have (ish) in the title.
`globals()` in skulpt is really a readonly property since changing it has no effect.
but in python you can change the `globals()` object as you like. 

currently we return a dictionary which is a copy of `Sk.globals` 
but given the above - I think it would be more sensible to return a `mappingproxy` since it is readonly and will raise the appropriate error if someone tries to do something like

```python
globals()['foo'] = 'bar'
```
It's probably better that this raises an error in `skulpt` until changing `globals()` has an effect.

---

This pr depends on `mappingproxy` being implemented and so won't work on master. 

I don't know if this is the best way... if anyone else (@albertjan 😉 ) has tried to implement `locals` and has a more elegant solution then yes please!

I don't particularly like that I've ended up updating the `_state` object at every block. But I couldn't see a way around it...
since you need to be sure that the block hasn't done a call to a module who has then update the `_state` object.

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -1373,15 +1373,8 @@ Sk.builtin.issubclass = function issubclass (c1, c2) {
 };
 
 Sk.builtin.globals = function globals () {
-    var i, unmangled;
-    var ret = new Sk.builtin.dict([]);
-    for (i in Sk["globals"]) {
-        unmangled = Sk.unfixReserved(i);
-        ret.mp$ass_subscript(new Sk.builtin.str(unmangled), Sk["globals"][i]);
-    }
-
-    return ret;
-
+    const globals = (Sk._state && Sk._state.globals) || {};
+    return new Sk.builtin.mappingproxy(globals);
 };
 
 Sk.builtin.divmod = function divmod (a, b) {
@@ -1534,8 +1527,9 @@ Sk.builtin.iter = function iter (obj, sentinel) {
     }
 };
 
-Sk.builtin.locals = function locals () {
-    throw new Sk.builtin.NotImplementedError("locals is not yet implemented");
+Sk.builtin.locals = function locals() {
+    const locals = (Sk._state && Sk._state.locals) || {};
+    return new Sk.builtin.mappingproxy(locals);
 };
 Sk.builtin.memoryview = function memoryview () {
     throw new Sk.builtin.NotImplementedError("memoryview is not yet implemented");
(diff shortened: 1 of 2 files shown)


meredydd:
This is a fairly substantial change, I think - it moves all local variables from the local JS scope into the `$loc` object, and updates a global variable on every block(!). Quite apart from the performance costs (which might be necessary for correctness), I suspect this might have implications for name handling - for example, if we're really going to store all local variables in `$loc`, we should probably get them from there too.

s-cork:
yeap I don't think I like it either! There must be a better way.

Performance wise there seemed to be minimal impact on `fields.py` when I ran it.

s-cork:
What if we could do something like

enter scope
```javascript
var _prevframe = Sk._frame;
Sk._frame = {f_globals: $gbl, f_locals: $loc};
```
and then when we exit a scope 
```javascript
Sk._frame = _prevframe;
```

This way it's more aware and doesn't need an update at every block!

It'll have to be a little bit cleverer since leaving the scope could happen during:
- a suspension
- a return
- an error

It does still require a change to how `functions` do `$loc` since we need access to these somehow if `locals` is to work.

It means that the code for locals/globals would look like:
```javascript
Sk.builtin.locals = function () {
    const locals = Sk._frame.locals;
    return new Sk.builtin.mappingproxy(locals);
```


This is what cpython's `locals` implementation looks like:

```c
PyObject *
PyEval_GetLocals(void)
{
    PyThreadState *tstate = _PyThreadState_GET();
    PyFrameObject *current_frame = tstate->frame;
    return current_frame->f_locals;
}
```

s-cork:
i've updated this - removing the dependency for `mappingproxy` - with #1098  (use py strings as keys in `sys.modules`) this pr will work on master branch.

alexwenbj:
It seems this PR not included in your 1092/all_changes repo? @s-cork

PR #1209, opened 2020-10-10
metaclass proof of concept

s-cork:
This branch is a proof of concept for metaclasses

More needs to happen for metaclasses to by inline with python but you can do simple things like:

```python3
>>> class A(type): pass
>>> class B(metaclass=A): pass
>>> type(B)
__main__.A
```

And more complicated things like

```python3
>>> class A(type):
...   def __new__(cls, name, bases, _dict, **kwargs):
...     _dict.update(kwargs)
...     return super(A, cls).__new__(cls, name, bases, _dict)
...
>>> class B(metaclass=A, foo='bar'): pass
>>> B.__dict__
mappingproxy({'__module__': '__main__',
              'foo': 'bar',
              '__dict__': <getset_descriptor '__dict__' of 'B' objects>
            })
```

The first example does work on master 😄 

The second example works on #1092 but does not work on master because `type` has no `__new__` implementation and `super` is a little quirky on master branch.

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -317,6 +317,38 @@ Compiler.prototype.cunpackstarstoarray = function(elts, permitEndOnly) {
     }
 };
 
+Compiler.prototype.cunpackkwstoarray = function (keywords, codeobj) {
+        
+    let keywordArgs = "undefined";
+
+    if (keywords && keywords.length > 0) {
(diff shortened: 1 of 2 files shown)


s-cork:
super-seeded by #1244

PR #1212, opened 2020-10-12
make exec work

s-cork:
This is a cherry picked commit on top of #1092

A suggestion for exec. 

I have suggested two implementations:
internal implementations are called via `Sk.builtin.exec`
and python are called via `Sk.builtins.exec`

The reason for separate implementations is that...it could be quite nice to pass a javascript object for globals and locals.

```javascript
globals = {foo: new Sk.builtin.str('bar')}
Sk.builtin.exec("print(foo)", globals)
```

But i'm not wedded to an approach. 
I've been working on implementing dataclasses - as a first step to get the module to work using the version in python - exec was necessary. So here it is...

UPDATE:
- I've limited to globals, locals only being dictionaries when called from python, whereas, Cpython supports dicts for globals and any mapping for locals. This can be updated in the future.
- I adjusted how repl.js worked since this implementation of exec means that you can keep track of the globals without having to use `forceGlobal`.

### DIFF
--- a/src/builtindict.js
+++ b/src/builtindict.js
@@ -231,9 +231,9 @@ Sk.abstr.setUpModuleMethods("builtins", Sk.builtins, {
         $name: "eval",
         $meth: function (source, globals, locals) {
             // check globals
-            const tmp_globals = checkGlobLoc(globals, "globals");
+            const tmp_globals = globLocToJs(globals, "globals");
             // check locals
-            const tmp_locals = checkGlobLoc(locals, "locals");
+            const tmp_locals = globLocToJs(locals, "locals");
             return Sk.misceval.chain(Sk.builtin.eval(source, tmp_globals, tmp_locals), (res) => {
                 reassignGlobLoc(globals, tmp_globals);
                 reassignGlobLoc(locals, tmp_locals);
@@ -249,9 +249,9 @@ Sk.abstr.setUpModuleMethods("builtins", Sk.builtins, {
     exec: {
         $meth: function (source, globals, locals) {
             // check globals
-            const tmp_globals = checkGlobLoc(globals, "globals");
+            const tmp_globals = globLocToJs(globals, "globals");
             // check locals
-            const tmp_locals = checkGlobLoc(locals, "locals");
+            const tmp_locals = globLocToJs(locals, "locals");
             return Sk.misceval.chain(Sk.builtin.exec(source, tmp_globals, tmp_locals), (new_locals) => {
                 reassignGlobLoc(globals, tmp_globals);
                 reassignGlobLoc(locals, tmp_locals);
@@ -491,7 +491,7 @@ Sk.abstr.setUpModuleMethods("builtins", Sk.builtins, {
 });
 
 // function used for exec and eval
-function checkGlobLoc(glob_loc, name) {
+function globLocToJs(glob_loc, name) {
     let tmp = undefined;
     if (glob_loc === undefined || Sk.builtin.checkNone(glob_loc)) {
         glob_loc = undefined;
--- a/src/builtindict.js
+++ b/src/builtindict.js
@@ -198,12 +198,12 @@ Sk.abstr.setUpModuleMethods("builtins", Sk.builtins, {
         $doc: "Return a Unicode string of one character with ordinal i; 0 <= i <= 0x10ffff.",
     },
 
-    // compile: {
-    //     $meth: Sk.builtin.compile,
-    //     $flags: {},
-    //     $textsig: "($module, /, source, filename, mode, flags=0,\n        dont_inherit=False, optimize=-1)",
-    //     $doc: "Compile source into a code object that can be executed by exec() or eval().\n\nThe source code may represent a Python module, statement or expression.\nThe filename will be used for run-time error messages.\nThe mode must be 'exec' to compile a module, 'single' to compile a\nsingle (interactive) statement, or 'eval' to compile an expression.\nThe flags argument, if present, controls which future statements influence\nthe compilation of the code.\nThe dont_inherit argument, if true, stops the compilation inheriting\nthe effects of any future statements in effect in the code calling\ncompile; if absent or false these statements do influence the compilation,\nin addition to any features explicitly specified."
-    // },
+    compile: {
+        $meth: Sk.builtin.compile,
+        $flags: {MinArgs: 3, MaxArgs:6},
+        $textsig: "($module, /, source, filename, mode, flags=0,\n        dont_inherit=False, optimize=-1)",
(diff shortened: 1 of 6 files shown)


alexwenbj:
It seems eval() still not implemented? Is it possible to implement it base on this exec()?🙂

s-cork:
good question - the problem with `eval` for skulpt is that it has to be an expression. 
And our `Sk.compile`/`ast` only knows how to do `exec`/`modules`

in `ast` you can see the `@todo` for `astFromParse` for other flag types

Something like this might work but I guess it's only really a temporary solution
```javascript
Sk.builtin.eval = function(source, globals, locals) {
    const parse = Sk.parse("?", source);
    const ast = Sk.astFromParse(parse.cst, "?", parse.flags);
        if (ast.body.length > 1 || !(ast.body[0] instanceof Sk.astnodes.Expr)) {
            throw 
        } else {
             loc = Sk.builtin.exec("__final_res__ = " + source, globals, locals)
             return loc.__final_res__;
        }
}
```

alexwenbj:
I have almost never used them, but as the doc describes, from my opinion, the difference between eval() and exec() mainly focus two aspects:
1. the arguments. exec()'s first arg may be an object, and eval() only accepts string as the first arg;
2. exec() has no returned value, eval() returns sth.

Ref:
https://docs.python.org/3/library/functions.html#exec
https://docs.python.org/3/library/functions.html#eval

s-cork:
I've adjusted with an implementation of eval based on the implementation of exec. 


It should work for most things and for simple tests from cpython it works 
(we don't support anything other than dicts for `globals/locals` and you can't override `__builtins__`)

One key difference between the two is - you can do:
```
exec('class A: pass`)
```
but can't do
```
eval('class A: pass') # SyntaxError: invalid syntax
```
Because the latter is not an expression.

alexwenbj:
It is the same as in CPython: eval('class A:pass') will raise a SyntaxError.

s-cork:
😉  yes that's what I meant - the code in this pr tries to mimic that behaviour.

alexwenbj:
Cool.

alexwenbj:
Which PR compatible with this PR? I replaced 3 files(builtin.js/builtindict.js/compile.js) in https://github.com/s-cork/skulpt/tree/pr/re_update, but there are 2 errors in console...

s-cork:
here's a branch that builds on top of re_update with `__annotations__` and this `exec` pr
https://github.com/s-cork/skulpt/tree/1092/all_changes

I'll keep that up to date with the current prs I'm making.
let me know if you have problems and will fix.

alexwenbj:
Got it ,thank you.@s-cork

alexwenbj:
`exec('class A: pass')` raised an error: `ReferenceError: $free is not defined`

s-cork:
@alexwenbj 
I can't seem to reproduce - is there more context?
When I try I get
```python3
>>> exec('class A: pass')
>>> A
<class '__main__.A'>
>>> A()
<__main__.A object>
```

alexwenbj:
I ran the code in my editor,the complete error message is:
```
constructor {nativeError: ReferenceError: $free is not defined
    at $A$class_outer (eval at Sk.misceval.chain.Sk.globals (h…, args: constructor, traceback: Array(2), $d: constructor}
$d: constructor
$version: 0
buckets: {}
entries: {}
in$repr: false
size: 0
__proto__: constructor
args: constructor
in$repr: false
v: [constructor]
__proto__: constructor
nativeError: ReferenceError: $free is not defined at $A$class_outer (eval at Sk.misceval.chain.Sk.globals (http://localhost/python/static/js/skulpt.min.js:360:434), <anonymous>:10:567) at Object.Sk.misceval.buildClass (http://localhost/python/static/js/skulpt.min.js:418:48) at $scope6 (eval at Sk.misceval.chain.Sk.globals (http://localhost/python/static/js/skulpt.min.js:360:434), <anonymous>:10:64) at Arguments.Sk.misceval.chain.Sk.globals (http://localhost/python/static/js/skulpt.min.js:360:446) at Object.Sk.misceval.chain (http://localhost/python/static/js/skulpt.min.js:412:335) at Object.Sk.builtin.exec (http://localhost/python/static/js/skulpt.min.js:360:423) at $meth (http://localhost/python/static/js/skulpt.min.js:1031:47) at constructor.$callMinArgs (http://localhost/python/static/js/skulpt.min.js:261:169) at $scope0 (eval at <anonymous> (http://localhost/python/static/js/skulpt.min.js:994:97), <anonymous>:11:207) at Arguments.<anonymous> (http://localhost/python/static/js/skulpt.min.js:994:369)
message: "$free is not defined"
stack: "ReferenceError: $free is not defined↵    at $A$class_outer (eval at Sk.misceval.chain.Sk.globals (http://localhost/python/static/js/skulpt.min.js:360:434), <anonymous>:10:567)↵    at Object.Sk.misceval.buildClass (http://localhost/python/static/js/skulpt.min.js:418:48)↵    at $scope6 (eval at Sk.misceval.chain.Sk.globals (http://localhost/python/static/js/skulpt.min.js:360:434), <anonymous>:10:64)↵    at Arguments.Sk.misceval.chain.Sk.globals (http://localhost/python/static/js/skulpt.min.js:360:446)↵    at Object.Sk.misceval.chain (http://localhost/python/static/js/skulpt.min.js:412:335)↵    at Object.Sk.builtin.exec (http://localhost/python/static/js/skulpt.min.js:360:423)↵    at $meth (http://localhost/python/static/js/skulpt.min.js:1031:47)↵    at constructor.$callMinArgs (http://localhost/python/static/js/skulpt.min.js:261:169)↵    at $scope0 (eval at <anonymous> (http://localhost/python/static/js/skulpt.min.js:994:97), <anonymous>:11:207)↵    at Arguments.<anonymous> (http://localhost/python/static/js/skulpt.min.js:994:369)"
__proto__: Error
traceback: Array(2)
0: {lineno: 1, colno: 0, filename: "?"}
1: {lineno: 1, colno: 0, filename: "<stdin>.py"}
length: 2
__proto__: Array(0)
__proto__: constructor
```

s-cork:
The eval error is because it does a JavaScript eval. I don't know. I'll drop you an email and see if we can work it out - it might take some back and forth.

alexwenbj:
It seems this PR works good(although in my env there is a small problem),would you please review and merge it?

src/builtindict.js
```diff
@@ -229,19 +229,40 @@ Sk.abstr.setUpModuleMethods("builtins", Sk.builtins, {
...
-    // exec: {
-    //     $meth: Sk.builtin.exec,
-    //     $flags: {MinArgs:2, MaxArgs: 3},
-    //     $textsig: "($module, source, globals=None, locals=None, /)",
-    //     $doc: "Execute the given source in the context of globals and locals.\n\nThe source may be a string representing one or more Python statements\nor a code object as returned by compile().\nThe globals must be a dictionary and locals can be any mapping,\ndefaulting to the current globals and locals.\nIf only globals is given, locals defaults to it."
-    // },
+    exec: {
+        $meth: function (source, globals, locals) {
+            // check globals
+            const tmp_globals = checkGlobLoc(globals, "globals");
```
meredydd, line 252: `checkGlobLoc` is a slightly weird name for this function, which actually converts from py dict to JS obj/scope

meredydd (approved):
One suggestion for improved function name, otherwise GTG

PR #1213, opened 2020-10-14
make assertion conditions only run if enable asserts is switched on

s-cork:
One current issue with assertions in skulpt is that, while they don't throw assertion errors the condition is still evaluated
Some conditions are slower than others and so this might then cause a performance issue on hot code.

There are a couple of solutions for this that I can see. 
1. we could be cleverer about webpack and remove the condition completely in production mode so that these conditions are never evaluated. 
2. change the assertion first argument to an arrow function and only check this condition in dev mode

1 would have the biggest win performance wise.

This pr does option 2 and replaces all `Sk.asserts.assert` condition statement with an arrow function
the advantage is that the assert condition is only evaluated in dev mode and quicker `(if enable_asserts)` condition is evaluated instead.

So something like
```javascript
Sk.asserts.assert(this instanceof Sk.builtin.str)
```
becomes
```javascript
Sk.asserts.assert(() => this instanceof Sk.builtin.str)
```

It also means that those using skulpt can turn on and off assertion statements as they like just using the `min.js` file

They can just do `Sk.asserts.ENABLE_ASSERTS = true` rather than worry about different build options.
It could potentially become one of the many undocumented `configure` arguments.


meredydd:
This looks sensible to me, and I'm guessing it will barely slow us down (unless V8 is already being _really_ clever about cross-function dead code elimination).

albertjan:
If we wrap the inards of `assert` with:

```
if (process.env.NODE_ENV === 'development') {
}
```

webpack will automatically remove the code from the production build.

PR #1214, opened 2020-10-15
support module and class __annotations__

### LINKED ISSUE #1210: annotations without an equals are not compiled
```python 3
>>> class A:
...    x: int
...
Cannot read property 'lineno' of null
>>> y: int
Cannot read property 'lineno' of null
```

It would also be great to support `__annotations__` since we pass them correctly (except for the above examples)

I think if the above works then `dataclasses` might even be possible... nearly.

s-cork:
current skulpt fails to compile
```python3
class A:
  x: int
```

it also doesn't do `__annotations__` for modules or classes. 
This pr aims to fix both of those. 

pr on the road to `dataclasses` 
closes #1210


s-cork:
I have two compile options above - I haven't played with the compiler much so I wasn't sure exactly... 

The first commit is very much cpython inspired from `compile.c`
https://github.com/python/cpython/blob/master/Python/compile.c#L1562-L1614
and
https://github.com/python/cpython/blob/master/Python/compile.c#L5239-L5265

The second commit seems like a good option here and saves walking over the `cbody` `stmts` twice.

albertjan (approved):

bnmnetp (approved):

PR #1216, opened 2020-10-18
support generic alias

### LINKED ISSUE #1146: __class_getitem__ support
I don't think there's been any work to handle `__class_getitem__`, which is new in Python 3.7. I wanted to raise it as an issue, since I'm thinking about writing some code that might rely on it and therefore might make a cursory implementation. Unless @s-cork already has it finished it somewhere... :)

s-cork:
this pr adds support for `__class_getitem__` (closes #1146 )

it implements partial support for pep 585: https://www.python.org/dev/peps/pep-0585/

```python3
list[int, ...]
dict[str, int]
type[int]

from itertools import chain
chain[list]
```

it provides code for `__mro_entries__` but since this is not checked in skulpt `class A(list[int]): pass` does not work
similarly `__instancecheck__` and `__subclasscheck__` are not yet called.

some of the cpython code accounts for `typing.TypeVar` but since that is not yet a thing much of the implementation doesn't yet make sense in skulpt. 

This won't work on master and builds on top #1092

All the tests in this pr are working on another branch. 
Tests taken from Cpython.

### DIFF
--- a/src/generic_alias.js
+++ b/src/generic_alias.js
@@ -1,0 +1,208 @@
+Sk.builtin.GenericAlias = Sk.abstr.buildNativeClass("types.GenericAlias", {
+    constructor: function GenericAlias(origin, args) {
+        this.$origin = origin;
+        if (!(args instanceof Sk.builtin.tuple)) {
+            args = new Sk.builtin.tuple([args]);
+        }
+        this.$args = args;
+        this.$params = null;
+    },
+    slots: {
+        tp$new(args, kwargs) {
+            Sk.abstr.checkNoKwargs("GenericAlias", kwargs);
+            Sk.abstr.checkArgsLen("GenericAlias", args, 2, 2);
+            return new Sk.builtin.GenericAlias(args[0], args[1]);
+        },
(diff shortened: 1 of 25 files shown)


src/generic_alias.js
```diff
@@ -0,0 +1,209 @@
...
+            return tup_arr.indexOf(item);
+        },
+        is$typevar(type) {
+            if (type.tp$name !== "TypeVar") {
+                return false;
+            }
+            const module = Sk.abstr.lookupSpecial(type, Sk.builtin.str.$module);
+            if (module === undefined) {
+                // throw some sort of error
+                throw "err?";
```
meredydd, line 170: Um? At least make this a Python error object.

src/generic_alias.js
```diff
@@ -0,0 +1,209 @@
...
+            }
+            if (module.toString() === "typing") {
+                return true;
```
meredydd, line 173: This makes baby Jesus cry.

meredydd (approved):
Looks good to me. Useful ground work for typing, which will be big!

PR #1217, opened 2020-10-19
fix a problem with compiled code after waking from suspension

s-cork:
This fixes a potential problem where the `this` might be undefined if a function wakes from a suspension.

Mostly `this` is `window` when a function wakes form a suspension but I have found this bug in the wild when it becomes `undefined` or `null`
I can't seem to reproduce it with a test in skulpt.


meredydd (approved):
Simple fix, approved.

PR #1218, opened 2020-10-19
patch for <> operator

s-cork:
an alt version of (closes #952) which adds the operator in py2 mode. 

I've tried to avoid polluting the `Sk` name space and `Sk.token` seemed a sensible name space for this patch

It also fixes a bug where all the `EXACT_TOKEN_TYPES` were incorrect since they needed to be `T_NOTEQUAL` and not `NOTEQUAL`

NB: `Funny` is not my name - it's cpython's name
https://github.com/python/cpython/blob/master/Lib/tokenize.py#L122

and it changes `Funny` from 

```python3
"(\r?\n|(~|\}|\|=|\||\{|\^=|\^|\]|\[|@=|@|>>=|>>|>=|>|==|=|<=|<<=|<<|<|;|:|/=|//=|//|/|\.\.\.|\.|->|-=|-|,|\+=|\+|\*=|\*\*=|\*\*|\*|\)|\(|&=|&|%=|%|!=))"
```
to
```python3
                                                         # <> here i am
"(\r?\n|(~|\}|\|=|\||\{|\^=|\^|\]|\[|@=|@|>>=|>>|>=|>|==|=|<>|<=|<<=|<<|<|;|:|/=|//=|//|/|\.\.\.|\.|->|-=|-|,|\+=|\+|\*=|\*\*=|\*\*|\*|\)|\(|&=|&|%=|%|!=))"
```

i.e. just inserts `<>` into the appropriate place.


src/tokenize.js
```diff
@@ -169,9 +169,32 @@ var String_ = group(StringPrefix + "'[^\\n'\\\\]*(?:\\\\.[^\\n'\\\\]*)*'",
...
+        delete Sk.token.EXACT_TOKEN_TYPES["<>"];
+    } else {
+        Sk.token.EXACT_TOKEN_TYPES["<>"] = Sk.token.tokens.T_NOTEQUAL;
+    }
+    EXACT_TOKENS_SORTED = Object.keys(Sk.token.EXACT_TOKEN_TYPES).sort();
+    Special = group.apply(
+        this,
+        EXACT_TOKENS_SORTED.reverse().map(function (t) {
+            return regexEscape(t);
+        })
```
meredydd, line 189: `EXACT_TOKENS_SORTED.reverse().map(regexEscape)`?

meredydd (approved):
This is *way* better than #952. Thank-you!

src/tokenize.js
s-cork: not sure - not my code - this was copied form the lines above. I guess it's something to do with making sure that the tokens are js regex-able for later

s-cork: Yeah so I think `.` becomes `\.`

PR #1221, opened 2020-10-26
feat(builtin function): raw_input shouldn't be available in python3

bofeng:
If user preferred to use Python3 in Skulpt, then the `raw_input` function should raise NameError exception.


s-cork:
I think a better approach might be to go into `builtindict.js` and just remove `raw_input` in the function `Sk.setupObjects`
there are other examples in that file you can see.

bofeng:
@s-cork oh, that's definitely a better method, I'll close this one and create a new pull request. Thank you!

PR #1222, opened 2020-10-27
feat(builtin functions): remove raw_input function in python3

bofeng:
If Python3 mode is preferred in Skulpt, then the raw_input function shouldn't be available.

Credits: [comment](https://github.com/skulpt/skulpt/pull/1221#issuecomment-716921934) from @s-cork

### DIFF
--- a/src/builtindict.js
+++ b/src/builtindict.js
@@ -137,6 +137,7 @@ Sk.setupObjects = function (py3) {
         Sk.builtins["StandardError"] = Sk.builtin.Exception;
         Sk.builtins["unicode"] = Sk.builtin.str;
         Sk.builtins["basestring"] = Sk.builtin.str;
+        Sk.builtins["raw_input"] = new Sk.builtin.func(Sk.builtin.raw_input);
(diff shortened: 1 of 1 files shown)


src/builtindict.js
```diff
@@ -124,6 +124,7 @@ Sk.setupObjects = function (py3) {
         delete Sk.builtins["StandardError"];
         delete Sk.builtins["unicode"];
         delete Sk.builtins["basestring"];
+        delete Sk.builtins["raw_input"];
```
s-cork, line 127: You'll also need to add it back on to builtins for py2 mode.

Some users of skulpt may switch between the py2/py3 setups depending on their user's configuration.

bofeng: Got it, I added this fix in the new commit: https://github.com/skulpt/skulpt/pull/1222/commits/7e207e4cbb286e327b98537db9dab855553cad34
Thank you very much for pointing it out.

s-cork (approved):

bnmnetp (approved):

PR #1225, opened 2020-11-05
pep 448 support {**dict_1, **dict_2}

### LINKED ISSUE #1220: ** within a dict comprehension is not supported
```python3
>>> {**{1:2}}
Cannot read property 'lineno' of null
```

see https://www.python.org/dev/peps/pep-0448/

s-cork:
(built on top of #1092 - relies on a fast suspension aware call to a private method `dict$merge`)
closes #1220 

Compile code adjusted from cpython 
https://github.com/python/cpython/blob/master/Python/compile.c#L3833-L3921

Tests adjusted from Cpython
https://github.com/python/cpython/blob/master/Lib/test/test_unpack_ex.py#L94-L133

(won't pass this master without the `dict$merge` function - can add this here I guess)


meredydd:
Pretty rad! (And actually a pretty simple implementation, too - from the 1s and 0s I'm guessing it -- and the tests -- are lifted directly from CPython.)

meredydd (approved):
Nice and neat, and passes the tests!

Issue #1231, opened 2020-11-22
Build failure in wrapmodules/minify stage


bennorth:
I can't pin down exactly when this started happening, but just now a fresh clone fails to build for me:
```shell
% git clone https://github.com/skulpt/skulpt.git skulpt-build-test
[... normal output ...]
% cd skulpt-build-test
% npm install
[... no obvious errors ...]
% npm run build
[...]
> node support/build/wrapmodules.js builtin

/tmp/skulpt-build-test/node_modules/@babel/core/lib/transformation/index.js:45
    throw e;
    ^

TypeError: unknown: Cannot read property 'bindings' of null
    at PluginPass.enter (/tmp/skulpt-build-test/node_modules/babel-plugin-minify-simplify/lib/index.js:331:44)
    at newFn (/tmp/skulpt-build-test/node_modules/@babel/traverse/lib/visitors.js:175:21)
    at NodePath._call (/tmp/skulpt-build-test/node_modules/@babel/traverse/lib/path/context.js:55:20)
    at NodePath.call (/tmp/skulpt-build-test/node_modules/@babel/traverse/lib/path/context.js:42:17)
    at NodePath.visit (/tmp/skulpt-build-test/node_modules/@babel/traverse/lib/path/context.js:92:31)
    at TraversalContext.visitQueue (/tmp/skulpt-build-test/node_modules/@babel/traverse/lib/context.js:115:16)
    at TraversalContext.visitQueue (/tmp/skulpt-build-test/node_modules/@babel/traverse/lib/context.js:121:21)
    at TraversalContext.visitMultiple (/tmp/skulpt-build-test/node_modules/@babel/traverse/lib/context.js:79:17)
    at TraversalContext.visit (/tmp/skulpt-build-test/node_modules/@babel/traverse/lib/context.js:141:19)
    at Function.traverse.node (/tmp/skulpt-build-test/node_modules/@babel/traverse/lib/index.js:82:17) {
  code: 'BABEL_TRANSFORM_ERROR'
}
```
There was [an update to `@babel/core` at 2020-11-20T21:06:00.584Z](https://www.npmjs.com/package/@babel/core/v/7.12.7) which might be relevant.

I have reproduced the above failure on an Amazon Lightsail instance (Ubuntu 20.04LTS image) with fresh node/npm installation, so I don't think it's a quirk of my setup here.

[...]

PR #1232, opened 2020-11-22
Use terser rather than babel-minify

bennorth:
As discussed in #1231, a fresh checkout is encountering build failures, perhaps because of a change to a dependency outside Skulpt itself.  The failure is in the minify stage.  One option is to move from [babel-minify](https://www.npmjs.com/package/babel-minify) (15k weekly downloads) to a more active minify tool such as [terser](https://www.npmjs.com/package/terser) (17M weekly downloads).  This PR suggests how that move might be done.

I've reviewed [the options terser offers](https://github.com/terser/terser#api-reference) for controlling minification, and tried to pick conservative ones.  Skulpt's unit tests pass with the new minified code (except `TimeTestCase.test_strftime`, which I think is unrelated), as do the unit tests of a project of mine which is based on Skulpt.  All the same I realise this is a deep change so would welcome input from the maintainers!


bennorth:
Using `terser` rather than `babel-minify` cuts down the number of dependencies a bit too.  Running
```shell
rm -r node_modules && npm install && npm list | wc -l
```
on current `master` gives 1209 entries; with this PR, that number drops to 1068.

It would also eliminate one (low-priority) `npm audit` result, for `yargs-parser`.

s-cork:
Thanks for the updates - one thing that would be great (that we don't currently do (afaik)) would be to transpile the stdlib based on the preferred build method.

more specifically - es6 syntax is not transpiled to es5 syntax in the `stdlib` - which is becoming less problematic with the demise of IE but would still be a gotcha for users that expect a full es5 compatible build when someone runs `npm run build-es5`.

Is a mechanism for build options within the scope of this pr?

bennorth:
(Force-pushed to rebase onto new `master`.)

rixner:
In principle, this seems like a good idea. My question, though, is does this make it easier, harder, or the same to ultimately achieve the objective of transpiling the libraries to match the desired language level of the core? That doesn't need to be part of this pull request, but I think we should be thinking ahead (and it's well past time for adding that capability...)

albertjan:
I think this is a good idea, because we're adopting a more widely used tool, which will obviously be better maintained. @s-cork what do you think?

s-cork:
I'd consider this an improvement on what we currently have - it means relying on a more maintained minifier. @bennorth the error from the unittest is a package.json vs package-lock.json inconsistency. If you could update the lockfile so that the terser version matches then we can this merged.

I agree with @rixner that there's a broader discussion here about what the best build process is for the stdlib/core. But outside the scope of this pr.

bennorth:
Thanks for looking at this @s-cork.  I can rebase onto fresh `master` and force-push, to make for a less tangled history?  There's been a new release of `terser` (v5.7.0) which I was thinking to update to at the same time.

bennorth:
Force-pushed after rebasing onto fresh `master`.

bennorth:
Force-pushed after rebasing onto fresh `master`.

s-cork (approved):
looks good to me

albertjan (approved):

PR #1233, opened 2020-11-22
Refactor time module and fix its tests

bennorth:
Follow-up to the aside in #1231 regarding the failed `TimeTestCase.test_strftime` test.

If I'm reading it right, Skulpt's implementation of `time.timezone` uses the offset as of 2002.  I'm in Ireland and [the history of its civil timekeeping](https://en.wikipedia.org/wiki/Time_in_the_Republic_of_Ireland#History) means that its timezone offset was different in 1970 compared to 2002.  It's likely that developers in other timezones weren't seeing this failure.  This PR updates the test to use a moment in 2002, for consistency with how `timezone` is computed.  I've checked it under timezones `Europe/Dublin`, `America/Denver`, `Europe/Istanbul`.

This PR also extracts a function from near-duplicated code between Skulpt's `localtime()` and `gmtime()` implementations, and fixes a minor typo.

### DIFF
--- a/src/lib/time.js
+++ b/src/lib/time.js
@@ -164,6 +164,16 @@ var $builtinmodule = function (name) {
         );
     }
 
+    function from_seconds(secs, asUtc) {
+        var d = new Date();
+        if (secs) {
+            Sk.builtin.pyCheckType("secs", "number", Sk.builtin.checkNumber(secs));
+            var seconds = Sk.builtin.asnum$(secs);
+            d.setTime(seconds * 1000);
+        }
+        return date_to_struct_time(d, asUtc);
+    }
+
     function localtime_f(secs) {
         Sk.builtin.pyCheckArgsLen("localtime", arguments.length, 0, 1);
         var d = new Date();
--- a/src/lib/time.js
+++ b/src/lib/time.js
@@ -190,7 +190,7 @@ var $builtinmodule = function (name) {
     function asctime_f(time) {
         if (!time || Sk.builtin.checkNone(time))
         {
-            time = localtime_f();
+            time = from_seconds();
         } else if (!(time instanceof struct_time_f)) {
             time = new struct_time_f(time);
         }
@@ -215,7 +215,7 @@ var $builtinmodule = function (name) {
     mod.asctime = new Sk.builtin.func(asctime_f);
 
     mod.ctime = new Sk.builtin.func(function(secs) {
-        return asctime_f(localtime_f(secs));
+        return asctime_f(from_seconds(secs));
     });
 
     function mktime_f(time) {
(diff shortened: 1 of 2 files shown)


bennorth:
(The build is failing because of #1231 (see possible fix in #1232) but I thought it cleanest to make this a separate PR.)

bennorth:
Further info re Travis build failure:  A local branch merging this PR and #1232 builds and passes all tests:
```shell
% git checkout -b test-1232-and-1233 use-terser
% git merge --no-edit --no-ff refactor-time-module-and-fix-its-tests
% rm -r node_modules
% npm install && npm run build && npm run test
[...]
test/unit/[...]
Passed: 462 Failed: 0
test/unit3/[...]
Passed: 938 Failed: 0
```

src/lib/time.js
```diff
@@ -219,7 +215,7 @@ var $builtinmodule = function (name) {
     mod.asctime = new Sk.builtin.func(asctime_f);
 
     mod.ctime = new Sk.builtin.func(function(secs) {
-        return asctime_f(localtime_f(secs));
+        return asctime_f(from_seconds(secs));
```
s-cork, line 218: I think you've lost the arg lengths check in this function while doing the refactor

s-cork (approved):
LGTM - works for my timezone in KL, Malaysia which has always failed.
One comment to address

src/lib/time.js
bennorth: Maybe I'm misunderstanding, but the original call, `localtime_f(secs)`, passes exactly one argument to `localtime_f()` and so won't the arg lengths check in the original `localtime_f()` always pass, no matter what was passed to `ctime()`?  Testing against `master`, i.e., *without* this PR, seems to confirm this:
```python
print(time.ctime())  # -> "Thu Nov 26 14:46:35 2020" as expected
print(time.ctime(1606401981))  # -> "Thu Nov 26 14:46:21 2020" as expected
print(time.ctime(1606401981, 1, 1, 1))  # -> "Thu Nov 26 14:46:21 2020" ?? should be error
```
I can add args length checks to `ctime()` and `asctime()` if that would be useful?

s-cork: ah yes - so it was a bug already - can you address this as part of the refactor
in cyptyon
```python
>>> time.ctime(1606401981, 1, 1, 1)
TypeError: ctime() takes at most 1 argument (4 given)
```

bennorth: I've rebased onto `master`, and added a couple of commits which test that the appropriate exception is raised when too many arguments (or not enough) are passed.  Some of these tests initially fail (`asctime`, `ctime`, `mktime`); these are fixed by the newly-last commit in this PR.

s-cork:
This pr has caused a bug - `localtime_f` is referenced on line 288 but was removed in this pr.

bennorth:
Apologies.  I'll have a look at that now.

bennorth:
I've just created #1251.  Sorry for overlooking this.

PR #1235, opened 2020-11-24
Update turtle.js

### LINKED ISSUE #1234: turtle.Screen().setworldcoordinates() Broken
I'm facing a bug in Skulpt. Here is a minimum code necessary to reproduce the bug:

```py
import turtle
turtle.Screen().setworldcoordinates(8, 6, 96, 96)
turtle.goto(90,  7)
turtle.dot()
turtle.goto(30,  7)
turtle.dot()
turtle.goto(30,  87)
turtle.dot()
turtle.goto(90,  87)
```
Multiple versions of this code work in my local IDE / iPython environment, as well as on http://www.repl.it/

In that case, during the drawing phase (before the second `.setworldcoordinates()` ) I get this, incorrect, drawing:

image

I would expect to get this, correct, drawing -- which was generated with the exact same code copy/pasted to another interpreter environment ( http://www.repl.it/ and also reproduced on my own personal computer)

image

It's pretty clear that only the bottom 2 points of the "square" appear in Skulpt.

However, this is all before the second `.setworldcoordinates()` runs. After the second  `.setworldcoordinates()` function call, Skulpt no longer renders the turtle view at all, and the page renders like this:

image

It just shows a bold black line where the green square should be.

Additionally, there is an invisible, but selectable, element in the shape of the green square but no amount of my own HTML/CSS/JS edits were able to make it visible.

image

I believe my use of `.setworldcoordinates()` is correct, although potentially I am using it wrong. My program produces expected output in other environments.

bnmnetp:
This simplifies the calculation where lay == 0 and fixes the calculation where lly > 0

close #1234

### DIFF
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -1676,12 +1676,7 @@ function generateTurtleModule(_target) {
         context.restore();
         context.save();
         context.scale(1 / xScale, 1 / yScale);
-        // test
-        if (lly <= 0) {
-            context.translate(-llx, -ury);
-        } else {
-            context.translate(-llx, -(ury - lly) - lly);
-        } 
+        context.translate(-llx, -ury);
     }
 
     function pushUndo(turtle) {
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -1676,13 +1676,11 @@ function generateTurtleModule(_target) {
         context.restore();
         context.save();
         context.scale(1 / xScale, 1 / yScale);
-        if (lly === 0) {
+        if (lly <= 0) {
             context.translate(-llx, -ury);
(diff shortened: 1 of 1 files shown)


bnmnetp:
@skulpt/committers I migrated us from travis-ci.org to travis-ci.com -- I was reminded that .org will go away in December and it was taking forever for things to queue and run.

The build and check for this did complete but for some reason the status is not reflected here yet.  If anyone has ideas on what I might have missed I'm happy to fix it up.

Brad

s-cork (approved):
wow - surprised the fix involved removing lines of code! 🏆

bnmnetp:
Yeah, sometimes its all too easy to overthink the math.

PR #1237, opened 2020-11-30
Changed group for letters in identifier regex

tarjeiba:
Non-ascii letters, e.g. 'æ', was not handled correctly by the '[A-Z]' and '[a-z]' groups. The new
groups are taken from https://stackoverflow.com/a/26135188.


tarjeiba:
This fails if the identifier is more than one letter long. `æ = 3` works fine, `æd = 5` does not.

bnmnetp (changes requested):
Please add a unit test that demonstrates failure before your update and success after.  Also tests for the other cases identified in the comments would be appropriate.

src/tokenize.js
```diff
@@ -253,7 +253,7 @@ function _tokenize(filename, readline, encoding, yield_) {
     var Number_ = group(Imagnumber, Floatnumber, Intnumber);
     var PseudoToken = Whitespace + group(PseudoExtras, Number_, Funny, ContStr, Name);
 
-    const PseudoTokenRegexp = new RegExp(PseudoToken);
+  const PseudoTokenRegexp = new RegExp(PseudoToken, 'u');
```
s-cork, line 256: We shouldn't use the 'u' flag here since we still support older browsers.
To make this work correctly for the failed example you should adjust the `Name` category from `\w+` to something similar to the link you provided to stackoverflow in the description.

tarjeiba: Thank you! I'll have a go at it.

s-cork:
how are you getting on with this pr?

tarjeiba:
@

> @tarjeiba, how are you getting on with this pr?

Apologies for being late with this, I have it scheduled for Monday.

s-cork:
@tarjeiba  how are you getting on with this branch? I've taken the approach here and have it working in another branch. Happy to pr that if you've stopped working on this branch - but equally happy to wait on this pr if you're still working on it.

tarjeiba:
> @tarjeiba how are you getting on with this branch? I've taken the approach here and have it working in another branch. Happy to pr that if you've stopped working on this branch - but equally happy to wait on this pr if you're still working on it.

@s-cork, I've stalled completely, unfortunately. You're more than welcome to PR your work. Thanks for asking! I'm looking forward to seeing how you've done it.

PR #1241, opened 2020-12-21
Implement yield from

s-cork:
In the move to python3 `yield from` is a feature that we're obviously missing.

This pr implements that feature.

Since `StopIteration` sends values up the chain of subiterators and Skulpt looses the value when `tp$iternext` is called, I put the value property from the `StopIteration` on the iterator itself.

This pr is 1 commit ahead of #1092. You can see the diff in the last commit: a4cbf3c

I implement all the tests from Cpython
I make no attempt to implement `throw` or `close` methods for generators and leave these features as outside the scope of this pr.
Tests that required `throw` and `close` were commented out.

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -1150,7 +1150,14 @@ Sk.builtin.next_ = function next_(iter, default_) {
             if (default_) {
                 return default_;
             }
-            const v = iter.$value;
+            const v = iter.gi$ret;
+            // gi$ret is the original value that was thrown by the StopIteration
+            // (or returned by a generator)
+            // We've now lost the original StopIteration
+            // We make the assumption that it was probably called like
+            // raise StopIteration # or  StopIteration()
+            // rather than
+            // raise StopIteration(None)
             if (v !== undefined && v !== Sk.builtin.none.none$) {
                 throw new Sk.builtin.StopIteration(v);
             } else {
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -1150,7 +1150,12 @@ Sk.builtin.next_ = function next_(iter, default_) {
             if (default_) {
                 return default_;
             }
-            throw new Sk.builtin.StopIteration();
+            const v = iter.$value;
+            if (v !== undefined && v !== Sk.builtin.none.none$) {
+                throw new Sk.builtin.StopIteration(v);
+            } else {
+                throw new Sk.builtin.StopIteration();
+            }
         }
         return nxt;
     });
--- a/src/compile.js
+++ b/src/compile.js
@@ -560,6 +560,7 @@ Compiler.prototype.cyieldfrom = function (e) {
         throw new Sk.builtin.SyntaxError("'yield' outside function", this.filename, e.lineno);
     }
     var iterable = this.vexpr(e.value);
+    // get the iterator we are yielding from and store it
     iterable = this._gr("iter", "Sk.abstr.iter(", iterable, ")");
     out("$gen." + iterable + "=", iterable, ";");
     var afterIter = this.newBlock("after iter");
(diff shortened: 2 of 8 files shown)


bnmnetp:
@skulpt/committers now that #1092 has been merged it would be great if someone with better knowledge of the internals than me would approve this! 😄

src/compile.js
```diff
@@ -511,6 +511,37 @@ Compiler.prototype.cyield = function(e) {
     return "$gen.gi$sentvalue"; // will either be none if none sent, or the value from gen.send(value)
 };
 
+Compiler.prototype.cyieldfrom = function (e) {
```
meredydd, line 514: Overall: This function is huge and complex and has no comments! Something explaining what we're doing here and how would have helped (as it was I got @s-cork on a call to walk me through it, and I am now convinced, but not everyone can do that ;) ).

Something like (to my understanding):
```javascript
// First, get the iterator.
// [afterIter block starts]
// Check if we've been sent a value, because if so we need to send it on into the iterator by calling its `send()` method.
// If that `send()` method raises `StopIteration`, the sent value needs to be redirected to us (as it hasn't been used yet).
// As a performance optimisation, if we're sending to a generator, we can pass it into `tp$iternext()` and it will do this all for us (see generator.js).
// Either way, if the iterator is done (undefined) and we still have an unused sent value, it will be in `[iterable].$value`, so we grab it from there and move on from the `yield from` ("afterBlock")
// Otherwise, we return the next value from the iterator, and set the generator to do this all again to get the next element.
```

meredydd (approved):
Approve, with two changes agreed with @s-cork:

1. Rename `iter.$value` to `iter.gen$ret` to make it clear what it's holding
2. Comment `cyieldfrom`

src/builtin.js
```diff
@@ -1061,7 +1061,12 @@ Sk.builtin.next_ = function next_(iter, default_) {
         if (default_) {
             return default_;
         }
-        throw new Sk.builtin.StopIteration();
+        const v = iter.$value;
+        if (v !== undefined && v !== Sk.builtin.none.none$) {
```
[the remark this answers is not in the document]
s-cork: Yes - you'd think this was redundant - but annoyingly - since we've lost the original StopIteration we don't know if was 
1. `StopIteration()`
2. `StopIteration(None)`

And if you return from a generator it's always `StopIteration()` over `StopIteration(None)`

In these two cases the `.value` property is both None. But the `.args` are different in these cases.
Which makes the reprs different
Which makes the tests fail

I've added a comment about this in the latest pr.
Essentially this is making a best guess that it was probably `StopIteration()` over `StopIteration(None)`
🤷

src/compile.js
```diff
@@ -511,6 +511,37 @@ Compiler.prototype.cyield = function(e) {
...
+    out("$ret=", iterable, ".tp$iternext(true, $gen.gi$sentvalue);");
+    out("} else {");
+    var send = this.makeConstant("new Sk.builtin.str('send');");
+    out("$ret=Sk.misceval.tryCatch(function(){return Sk.misceval.callsimOrSuspendArray(Sk.abstr.gattr(", iterable, ",", send, "), [$gen.gi$sentvalue]);},");
+    out("function (e) { if (e instanceof Sk.builtin.StopIteration) { " + iterable + ".$value = e.$value; return undefined; } else { throw e; } }");
+    out(");");
+    out("}");
+    this._checkSuspension(e);
+    out(retval, "=$ret;");
+    out("if(", retval, "===undefined){$gen.gi$sentvalue=$gen." + iterable + ".$value || Sk.builtin.none.none$;$blk=", afterBlock, ";continue;}");
```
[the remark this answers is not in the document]
s-cork: I couldn't find one that did what I needed here.

src/simple_iterators.js
```diff
@@ -66,6 +66,7 @@ Sk.builtin.seq_iter_ = Sk.abstr.buildIteratorClass("iterator", {
             },
             (e) => {
                 if (e instanceof Sk.builtin.IndexError || e instanceof Sk.builtin.StopIteration) {
+                    this.$value = e.$value || Sk.builtin.none.none$;
```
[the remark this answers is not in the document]
s-cork: Done

src/compile.js
s-cork: Added comments and some formatting to make the code more readable

src/builtin.js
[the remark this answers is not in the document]
s-cork: I think I've fixed this redundancy #1306

meredydd (approved):
So much nicer to understand now! Ship it :)

meredydd:
Does anyone else want to review this one? It feels like @s-cork and I are doing most of the compiler stuff atm, so unless @albertjan wants to jump in I think we're just going to merge this.

albertjan:
👀

src/errors.js
```diff
@@ -474,9 +474,27 @@ Sk.exportSymbol("Sk.builtin.UnicodeEncodeError", Sk.builtin.UnicodeEncodeError);
  */
 Sk.builtin.StopIteration = Sk.abstr.buildNativeClass("StopIteration", {
     constructor: function StopIteration(...args) {
+        this.$value = args[0] || Sk.builtin.none.none$;
         Sk.builtin.Exception.apply(this, args);
     },
     base: Sk.builtin.Exception,
+    slots: {
```
albertjan, line 481: this new way of describing classes is so much nicer 🎉

test/unit3/test_yield_from.py
```diff
@@ -0,0 +1,1053 @@
...
+            "Yielded g1 ham",
+            "Starting g2",
+            "Yielded g2 spam",
+            "Yielded g2 more spam",
+            "Finishing g2",
+            "Yielded g1 eggs",
+            "Finishing g1",
+        ])
+
+    # def test_raising_exception_in_delegated_next_call(self):
```
albertjan, line 107: @s-cork could you add a comment why this test doesn't pass?

src/builtin.js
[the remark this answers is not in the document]
albertjan: 🤯

test/unit3/test_yield_from.py
s-cork: Sure - it's because we don't have try finally support in generators. 
generators don't save enough state to handle try, except, finally blocks.
(they'd need to save the `$err` from the compilation code and the `$postfinally`)

I can add a comment but it might be premature since I uncomment these tests in #1306

src/compile.js
```diff
@@ -511,6 +511,58 @@ Compiler.prototype.cyield = function(e) {
     return "$gen.gi$sentvalue"; // will either be none if none sent, or the value from gen.send(value)
 };
 
+Compiler.prototype.cyieldfrom = function (e) {
```
albertjan, line 514: In python all this work is done in `compiler_sync_comprehension_generator` I'm slightly worried we're trying to keep the compiler looking like the python2 compiler 😄 

Not something we should fix in this PR just something I wanted to flag.

albertjan (approved):
🚢

s-cork:
Awesome - just did a rebase on master after conflicts with `Sk.builtin.next_` from #1302 - ready for merge

src/compile.js
s-cork: I used the code from ceval.c `Target(YIELD_FROM)` for the general structure.
https://github.com/python/cpython/blob/main/Python/ceval.c#L2563

PR #1242, opened 2020-12-21
fix(turtle): speed function support string parameter

bofeng:
According to the official Python spec here: https://docs.python.org/3/library/turtle.html#turtle.speed

The current version:
1, speed doesn't take string parameter: fastest, fast, normal, slow, slowest
~2, set speed(0) will still draw things with animation, while it should draw instantly w/o animation~
3, set speed(val) where val > 10 or < 0.5, speed is not set to be 0

This pull request fixed the above 3. 

Also, the current version computes an internal speed value using `max(0, 2 * speed - 1)` for frame update, I changed it to `5 * speed`. This will make the turtle go faster than the current version when speed is set to 10 or "fast". Since "slow" and "fast" are just relative, as long as speed(1) is slower than speed(10), I think this change should be fine.


src/lib/turtle.js
```diff
@@ -849,9 +849,39 @@ function generateTurtleModule(_target) {
...
+                    } else if (speed === "fast") {
+                        speed = 10;
+                    } else if (speed === "normal") {
+                        speed = 6;
+                    } else if (speed === "slow") {
+                        speed = 3;
+                    } else if (speed === "slowest") {
+                        speed = 1;
+                    } else {
+                        speed = this._speed;
```
s-cork, line 868: in python if the speed string is not one of these values then an error will be raised (the internal implementation of python will try to to compare int and str)

perhaps something like

```javascript
var speeds = {fastest: 0, fast: 10, ... }
if (speed in speeds) {
   speed = speeds[speed]
}
if (typeof speed !== "number") {
  throw ...
}
```

src/lib/turtle.js
```diff
@@ -849,9 +849,39 @@ function generateTurtleModule(_target) {
...
+                if (typeof(speed) === "number") {
+                    if (speed > 10 || speed < 0.5) {
+                        speed = 0;
+                    }
+                    if (speed === 0) {
+                        _config.animate = false;
+                        getFrameManager().refreshInterval(0);
+                    } else {
+                        _config.animate = true;
+                        getFrameManager().refreshInterval(this.$delay());
```
s-cork, line 880: in python these values are rounded to the nearest int
```python
>>> t.speed(1.4)
>>> t.speed()
1
```
so maybe we should do the same here

src/lib/turtle.js
bofeng: I added a new commit for this one, the exception is a little tricky. In Python it will throw `TypeError: '<' not supported between instances of 'float' and 'str'` if the speed string is not one of "fastest", "fast", "normal", "slow", "slowest" when it tries to run code `elif 0.5 < speed < 10.5:`; Here I threw a TypeError to indicate it is not one of the five value, but technically I guess it should be "ValaueError" ? So TypeError here is just trying to make it throw the same exception like Python does. Thoughts?
```
            if (typeof(speed) !== "number") {
                if (typeof(speed) === "string") {
                    const valid_speeds = Object.keys(speeds).join(", ");
                    throw new Sk.builtin.TypeError("speed string expected one of " + valid_speeds);
                }
                throw new Sk.builtin.TypeError("speed expected a string or number");
            }
```

src/lib/turtle.js
bofeng: This one b/c we need to consider the bankers rounding, different behavior in Py2 and Py3, e.g. 2.5 rounds to 3 in Py2, 2.5 rounds to 2 in Py3, I used the builtin round function to do the rounding, which takes care of "bankers rounding" on/off case:
```
speed = Sk.builtin.asnum$(Sk.builtin.round(Sk.builtin.assk$(speed)));
```

alexwenbj:
Without animation is only for one-step drawing, not all the drawings.For example:
for i in range(4):
    forward(100)
    left(90)
both forward(100) and left(90) will execute without animation, but not all the 8 steps drawings.

The original speed(0) implementation is correct.

bofeng:
> Without animation is only for one-step drawing, not all the drawings.For example:
> 
> ```
> for i in range(4):
>     forward(100)
>     left(90)
> ```
> 
> both forward(100) and left(90) will execute without animation, but not all the 8 steps drawings.
> 
> The original speed(0) implementation is correct.

@alexwenbj ah, I see, so even with speed(0), animations between steps still exist. Tested in standard Python's IDLE, works exactly like you said. Thank you very much! 
I'll remove the "no animation" part code change, so it will just fix the "string parameter" and "speed value" bug.

bofeng:
I committed the change, so now this pull-request only fix the below 2 problems:
1. speed doesn't take string parameter: fastest, fast, normal, slow, slowest
2. set `speed(val)` where val > 10.5, speed is not set to be 0; and if val is a decimal within the range, speed is not set to the rounded number (need to consider the bankers rounding on/off cases).

I also changed the internal computed_speed to `2 * speed`, it was `max(0, 2 * speed - 1)` mainly for taking care of the < 0.5 case, but since now the > 10.5 and < 0.5 case has been handled in a separated condition statement, the minus 1 part is not needed anymore.

@s-cork @alexwenbj  Thank you again for reviewing it! Please let me know if further editing needed.

s-cork (approved):
LGTM but turtle isn't something in skulpt I've spent much time with. @skulpt/committers over to you.

bnmnetp (approved):

PR #1240, opened 2020-12-19
Implement __slots__

s-cork:
This pr implements `__slots__`
it's 1 commit ahead of #1092 
For the diff see the last commit on this pr or you can check it out here: 590d8ac

It includes all the tests from cpython unittests (with a couple of edge cases commented out)

I also needed to implement `isidentifier` for `str`

The copy module does't work for classes that define `__slots__` but since `__slots__` previously wasn't supported, copying them wouldn't really have worked anyway.
(to fix the copy module we can look to define the `__reduce_ex__` descriptor)

closes #1050

PR #1244, opened 2020-12-22
implement metaclass, prepare, init_subclass and set_name

s-cork:
this pr implements `metaclass`, `__prepare__` as well as `__init_subclass__` and `__set_name__` from [pep487](https://www.python.org/dev/peps/pep-0487/)

this pr is 1 commit ahead of #1240 
for the diff see the last commit: f14c6fd

I've added tests from cpython from `test_init_subclass.py`
as well as uncommenting tests from `test_descr.py` that deal with metaclasses
This pr doesn't implement the protocol for `mro` and so those tests were left commented out

closes #1209

---
**super**
I changed the compile implementation for super which is currently incorrect since it assumes that super should always be called with the variable named `self`.
This is a problem for class methods like `__new__` that typically use `cls` and not `self`.
I create a variable`$sup` that is set to `arg[0]`. 
It's not perfect.
I had to comment out an obscure runtime bug where if you delete arg[0] then the $sup still has a reference to this and doesn't raise a `RuntimeError`. I think this is obscure enough to sacrifice.

But there might be a better way to do this. Keep alive the current reference to the `self` argument to `super` somehow.


s-cork:
I've updated this pr - it still relies on `__slots__` (#1240) as per the description.

In the final commit I've added, what I think is, a clever hack when using a `metaclass` and the `__prepare__` mechanism.
There are two main reasons to use `__prepare__`.
1. pre-populate `locals` before calling the codeobject
2. provide an alternative mapping that isn't a `dict` object

Since we use a javascript object for `locals` the use of `__prepare__` for option 2 might be pointless!
So instead - if `__prepare__` returns a `mapping` which is not a strict `dict`, then `locals` becomes a `Proxy` for that `mapping` so now doing some mind blowing 🤯  python hacks with a custom mapping should just work 🤞 .

The caveat is that IE does not support `Proxy` and the polyfill for es5 isn't sufficient because the polyfill only supports `get` and not `set`. So if it's IE (which we should stop supporting soon anyway) then we don't use the `Proxy` mechanism.
(All other es6 compliant browsers have `Proxy` and I think this approach is a nice compromise)

meatballs:
Is there some objection to this pr? Any reason not to merge it?

PR #1248, opened 2021-01-04
Add ability to delete attributes

bennorth:
Add support for
```python
del foo.bar
```
which deletes the attribute named `bar` from the object `foo`.

CPython handles attributes deletion by passing `NULL` to the set-attribute machinery.  I've largely taken the same approach here.  This PR's implementation of some details doesn't completely follow the current CPython one, but I think the end result is workable.  I'm not 100% familiar with the existing code in this area, so please let me know if I've overlooked anything.  Thanks.

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -879,7 +879,8 @@ Compiler.prototype.vexpr = function (e, data, augvar, augsubs) {
                     this._checkSuspension(e);
                     break;
                 case Sk.astnodes.Del:
-                    Sk.asserts.fail("todo Del;");
+                    out("$ret = Sk.abstr.sattr(", val, ",", mname, ", null, true);");
+                    this._checkSuspension(e);
                     break;
                 case Sk.astnodes.Param:
                 default:
--- a/src/constants.js
+++ b/src/constants.js
@@ -35,6 +35,7 @@ Sk.builtin.str.$cmp = new Sk.builtin.str("__cmp__");
 Sk.builtin.str.$complex = new Sk.builtin.str("__complex__");
 Sk.builtin.str.$contains = new Sk.builtin.str("__contains__");
 Sk.builtin.str.$copy = new Sk.builtin.str("__copy__");
+Sk.builtin.str.$delattr = new Sk.builtin.str("__delattr__");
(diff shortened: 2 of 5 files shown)


s-cork:
note that this is also addressed in #1092 

I took a similar approach - but used `undefined` rather than `null` for no particular reason other than javascript had a choice where C doesn't. 
I also used `tp$descr_set` for deleting and setting, as opposed to the slot `tp$descr_del` since the `tp$descr_del` isn't a thing in cpython. (the `tp$descr_set` is responsible for setting and deleting in the same way that `tp$setattr` is responsible for setting and deleting attributes). But this is certainly a way we can decide to deviate from Cpython. 

if #1092 get's merged then this should be closed. If #1092 isn't merged this approach looks good to me.

bennorth:
@s-cork:
> \[In #1092, I\] used `tp$descr_set` for deleting and setting, as opposed to the slot `tp$descr_del` since the `tp$descr_del` isn't a thing in cpython. (the `tp$descr_set` is responsible for setting and deleting in the same way that `tp$setattr` is responsible for setting and deleting attributes). But this is certainly a way we can decide to deviate from Cpython.

Thanks for the pointer towards #1092, which I'd missed, sorry.  I was wondering about `tp$descr_del`.  The CPython machinery for handling special methods as slots in the `PyTypeObject` didn't look straightforward to port to Skulpt, so I took the easy way out, and added another `tp$...` slot.

@s-cork again:
> if #1092 get's merged then this should be closed. If #1092 isn't merged this approach looks good to me.

Might it be worth merging this present PR as a stop-gap until #1092 is merged, if everyone is OK with it?

s-cork (approved):

s-cork:
closed via #1092 - I think this is well tested in 1092 via `unit3/test_descr.py` but adding these tests to the test suite wouldn't hurt - especially if they're tests from cpython.

bennorth:
@s-cork:
>  I think this is well tested in 1092 [...] but adding these tests to the test suite wouldn't hurt

I've just created #1269 which cherry-picks the commit adding the tests.  They do indeed all pass :)  That PR also adds another one-off test which had previously failed, but now passes post-#1092.

PR #1251, opened 2021-01-06
Bugfix: Overlooked localtime f usage

bennorth:
As @s-cork notes in #1233, I overlooked a usage of `localtime_f()` in that refactoring, sorry.  Add a test for the situation where this comes up, and fix the failure by using the new `from_seconds()` instead.


s-cork (approved):

s-cork:
Small change. Merged.

PR #1253, opened 2021-01-09
updated datetime module

s-cork:
here's an updated version of the datetime module.

this is based on the #1240 pr since datetime classes uses slots for classes it seemed to make sense.

I took the current cpython pure python implementation of the datetime module and made some tweaks for backward compatibility of python 2.7.

I've incorporated tests from cpython and commented out tests that weren't relevant or that didn't pass.

---

For the relevant commits see the last 6 commits in this pr.
be1e9d5: float as_integer_ratio - required in the datetime implementation
5fa5843: adjust struct_seq for time module to allow tm_gmtoff and tm_zone
5fc1666: datetime tests from cpython
4576fdd: adjusted datetime tests for skulpt
84505f8: cpython's pure python implementation of datetime
53e9c25: adjusted datetime module for skulpt

At the moment the datetime tests take 25 seconds on my machine and the whole of unit 3 tests take 100 seconds!
This seems a little excessive - I can look to reduce iterations for tests that are causing the bottle neck perhaps.

I also could look to rewrite the datetime module from pure python into javascript.
This would improve the speed of the module - but may reduce the maintainability. Thoughts?

If this pr is merged we may want to consider the size of pure python implementations in the `skulpt-stdlib` - The datetime module is big! 
Perhaps we can look to do some minification of python files at the same time that we minify the javascript stdlib files.

--- 

This pr doesn't implement `strptime` - it seemed like that should be a separate pr.

closes #1250

PR #1254, opened 2021-01-14
updated datetime module in javascript

### LINKED ISSUE #1250: datetime needs an update
cpython now has a pure python implementation of the datetime module
Skulpts implementation is minimal and incomplete since it's taken from an early version of pypy.

now that we have bytes we might be able to port cpython's implementation (with some adjustments to support python 2.7). 

I'm currently working on this and hope to have a pr ready soon.

I might even see if i can implement `strptime` (at least partially).

Turning this into an issue incase anyone else is working on this already.

s-cork:
This pr is an update of the datetime module based on #1253 but written in javascript rather than python

Where #1253 `test_datetime.py` takes around 25 seconds on my machine, this pr takes around 3 seconds. 😄 

All tests passing from #1253
I've left this as a separate pr for a couple of reasons:
- I use es6 and we currently don't have a transpile options for es5 in our stdlib modules (this will break ie11)
- #1253 might be more trustworthy since it is just the cpython version with a handful of tweaks

In general I've tried to make the code very similar to the pure python version. Hopefully this will make it more maintainable to anyone who needs to update it in the future. 
Functions are generally in the same location, have the same names and follow the same algorithm.

The advantage of this pr is speed and code size (this will get minified where #1253 won't)

closes #1250, closes #1253

For the javascript implementation see the last commit in this pr.


bnmnetp:
This seems to touch a LOT of files for something that is implementing a library module.  Or is this just not up to date with the latest master?

I think this javascript implementation seems like the way to go especially as it will let us add strptime.

Can you look at the conflicts @s-cork?  I'll try to test and review then.

s-cork:
@bnmnetp absolutely - i'm working my way through the backlog of prs that depended on #1092 and making them mergable - I haven't reached this one yet 😄  - but when I do it should be only a couple of files changed. You'll see a force push on this pr when I get there.

s-cork:
Rebase on this one complete

I've implemented the same static type change that I used in functools. 
This is so reprs get displayed correctly for builtintypes from the stdlib.
I fixed a bug bytes (I found this during the work on this pr)
There was also another niche change. `typeName` and `tp$name` are subtly different and binops use `tp$name` in cpython rather than `typeName`. I also found this when I was working with datetime objects.

```python
>>> from datetime import timedelta
>>> timedelta() + 1
TypeError: unsupported operand type(s) for +: 'datetime.timedelta' and 'int'
# this uses tp$name since the module.name is reported
# if it was typeName then it would just be 'timedelta'
```
A bit of pointless change maybe. But hey.

I've left the updated `datetime.py` module in the commit history for reference.

meredydd (approved):
I'm approving this, pretty much entirely on the basis that it passes CPython's tests. Datetime handling code is big, and famously gnarly, and I don't think we can reasonably expect any of us to spot tiny mistakes that the tests won't. Let's ship it.

bnmnetp (approved):

PR #1257, opened 2021-01-16
add strptime support to datetime module with tests

s-cork:
built on top of #1254 

It is heavily based on the _strptime python implementation.
Like #1254 I have tried to ensure that function names, function order and function logic is very close to the cpython implementation. 

Equivalent code in cpython:
https://github.com/python/cpython/blob/master/Lib/_strptime.py

Tests added from `test_strptime.py` and uncommented tests from `test_datetime.py`

This pr uses es6 in stdlib so problematic for an es5 build.

The last 3/4 commits are relevant to this PR. 

Todo - we should potentially look to improve our `strftime` implementation. 
Currently we don't support different locales and `"%G"` and `"%V"` are not currently supported


alexwenbj:
A bug introduced in the new version:(I used your 1092_all_changes pr)
```
from datetime import timedelta

d = timedelta(microseconds=1)
print(d)
```
It raises an error:
`TypeError: 'float' object cannot be interpreted as an integer`
The code above is OK in the skulpt.org, so maybe it is this new implementation introduces a bug. @s-cork

s-cork:
I tried on this pr branch and `1092/all_changes` and I don't get the error 🤷‍♂️ ??

What am i missing?

Screen Shot 2021-02-10 at 18 41 48

alexwenbj:
Sorry, it is my fault, it is OK in my another env. I should do sth wrong in the buggy env, I will check it. @s-cork

alexwenbj:
It is very strange, the code only broke in my env with file id 4361 and 4362,for example:  http://localhost/python/4361 or python/4362,and the number is python file id saved in db...

s-cork:
rebased on top of #1254 - this pr is 4 commits ahead

albertjan (approved):
Nice!

meredydd (approved):
Approved, on a similar basis to `datetime` - with code that hairy, we have to trust the tests (and passing the CPython tests is a pretty good indicator!)

Issue #1014, opened 2020-02-03
Proposed Breaking Changes for Performance


meredydd:
In the process of #1009 and #1013, I have made a list of warts in the Skulpt API that are standing in the way of optimisation improvements.

I would like to propose landing #1009 and #1013, then cutting a release (hi @bnmnetp), then making all these breaking changes at once, in a well-signalled and -documented way, so that our users can follow along. Tagging all @skulpt/committers for feedbck, and we should really ping the rest of our big users too.

Here's my hit list (with benefits to us, and costs to our users, spelled out). Contributions - and feedback from other Skulpt users - welcome!

---

**1. Stricter calling convention - `tp$call()` only**

`Sk.misceval.applyOrSuspend()` supports calling some weird stuff. Raw JS functions, JS objects with `__call__` methods, etc. We should declare that anything callable has a `tp$call` slot, and anything that doesn't, isn't.

**Benefit:** We can remove cruft from `apply()`, and make generated function call sites simpler (ie faster).

`applyOrSuspend()` then becomes vestigial - its call signature is no longer particularly useful to the compiler (because Python 3 supports multiple `*varargs`/`**kwargs`), and should be deprecated or (**more radical**) retired.

`callsimArray()` becomes the officially supported external interface. It becomes a simple wrapper for `obj.tp$call(args, kws)` that throws a nice "that's not callable" error if `tp$call` isn't present.

**Cost:** Probably minimal-to-zero. More if we deprecate `applyOrSuspend()`.

----

**2. Remove direct use of dunders**

There is lots of code that checks `obj.__call__` or `obj.__len__`, when it should just be hitting `tp$call` or `sq$len` slots. (Some of this code is on hot paths, or paths that were hot and now handle the 99.9% case but call back to this vestigial thing because we don't want to break old weird user code.)

**Benefit:** Some perf, mostly code size & comprehensibility

**Cost:** Probably minimal-to-zero

----

**3. Stricter attribute convention - `tp$getattr` only**

Right now, `Sk.abstr.gattr()` (which is very hot) has a bunch of checks for whether it's being called on `null`, or on some weird native JS object that doesn't have `tp$getattr`. We should drop this -- the only valid values should be real Python objects, and all real Python objects should have `tp$getattr`.

At this point, all `Sk.abstr.gattr(obj, name)` calls from native code can safely be replaced with `obj.tp$getattr(name)`, which removes overhead and is more optimiser-friendly.

[...]

PR #1261, opened 2021-01-24
Add partial docstring support

bennorth:
Assign `__doc__` attribute to functions, methods, classmethods, classes, and modules.

\[Edited: the first version of this PR did not support module docstrings.\]

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -2163,6 +2163,29 @@ Compiler.prototype.maybeCDocstringOfBody = function(body) {
     return this.vexpr(expr);
 };
 
+/** JavaScript for the docstring of the given node.  Only called from
+ * buildcodeobj(), and expects a FunctionDef, Lambda, or GeneratorExp
+ * node.  We give a "None" docstring to a GeneratorExp node, although
+ * it is not carried over to the final generator; this is harmless.
+ */
+Compiler.prototype.cDocstringOfCode = function(node) {
+    switch (node.constructor) {
+    case Sk.astnodes.AsyncFunctionDef:  // For when it's supported
+    case Sk.astnodes.FunctionDef:
+        return (
+            this.maybeCDocstringOfBody(node.body)
+            || "Sk.builtin.none.none$"
+        );
+
+    case Sk.astnodes.Lambda:
+    case Sk.astnodes.GeneratorExp:
+        return "Sk.builtin.none.none$";
+
+    default:
+        Sk.asserts.fail(`unexpected node kind ${node.constructor.name}`);
+    }
+}
+
 Compiler.prototype.cfunction = function (s, class_for_super) {
     var funcorgen;
     Sk.asserts.assert(s instanceof Sk.astnodes.FunctionDef);
--- a/src/compile.js
+++ b/src/compile.js
@@ -2138,7 +2138,6 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
         }
     }
     else {
-        var res;
         if (decos.length > 0) {
             out("$ret = new Sk.builtins['function'](", scopename, ",$gbl", frees, ");");
             for (let decorator of decos.reverse()) {
--- a/src/compile.js
+++ b/src/compile.js
@@ -2080,6 +2080,13 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
     }
 
     //
+    // Skulpt doesn't have "co_consts", but we keep the docstring (or None)
+    // with the code object for possible future use with a descriptor.  This
+    // happens for named functions, lambdas, and generator-expressions.
+    //
+    out(scopename, ".co_docstring=", this.cDocstringOfCode(n), ";");
+
+    //
     // attach flags
     //
     if (kwarg) {
--- a/src/compile.js
+++ b/src/compile.js
@@ -2145,6 +2145,24 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
     }
 };
 
+/** JavaScript for the docstring of the given body, or null if the
+ * body has no docstring.
+ */
+Compiler.prototype.maybeCDocstringOfBody = function(body) {
+    if (body.length === 0)  // Don't think this can happen?
+        return null;
+
+    const stmt_0 = body[0];
+    if (stmt_0.constructor !== Sk.astnodes.Expr)
+        return null;
+
+    const expr = stmt_0.value;
+    if (expr.constructor !== Sk.astnodes.Str)
+        return null;
+
+    return this.vexpr(expr);
+};
+
 Compiler.prototype.cfunction = function (s, class_for_super) {
     var funcorgen;
     Sk.asserts.assert(s instanceof Sk.astnodes.FunctionDef);
--- a/src/compile.js
+++ b/src/compile.js
@@ -2730,8 +2730,19 @@ Compiler.prototype.exitScope = function () {
  * @param {Sk.builtin.str=} class_for_super
  */
 Compiler.prototype.cbody = function (stmts, class_for_super) {
-    var i;
-    for (i = 0; i < stmts.length; ++i) {
+    var i = 0;
+
+    // If we have a docstring, then assign it to __doc__, and skip over
+    // the expression when properly compiling the rest of the body.  This
+    // happens for class and module bodies.
+    //
+    const maybeDocstring = this.maybeCDocstringOfBody(stmts);
+    if (maybeDocstring !== null) {
+        out("$loc.__doc__ = ", maybeDocstring, ";");
+        i = 1;
+    }
+
+    for (; i < stmts.length; ++i) {
         this.vstmt(stmts[i], class_for_super);
     }
 };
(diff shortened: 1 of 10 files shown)


s-cork:
a couple of observations that might help:

```python
class A:
    """docstring"""
    x = __doc__

A.x # 'docstring'

```
Which means that we should really compile this as 
`$loc.__doc__ = $scope123.$const456`
as the first line within the class scope rather than adding it retrospectively after the `buildClass` call.

and for functions the current compile code in this pr for
```python
def foo():
   """bar"""
```

```javascript
$scope156.co_name = new Sk.builtins['str']('foo');
$scope156.co_varnames = [];
$scope156.co_fastcall = 1;
var $funcobj159 = new Sk.builtins['function']($scope156, $gbl);
$funcobj159.$d["__doc__"] = $scope155.$const160;
$loc.foo = $funcobj159;
```

In cpython the doc string of a function is always the first constant in `co_consts`

```python
def foo():
   """bar"""

>>> foo.__code__.co_consts
['bar', None]
```
skulpt doesn't use `co_consts` but I think it would be better to add the docstring to the code obj in someway, before the call to `Sk.builtins.func` which would also make it easier in the future to return the docstring from a `__doc__` descriptor 😄 .

Worth also considering functions that get user defined rather than compiled - so that the `__doc__` would return `None` in this case.

bennorth:
Thanks for that feedback, @s-cork.  All good points — I'll look into them and update this PR when done.

bennorth:
I've just force-pushed a fresh set of commits, which I think addresses the points from @s-cork:

#### Immediate assignment to `__doc__` in class and module body

This is now done within `Compiler.cbody()`, and the case
```python
class A:
    """docstring"""
    x = __doc__

A.x # 'docstring'
```
now behaves as expected.  It also gives us module docstrings.  The unit tests now cover these cases.

#### Store docstring in code object

> In cpython the doc string of a function is always the first constant in `co_consts` [...] skulpt doesn't use co_consts but I think it would be better to add the docstring to the code obj in someway

I've assigned to `co_docstring` in the code object.  This is currently unused, but as you say it could be used at some point by a descriptor to extract the docstring.

#### User-defined functions

> Worth also considering functions that get user defined rather than compiled

I'm not quite sure what you mean here, sorry.  Do you mean things like `f = lambda x: x+1` (the proposed code gives lambdas a `None` docstring)?  Or closures like
```python
def make_adder(n):
    "Function adding N"
    def add(x):
        "Compute N + X"
        return n + x
    return add
```
(this is from the unit tests)?

s-cork:
Sorry I meant that when I do something like...

```javascript
var foo = new Sk.builtin.func(function foo() {}); // i.e. not through the compiler but somewhere in my javascript
```
I also would want it to have a `__doc__` that returns `None`
I could also then do
```javascript
function foo () {};
foo.co_docstring = new Sk.builtin.str('bar')
mod.foo = Sk.builtin.func(function foo() {}); // I now have a __doc__
```

so rather than setting the `$d` in compile code, I was thinking that the func constructor should set the `__doc__` to `co_docstring || None`

If that makes sense.

bennorth:
Aha, I see, thanks.  That suggestion simplifies things nicely.  I've pushed an additional commit (plus a test) so you can check whether this does what you mean, and if so, I'll rebase/squash so the PR has a sensible chain of commits.  Let me know what you think.

Would it be worth adding a small helper function to streamline your
```javascript
function foo () {};
foo.co_docstring = new Sk.builtin.str('bar')
mod.foo = Sk.builtin.func(foo); // I now have a __doc__
```
use-case slightly?  Something like
```javascript
Sk.builtin.funcWithDoc = function (func, jsDocstring) {
    func.co_docstring = new Sk.builtin.str(jsDocstring);
    return func;
};
```
so your example could do
```javascript
function foo () {};
mod.foo = Sk.builtin.func(Sk.builtin.funcWithDoc(foo, "bar")); // I now have a __doc__
```
This doesn't gain much readability for a named function, but would avoid having to give an inline function a name just to give it a docstring.  An alternative name would be `withDocstring`.  I'll give it more thought and add a commit providing this, if you like the idea.

src/compile.js
```diff
@@ -2145,6 +2152,47 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
...
+            this.maybeCDocstringOfBody(node.body)
+            || "Sk.builtin.none.none$"
+        );
+
+    case Sk.astnodes.Lambda:
+    case Sk.astnodes.GeneratorExp:
+        return "Sk.builtin.none.none$";
+
+    default:
+        Sk.asserts.assert(false, `unexpected node kind ${node.constructor}`);
```
s-cork, line 2192: you can use `Sk.asserts.fail(msg)` here

src/compile.js
```diff
@@ -2683,8 +2731,19 @@ Compiler.prototype.exitScope = function () {
...
-    var i;
-    for (i = 0; i < stmts.length; ++i) {
+    var i = 0;
+
+    // If we have a docstring, then assign it to __doc__, and skip over
+    // the expression when properly compiling the rest of the body.  This
+    // happens for class and module bodies.
+    //
+    const maybeDocstring = this.maybeCDocstringOfBody(stmts);
+    if (maybeDocstring != null) {
```
s-cork, line 2741: we can use `!==` here

src/compile.js
```diff
@@ -2145,6 +2152,47 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
     }
 };
 
+/** JavaScript for the docstring of the given body, or null if the
+ * body has no docstring.
+ */
+Compiler.prototype.maybeCDocstringOfBody = function(body) {
+    if (body.length == 0)  // Don't think this can happen?
```
s-cork, line 2159: suggestion `===`

src/compile.js
```diff
@@ -2145,6 +2152,47 @@ Compiler.prototype.buildcodeobj = function (n, coname, decorator_list, args, cal
...
+        return null;
+
+    const stmt_0 = body[0];
+    if (stmt_0.constructor !== Sk.astnodes.Expr)
+        return null;
+
+    const expr = stmt_0.value;
+    if (expr.constructor !== Sk.astnodes.Str)
+        return null;
```
s-cork, line 2168: in the style guide we should add `{` but compile.js is ignored for es linting

s-cork (approved):
lgtm - nice pr - seems very closely aligned to cpython - @skulpt/committers ready for you

one other thing you might want to look at is in `type.js`
```javascript
        // set __module__ if not present
        if(dict.mp$lookup(Sk.builtin.str.$module) === undefined) {
            dict.mp$ass_subscript(Sk.builtin.str.$module, Sk.globals["__name__"]);
        }
```
You could do something similar for `__doc__` so that even a class that doesn't define a `__doc__` still gets `None`

s-cork:
re the helper function - I think that's beyond the scope of this pr but an interesting discussion point and presents a wider question about our api for instantiating a `new Sk.builtin.func` with the correct `__name__` and `__doc__` from javascript. See #1014 where I think the api is also mentioned. Currently if you try to print most of the builtin funcs you get the `<Native JS>` rather than the actual name of the object. 

in my head it would be something like this. 

```javascript
mod.foo = new Sk.builtin.func(new Sk.builtin.codeobj(() => {
    /// funciton body
}, {
    name: "foo",
    doc: "bar",
    //... etc
}));
```

bennorth:
Thanks for further feedback.  I've made the individual changes you suggest, leaving out the `{}`s on the `return null`s.  I also now assign to `__doc__` of a class if there is no docstring given.  New set of commits force-pushed.

src/compile.js
bennorth: Thanks; didn't spot that one!

src/compile.js
bennorth: I've left it without the `{}`s to keep it concise; happy to put them in if preferred.

bennorth:
Agreed that an API for adding a docstring when calling `Sk.builtin.func()` would be scope-creep here — thanks for writing the idea up (#1263).

s-cork (approved):

bnmnetp:
@bennorth it seems the PR I accepted yesterday has caused conflicts in several good PRs.  I would like to get this accepted if you can resolve these...

bennorth:
@bnmnetp The machinery added by #1092 made a few of the changes in this PR unnecessary, so I thought it would be cleaner to create a separate PR — #1265 — for the docstring feature, now that #1092 is in.  If you're happy with that approach, I can close this PR.

bnmnetp:
Nice!  That does simplify this PR considerably!

You can go ahead and close this one.

BTW, I tried to have a look at Pytch but could not get the tutorial to work.  Something in the console about an unhandled Promise.  This seems like it could be an interesting addition to the RunestoneInteractive project.

bennorth:
@bnmnetp Thanks; will do.

(And many thanks for looking at Pytch.  I don't want to hijack this PR for Pytch troubleshooting, but: We know that it doesn't work under Safari, and will investigate as resources allow.  Meanwhile, if you had time for another look using Chrome or Firefox, that would be great!  I hadn't come across Runestone Interactive before now — its goals are excellent, and I will read about it in more detail.)

PR #1262, opened 2021-01-25
fix a typo.

alexwenbj:
fix a pypo.


s-cork:
I wonder if we should also remove this from the `builtindict.js` in python 3 mode since it's not available in py3?

alexwenbj:
> I wonder if we should also remove this from the `builtindict.js` in python 3 mode since it's not available in py3?

Yes, I think it should be removed from py3.

bnmnetp (approved):
This LGTM, but @alexwenbj we need to fix the merge conflicts.  Once those are done we are ready to merge.

meredydd:
Hi @alexwenbj can you rebase this so we can get it merged?

alexwenbj:
Sorry, I didn't realize the conflicts and resolved late. Get things done now. @meredydd

s-cork:
added a small commit to make sure unichr is a py function and get's added back in the switch from py3 to py2.

PR #1264, opened 2021-02-09
Suspensions - throw them - don't return them

s-cork:
This is an idea inspired by how React does Suspense.

In react you can cause a Suspense by throwing a promise.
So I thought I'd apply the idea to Skulpt.

I figured this might be advantageous performance wise, because it saves us from doing instances checks at every potential suspending operation.

Since our compile code is already wrapped in a big try catch we can then handle the Suspensions in the catch part.

---

### Compile Example

```python
foo() # lets compile a simple function call
```
would previously be compiled as (not all the compile code just the relevant parts)

```javascript
while (true) {
    try {
        switch ($blk) {
            case 0:
                var $loadname1 = $loc.foo !== undefined ? $loc.foo : Sk.misceval.loadname("foo", $gbl);
                $ret = $loadname1.tp$call([], undefined);
                $blk = 1;
            case 1:
                // resume return from suspension
                if ($ret && $ret.$isSuspension) {
                    return $saveSuspension($ret, "<stdin>.py", 1, 0);
                }
                var $call2 = $ret;
        }
    } catch {
        // handle err
    }
}

```

would now get compiled as
```javascript
while (true) {
    try {
        switch ($blk) {
            case 0:
                var $loadname1 = $loc.foo !== undefined ? $loc.foo : Sk.misceval.loadname("foo", $gbl);
                $blk = 1; // set the next block now since the function call might throw a suspension
                $ret = $loadname1.tp$call([], undefined);
            case 1:
                // resume return from suspension
                var $call2 = $ret;
        }
    } catch {
        if (err instance of Sk.misceval.Suspension) {
            $saveSuspension(err)
        }
        // handle err 
    }
}

```

You can see that we handle the suspension in the catch, rather than checking whether `$ret` is an instance of `Suspension`

---

### Breaking changes

The major change for this api is that you can't call `Sk.misceval.chain` with a value - it must be called with a function since that function might throw a Suspension. The same goes for `retryOptionalSuspensionOrThrow`.

---

### Performance

Fields.py runs 5-10% faster. So not massive but an improvement nonetheless.
And we can probably improve performance further by inlining certain functions as it becomes easier to handle suspensions in compile code since they no longer require instance checks.

---

### Other changes with this PR

I had to adjust anywhere that used `Sk.misceval.chain`/`Sk.misceval.retrySuspensionOrThrow`. 

I also ended up with a bug in import.js because delaying evaluation of `Sk.misceval.chain` led to a self referencing stack overflow 😞 . It was simple enough to fix. But then I got carried away and modernised the `import.js` file.

I also changed the `Suspension` constructor. I didn't like how the resume function wrapped the `child.resume`. There is no example where a suspension doesn't end up overriding the resume function after construction. So I've changed this, and an example of using the constructor can be seen in the final commit as well as in the first commit with `Sk.misceval.chain`, `Sk.misceval.iterFor` and `Sk.misceval.tryCatch`.

---
### Commits

Only the last 4 commits are relevant
commit -3:  main changes to misceval.Suspension api and related functions
commit -2: places I had to adapt skulpt for this api
commit -1: a compile bug and import bug with import modernization
commit 0: use suspension constructor for `promiseToSuspension`

---
### Final Notes

This pr is a draft since I haven't applied the changes to the `debugger` part of skulpt.
(I haven't used the debugger part of skulpt much)

**I'd be interested in whether there was appetite for this change before I finish it.**

This pr is built on #1092. (it would be difficult to put this on master and merge in #1092).


LeszekSwirski:
> I'd be interested in whether there was appetite for this change before I finish it.

I can't really do a full review, but FWIW I like this approach.

acbart:
I was just browsing while I avoided doing the real work I have to do right now. I wanted to comment that not only does this seem really smart from a speed perspective, but I would anticipate a nice boost on compiled file size. One of the major limiting factors for BlockPy has been the size of the compiled files - because I solved my speed problem by just precompiling the bigger libraries. I seem to recall that the suspension checks were contributing a large chunk of space due to their frequent repetition. So I'm very interested in this!

s-cork:
Update: I've worked on this PR and have implemented this throughout the codebase and checked it works in debug mode where all the suspensions are thrown.

@acbart I had a look at code output size. For fields.py master compiles 5500 lines and this branch compiles 5000 lines (of formatted code). 

In addition to the description in this PR I've changed the call signature for Suspension:
```typescript
new Sk.misceval.Suspension(promise: Promise) 
// should resolve/reject to the value to be returned/raised

new Sk.misceval.Suspension(type_name: string)
// creates an optional suspension whose resume function is () => {};
// hook into this suspension type via suspension handlers passed to asyncToPromise
// see compile.js 'Sk.debug' and friends

new Sk.misceval.Suspension(resume: () => any, child: Suspension)
new Sk.misceval.Suspension(resume: () => any, child: Suspension, locals: any)
// the locals attribute is attached to this suspension and not copied to children
// example usage in Sk.misceval.chain, Sk.misceval.iterFor, Sk.misceval.tryCatch
// and $saveSuspension in compile.js
```
The `data` attribute is still there but its implementation is abstracted away internally.
I've updated the calls internally to reflect this change which cleans up its usage.


I've also added a `ChainBreak` class. Similar to `Break`. This allows for an early return from a long series of chained functions. I've shown a couple of examples of its usage in the codebase.


regarding `Debugger.js` It doesn't look like it currently works on master (the codemirror implementation that used it has since disappeared from the codebase - #913). So i've added a note to say as much. I don't know how much usage it has in the skulpt community? I can pick it up in this PR if it's a problem for merging this branch but it might be best left to someone who has a vested interest in improving it/re-implementing it.

Issue #1263, opened 2021-01-29
api for Sk.builtin.func


s-cork:
With minifiers func names are often lost and so

```javascript
mod.foo = Sk.builtin.func(function foo(self) {
    return Sk.builtin.none.none$;
})
```
loses its name under minification and then the end user printing `foo` gets `<function <native JS>>`
And with modern javascript arrow functions don't have a name property anyway. 

Currently to guarantee a name (or other useful code object parameters) you might do

```javascript
function foo(self) {
    return Sk.builtin.none.none$
}
foo.co_name = new Sk.buitin.str("foo")
mod.foo = new Sk.builtin.func(foo)
```
But this api isn't particularly pleasing to write.

An easy win might be something like this

```javascript
mod.foo = new Sk.builtin.func({
    code: (self) => Sk.builtin.none.none$,
    co_name: new Sk.builtin.str("foo"),
    ...
});

```

This api has the advantage of being a 1 line change in `func.js` that doesn't break existing code.
Being able to encapsulate a code obj entirely inside the call to the func constructor might be a preferred api for some. 

```javascript
Sk.builtin.func = function (code, ...) {
    ...
    this.func_code = code.code || code;
    ...
}
```

Just an idea! It's come up in #1014 and #1261 so thought it worth raising as a discussion point

PR #1265, opened 2021-02-17
Partial docstring support (post #1092)

bennorth:
Give `__doc__` attribute or property to functions, methods, classmethods, classes, and modules.

Simpler replacement for #1261, based on machinery added by #1092.

I'm not fully sure about 5f983e3.  It assumes that the passed-in `code.$doc`, if present, is already a Python object.  Should the code assert that `code.$doc`, if not `undefined`, is either a Python string or `None`?  Should it allow a JavaScript string, and convert to a Python string on construction of the `builtin.func`?  Those might be questions for a separate PR, and are related to #1263.

### DIFF
--- a/src/function.js
+++ b/src/function.js
@@ -34,7 +34,7 @@ Sk.builtin.func = Sk.abstr.buildNativeClass("function", {
 
         this.$name = (code.co_name && code.co_name.v) || code.name || "<native JS>";
         this.$d = Sk.builtin.dict ? new Sk.builtin.dict() : undefined;
-        this.$doc = code.$doc;
+        this.$doc = code.co_docstring || Sk.builtin.none.none$;
         this.$module = (Sk.globals && Sk.globals["__name__"]) || Sk.builtin.none.none$;
         this.$qualname = (code.co_qualname && code.co_qualname.v) || this.$name;
 
--- a/src/function.js
+++ b/src/function.js
@@ -129,6 +129,12 @@ Sk.builtin.func = Sk.abstr.buildNativeClass("function", {
             $get() {
                 return this.$doc;
             },
+            $set(v) {
+                // The value the user is setting __doc__ to can be any Python
+                // object.  If we receive 'undefined' then the user is deleting
+                // __doc__, which is allowed and results in __doc__ being None.
+                this.$doc = v || Sk.builtin.none.none$;
+            },
         },
     },
     proto: {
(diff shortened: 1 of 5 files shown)


src/function.js
```diff
@@ -34,7 +34,7 @@ Sk.builtin.func = Sk.abstr.buildNativeClass("function", {
 
         this.$name = (code.co_name && code.co_name.v) || code.name || "<native JS>";
         this.$d = Sk.builtin.dict ? new Sk.builtin.dict() : undefined;
-        this.$doc = code.$doc;
+        this.$doc = code.$doc || Sk.builtin.none.none$;
```
s-cork, line 37: I don't mind this - I actually prefer what had before - e.g. `co_docstring` just because it feels more inkeeping with other `co_*` objects.

so
```javascript
this.$doc = code.co_docstring || Sk.builtin.none.none$;
```

context: I think this was a place holder and was similar to the `sk_method` api. But the apis are separate so they can deviate here.

src/function.js
```diff
@@ -127,7 +127,7 @@ Sk.builtin.func = Sk.abstr.buildNativeClass("function", {
         },
         __doc__: {
             $get() {
-                return new Sk.builtin.str(this.$doc);
+                return this.$doc;
             },
```
s-cork, line 131: we should add a `$set`

```javascript
$set(v) {
    this.$doc = v; // can be an arbitrary python object
}
```

And perhaps a test.
You should be able to set the `__doc__` in python for any function.
Note that the `__doc__` of a method is readonly

```python
class A: 
    def foo(self): 
        "docstring" 
                                                                
A.foo.__doc__ = 'bar'                                                  
A.foo.__doc__                                                          
#'bar'

a = A()                                                                

a.foo.__doc__ = 'baz' 
# AttributeError: attribute '__doc__' of 'method' objects is not writable
# this is already implemented

A.foo.__doc__ = 1 # can be arbitrary
```

s-cork (approved):
a couple of suggestions but lgtm

src/function.js
bennorth: Good idea.  I've added the code you suggest, and also some tests based on your examples.

src/function.js
bennorth: Happy for it to be `co_docstring`.  I've updated the PR and force-pushed.

Issue #1022, opened 2020-02-09
Outdated NPM package


gda-gp:
Noticed the NPM package is very outdated with the last publish 3 years ago.

https://www.npmjs.com/package/skulpt

Any plans for an updated package?

albertjan:
@daluege are you the same that published this package? If so? Would you like to transfer ownership of this package so we can update it?

gda-gp:
@albertjan could be worth going through the official channels of an abandoned package so you can keep it up to date:

https://www.npmjs.com/policies/disputes

```
    Get the author email with npm owner ls <pkgname>
    Email the author, CC support@npmjs.com
    After 4 weeks, if there's no resolution, we'll address it.
```

Being able to install with npm would simplify the dependency management for everyone.

albertjan:
I'll try

timmartin:
Is there any update on this? I'm getting security errors when I import the package due to old dependencies, it would be nice to sort this out.

albertjan:
@timmartin @gda-gp I've emailed npm and @daluege so please bear with us a little longer.

albertjan:
I've got access to the npm package :tada:

albertjan:
How are you all using the package? Are you importing files from `src/` or are you importing built files from `/dist`?

gda-gp:
I import the built files from /dist.

albertjan:
@skulpt/committers @gda-gp @timmartin what shall we do here? publish 1.0.0 and then publish a new version of master? 

I'd like to publish 1.0.0 because that's also published everywhere else.

timmartin:
I'm not a maintainer, but for my money given there's already a v1.0.0 tag that people may have been using, it would make sense to publish from that tag as v1.0.0 on NPM. I suppose it doesn't greatly matter since the only people using the NPM dist will be stuck on an old version anyway, so they can just as easily upgrade to 1.1.0 or whatever.

albertjan:
That was my plan @timmartin 😄

albertjan:
published 🎉 without dependencies this time 😄 we'll have to work on our deployment/release documentation and process

vthommeret:
This could be user error, but I was previously (5 days ago) able to `npm install skulpt` and import skulpt but now I'm getting `Module not found: Can't resolve 'skulpt'` with `import skulpt from 'skulpt'`.

Based on this issue it seems like I was maybe getting an older version? It was pinned to:

`"skulpt": "^0.11.1",`

[...]

PR #1266, opened 2021-02-18
Update the version number in package.json to 1.2.0

timmartin:
There's a tag for v1.2.0, but this hasn't been updated in the `package.json`. I think we should update this and then push an updated version to the NPM registry.


albertjan:
Hey @timmartin @s-cork should we reintroduce the main.js from the original npm package? So we can do:

```
const Sk = require('skulpt');
Sk.configure()
```

instead of:

```
require('./dist/skulpt.min.js');
require('./dist/skulpt-stdlib.js');

Sk.configure();
```

albertjan:
See this comment: https://github.com/skulpt/skulpt/issues/1022#issuecomment-636905598

s-cork:
@albertjan seems sensible to me

albertjan:
@timmartin can you add the `main` field  in the package.json https://docs.npmjs.com/cli/v7/configuring-npm/package-json#main and add a `main.js` in the root that has:

```
require('./dist/skulpt.min.js');
require('./dist/skulpt-stdlib.js');

module.exports = Sk;
```

timmartin:
Good suggestion, that will fix a major pain point I encountered when I first tried to import the library.

FWIW this PR is showing up as being an outdated branch, because I pulled the branch for this change from the `1.2.0` tag. That seemed like the most sensible thing to do because I'd expect the 1.2.0 release on NPM to include only the changes in the 1.2.0 tag and not further changes on `master` since. But if that makes the merge and deploy untidy I can rebase onto master and we can just live with the inconsistency.

albertjan (approved):

bnmnetp:
If we are going to do a new release wouldn't we want to have #1092 as a part of it??  Maybe we should just do a new tag.

timmartin:
I'm not intending this to be a new release, I'm hoping to get the release we already have made (1.2.0) into the NPM registry so that NPM users can have access to it.

camilstaps:
@bnmnetp @albertjan is anything blocking updating the npm package right now? The outdated version there can be confusing (https://github.com/skulpt/skulpt/issues/1281).

bnmnetp:
I'm happy to merge this.... But I think I missed the memo on what we need to do to update npm.

Would be nice to create a workflow for that piece.

timmartin:
Someone with the necessary credentials needs to run `npm publish`. If we want, this can be done automatically from GitHub actions - I've got [a configuration file here](https://github.com/timmartin/skulpt/blob/scoped_package/.github/workflows/deploy.yml) that does the necessary deploy on any tagged git commit for my private version of Skulpt; we could do the same with mainline if we want.

camilstaps:
@albertjan sorry for troubling you again, but do you think you could run `npm publish` (seeing as you published 1.0.0)? It would be great to have a more recent version in the repository.

albertjan:
Done deployed master as 1.2.0 (soz @s-cork should I republish it as 1.3.0?). Thing's I've learned:

- we need to think about what we want to add to the package and add them to the `files` in package.json
- we need to remove `private: true` from package.json
- we need a command to build minified and non minified versions

What you have to do now:

- build dev and prod builds put them in `dist/`
- remove `dist/` dir from .gitignore
- remove `private: true` from package.json
- publish it

camilstaps:
Thanks!

PR #1268, opened 2021-02-18
fix a bug with richcompare inheritance and reflected number slots

s-cork:
This is a bug that i tried to fix but didn't fix and have now fixed.
It didn't quite make it to #1092 but it was on my local branch

Refelected number slots work differently to reflected compare slots for subclasses - which is the core of the bug

Some code snippets to highlight python intricacies addressed by this pr.
```python
class A:
  def __add__(self, other):
    return 4
  def __radd__(self, other):
    return 5

class B(A): pass

a = A(); b = B()
>>> a + b
4 
# we only take the __radd__ slot if the subclass has actually overridden it
>>> B.__radd__ = lambda self, other: 6
>>> a + b
6 
# B has overridden the reflected add slot so it takes priority
# this is correct on master
>>> del B.__radd__
>>> A.__add__ = lambda self, other: NotImplemented
>>> a + b
5 
# this is the bug fix in this pr
```
The first bug fix is very niche - if the subclass does not override the reflected slot, and the superclass returns `NotImplemented` the reflected slot of the superclass should be called.

RichCompare
```python
class A:
  def __lt__(self, other):
    return True
  def __gt__(self, other):
    return False

class B(A): pass

a = A(); b= B();

>>> a < b
False 
# even though the subclass does not override gt it still takes priority
# this is different behavior to __radd__ slot above
```
reflected richcompare slots behave differently to reflected number slots.
I misunderstood this and this pr fixes that mistake.


meredydd (commented):
Niche, but looks correct.

This looks good, just needs a test! You have test cases in the PR body - stick 'em in a `test_xxx.py` and call it a day.

albertjan (approved):

PR #1270, opened 2021-02-20
Simple workflow to run unit tests

bnmnetp:
This adds a workflow to run our unit tests and would allow us to remove Travis-CI.

Thanks to @timmartin for the test.yml file.

- [x] update README to add a badge.
- [ ] post accept update the protections on master to use this workflow instead of Travis.

The only real change to the repo is that we will add the package-lock.json file, which is recommended practice anyway, to the repo.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,18 @@
+name: unit-test
+on:
+  push:
+    branches: "*"
+    tags-ignore: "v*"
+jobs:
+  unit-test:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


bnmnetp:
It would be a nice addition if there was a way to create a workflow that updated our docs branch with a new version of skulpt.* after each merge to master.

bnmnetp:
I don't think this requires much review.  Any @skulpt/committers have any objections?  If not I'll plan to merge this on Monday.

s-cork (approved):

PR #1273, opened 2021-02-26
fix a typo for max errors

s-cork:
Small change - a couple of typos in the error messages for the max function


bnmnetp (approved):

s-cork:
@bnmnetp do we know why the unit-test hasn't completed here?

rixner:
This seems trivial enough to merge, but I don't want to set the precedent of merging when the checks haven't completed.  But, it seems safe to say that if they haven't completed in 18 days, they are not going to.  Does anyone have any insight into what's going on?

PR #1274, opened 2021-02-27
Minor bugs in stdlib

s-cork:
I ran our stdlib through google closure compiler and here's the bugs it picked up and the resulting changes.

Mostly it was undeclared variables.

The diff looks big for `model.js` but that's because it needed to do an import on the first line since `webgl` was undeclared and should have been imported... that meant everything was indented incorrectly from then on...


meredydd (approved):
Yeah, these are all Good Changes. Lots of little stuff (some obvious bugs, some JS correctness)

PR #1279, opened 2021-03-16
Fix resume from yielding

rixner:
Currently, if the yieldLimit is set and you yield from a deeply nested long running function, state may be lost.

The situation occurs when you resumqe from a yield. At that point, each Python function resumes it's children.  However, if one of those children exceeds the yield limit again before returning, it will return an Sk.yield suspension.  That gets stored in ```$ret``` in ```$wakeFromSuspension```.  The code then enters the main while loop, where it detects that it is past the yield limit.  At that point, it creates a *new* yield suspension (throwing away the suspension currently stored in ```$ret```) and returns that.  This will occur all the way back up the chain.  When you try to resume again, all state is lost.

This pull request simply guarantees that you will run through the while loop body once after waking from a suspension, thereby guaranteeing that ```$ret``` is dealt with correctly.  I feel that this is the smallest perturbation to the code that fixes the issue.

You can see the problem by setting a relatively short ```yieldLimit``` in the Skulpt configuration and running the following program:

```
def one():
    for i in range(1000):
        two()

def two():
    for i in range(100000):
        print(i)

one()
```

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -224,12 +224,13 @@ Compiler.prototype.outputInterruptTest = function () { // Added by RNL
             output += "if ($dateNow - Sk.execStart > Sk.execLimit) {throw new Sk.builtin.TimeLimitError(Sk.timeoutMsg())}";
         }
         if (Sk.yieldLimit !== null && this.u.canSuspend) {
-            output += "if ($dateNow - Sk.lastYield > Sk.yieldLimit) {";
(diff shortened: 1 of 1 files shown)


s-cork (approved):
lgtm - also seems to fix a problem I faced the other day: trying to stop the execution of a function that continues to call itself.

Issue #912, opened 2019-06-19
Fresh build triggers deprecation errors with nmber?


acbart:
I just cloned a new fork of Skulpt and tried to build a fresh copy:

    > git clone .../skulpt
    > cd skulpt/
    > npm install
    > npm run build

This triggered a bunch of build errors, apparently related to the deprecation of the nmber module. I've attached the log.

Is there some other way I should be building the library? For reference, this is on Windows.

[skulpt_nmber_error.log](https://github.com/skulpt/skulpt/files/3306316/skulpt_nmber_error.log)

acbart:
I believe the offending line of code is from time.js: https://github.com/skulpt/skulpt/blob/master/src/lib/time.js#L265

Perhaps uncovered/caused by the work being done over there. Changing the relevant line to:

    mod.accept2dyear = Sk.builtin.assk$(1, Sk.builtin.int_);

seems to have fixed it. It's taking a *long* time to do my build, however, so I will have to let you know. I think I'm past 10 minutes now?

acbart:
Ah, this is happening with build, but not devbuild. I'm guessing we're not there yet with the new version?

albertjan:
this doesn't happen for me what version of node are you on?

albertjan:
Travis runs all builds fine.

bnmnetp:
@acbart -- maybe a windows thing going on.  build, and diet run fine on macOS.  A clean install on a fresh DO droplet works fine as well.  As @albertjan points out our Travis-CI builds use the same build infrastructure and they are all passing.

acbart:
My node was 6.10.1, so super out-of-date. After updating, I'm now on 12.4.0, but it's still causing the same issue.

albertjan:
@acbart can you remove your `node_modules` and reinstall?

acbart:
Still seems to be present after removing node_modules and reinstalling. 

Just to let myself move forward, I removed number.js from the main.js, got rid of its remaining references in time.js, processing.js, and env.js, and deleted the file. That seems to have either caused or uncovered a new, more cryptic error:

    ERROR in Unexpected token u in JSON at position 0

I've attached the log, will continue to investigate.

[skulpt_unicode_u_error.log](https://github.com/skulpt/skulpt/files/3306588/skulpt_unicode_u_error.log)

acbart:
[...]

PR #1285, opened 2021-03-29
Use google closure compiler for the stdlib

s-cork:
This pr aims to use the closure compiler to build the stdlib.
We currently only minify and so we have ended up with a stdlib that is not es5 compliant despite a build option for it.
I've removed the `prebuild` hook since we no longer use it

**In addition:**
I have also added a function call in the `postbuild` hook that will convert constructors in `skulpt.min.js`
```javascript
Sk.builtin.str=Sk.abstr.buildNativeClass("str": {constructor: function() {}})
// becomes
Sk.builtin.str=Sk.abstr.buildNativeClass("str": {constructor: function str()...
```

There's no compilation option to achieve this with google closure compiler (there are some issues related to this on their github page). 
Having the anonymous function can be annoying in production when you're debugging in the console since all the objects are displayed as `constructor`. The latter version means that you can see the types easily in the console e.g `str`.

**Downsides to this pr:**
- doing a full build on the stdlib is (significantly) slower than before

I've added a compile option `npm run compile array.js`
This should help someone fixing compilation issues when developing a single file in the stdlib.

**Possible extensions to this pr:**
- remove babel-minify as a dependency

The relevant commit is the final commit in this pr. The penultimate commit is #1274
It builds on top of #1274 since the changes there were required to pass the compilation in this pr

Would close #1232


rixner:
I'm not in favor of using the closure compiler for this.  I've actually been looking at whether we can/should remove the closure compiler altogether from our build process.  The closure compiler does three things:

1. Type checking
2. Optimization/minification
3. Transpilation to the desired language level

When combined with webpack, this is slow.  And we're not really taking advantage of 2.  We're really just doing minification and not optimization.  The power of the closure compiler is in the advanced optimizations, which we are not using (and are not likely to use, given the constraints it places upon the code for correctness).  This tool chain also introduces problems for some people on some platforms (#912, for example).  Given that we aren't really getting all of the advantages of the closure compiler, I'm not really convinced it is the best tool for the job.

As a proof of concept, I have a version of skulpt that uses [esbuild](https://esbuild.github.io/) for bundling/minification.  Instead of taking 7 seconds on my machine using webpack/closure, esbuild takes 37ms to bundle.  The bundle is 2kb smaller and has equivalent runtime performance.  The issue, though, is that I think esbuild is too immature at this point to use for Skulpt, and it cannot go down to es5 (es6 seems to be the lowest it will go).  And the question is whether we can get an alternate type checking solution working (I will point out, though, that we went years with our type checking completely disabled and no one noticed - not that we should repeat that mistake, but...).

So, I favor looking for another overall solution for the project, rather than letting the closure compiler creep in further.

s-cork:
@rixner I'm all in favour of another solution. I do think that any solution should have an es5 option that does the transpiling and polyfilling. Unless we decide to officially discontinue support for ie11.

Esbuild looks interesting! I've also explored rollup with skulpt which seemed promising. 

I'll leave this pr open until a better option comes along or I get a chance to play with other build options.

rixner:
@s-cork I agree that we need to keep es5 capability unless/until we all agree to drop it.  For the record, I no longer care about it and target es2015 exclusively (I'm not yet ready to go more recent than that, though).

My point was less about using esbuild, though I do find it interesting and am using it quite happily on some other projects, and more about the fact that I think there are likely better solutions for Skulpt than webpack/closure.  If type checking still isn't enabled with closure, that's yet another strike against it in my mind.  Can't we just use babel to transpile both the core and the libraries, then, to whatever language version we want?

rixner:
I've been looking at [swc](https://swc.rs/).  It is also fast and seems like it is largely a drop in replacement for babel.  Perhaps it can be used directly on the stdlib?  It's not a bundler, though, so if we were to use it on the core, we'd still need webpack (or esbuild or rollup or whatever bundler), but could potentially integrate with swc to produce acceptably transpiled output?  swc does seem to have a companion bundler (spack), but the documentation for it is less than sparse...

Perhaps once esbuild hits a 1.0 release, the combination of esbuild/swc would massively speed up the build pipeline?  We don't really need/use the massive flexibility of webpack.

I'm not sure how good the code that swc outputs is, though...

s-cork:
closing this - there's a wider discussion here. And this pr isn't in the right direction.

PR #1292, opened 2021-04-15
Fix bitshift

### LINKED ISSUE #1291: Incorrect right bit shift results
It looks like the right bit shift fails for numbers which don't fit in a 32-bit signed integer.

```python
x = (1<<34)>>33
print("x should be 2 is: %s" % (x))
y = (170 << 28) >> 27
print("y should be 340 is: %s" % (y))
```

Found:
```
x should be 2 is: 0
y should be 340 is: 20
```

Lanzaa:
Fixes #1291 incorrect right bit shift.

### DIFF
--- a/src/int.js
+++ b/src/int.js
@@ -142,10 +142,8 @@ Sk.builtin.int_ = Sk.abstr.buildNativeClass("int", {
         }, JSBI.leftShift),
         nb$rshift: numberShiftSlot(
             (v, w) => {
-                const tmp = v >> w;
-                if (v > 0 && tmp < 0) {
-                    return tmp & (Math.pow(2, 32 - w) - 1);
-                }
+                // Avoid forced signed 32 bit conversion and just divide.
+                const tmp = Math.floor(v / shiftconsts[w+1]);;
                 return tmp;
             },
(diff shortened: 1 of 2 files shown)


src/int.js
```diff
@@ -142,10 +142,8 @@ Sk.builtin.int_ = Sk.abstr.buildNativeClass("int", {
         }, JSBI.leftShift),
         nb$rshift: numberShiftSlot(
             (v, w) => {
-                const tmp = v >> w;
-                if (v > 0 && tmp < 0) {
-                    return tmp & (Math.pow(2, 32 - w) - 1);
-                }
+                // Avoid forced signed 32 bit conversion and just divide.
+                const tmp = Math.floor(v / shiftconsts[w+1]);;
                 return tmp;
```
s-cork, line 147: Suggestion - can we tidy this up - remove the extra semi colon and the `tmp` variable.

s-cork (approved):
lgtm

src/int.js
Lanzaa: Thank you for the quick code review.

I completely missed that extra semi colon.

PR #1301, opened 2021-05-03
functools: implement lru_cache

s-cork:
This extends the functools implementation to include `lru_cache`.

It also adds the `@cache` decorator which was included in python 3.9

I used the cpython tests and adjusted for skulpt.

As part of the pr I also assigned Sk objects into the local scope for functools.


bnmnetp (approved):
Looks like just one minor change to the core. The rest is in the library function so should be safe to merge.

albertjan (approved):

PR #1302, opened 2021-05-05
make next suspension aware

s-cork:
I noticed that next wasn't suspension aware - a small pr to make it so.


meredydd (commented):
Code LGTM!

Approved if  with a proviso that you `grep` the codebase to check whether `Sk.builtin.next_` is being used in non-suspendable contexts

s-cork:
Yeap - we don't call this function in the codebase apart from to pass it to the builtins.

meredydd (approved):
In that case, we're good to go.

PR #1306, opened 2021-05-11
feat: add contextlib managers and decorators

### LINKED ISSUE #1379: return with arguments causes SyntaxError in generator
```python
def demo_generator():
    yield 1
    yield 2
    return 'StopIteration description'
    yield 3


gen = demo_generator()
print(next(gen))
print(next(gen))
print(next(gen))
```

**Expected behaviour:**
1
2
StopIteration: StopIteration description

**Real behaviour:**
<SyntaxError: 'return' with argument inside generator at <unknown>>

s-cork:
Adds `contextlib.contextmanager`, `closing`, `suppress` and `nullcontext`. Generator context managers work both in `with` statements and as reusable decorators, preserving wrapped results and metadata. Entry, body, exit and close-method lookup support Skulpt suspensions.

Failed entry does not invoke exit. Exit exceptions propagate without a second exit call, and handled body exceptions are suppressed. A manager that yields unexpectedly a second time is closed before raising, following newer CPython cleanup behavior rather than Python 3.7's deferred finalization.

## Dependencies and review order

The generator/compiler work formerly in this PR now has separate review units:

1. #1602: lexical class cells, the stack's base.
2. #1603: completion-aware suspension cleanup.
3. #1604: explicit exception causes and context protocol errors.
4. #1605: generator frames and delegation. This PR's direct dependency.
5. #1306: this contextlib module.

This PR now contains only the contextlib layer. The original head is preserved on [s-cork/skulpt:stu-dev/pr1306-original](https://github.com/s-cork/skulpt/tree/stu-dev/pr1306-original).

## Validation and compatibility

Selected CPython contextlib tests include the unchanged reusable-decorator regression. All 35 contextlib tests pass in Skulpt; 34 pass in CPython 3.7.17, with the deliberate immediate-close test excluded from that comparison. JavaScript tests cover real suspensions during entry, body, exit and close-property lookup, including suppression and failed entry.

The complete stack passes development and optimized builds: 562 runtime tests, 465 Python 2 tests and 2,992 Python 3 tests, plus the focused JavaScript suspension tests. Generator additions also pass against CPython 3.7.17.

The remaining contextlib APIs, async managers, traceback/frame introspection and automatic generator finalization are outside this change.

Merge after #1605. Retarget to master and rerun CI before merging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "devbuild": "webpack --mode development",
     "postdevbuild": "node support/build/wrapmodules.js builtin",
     "watch": "webpack --watch --mode development",
-    "test": "node test/generator_suspensions.js && node test/testwrapper.js && node test/testunit.js && node test/testunit.js --python3",
+    "test": "node test/contextlib_suspensions.js && node test/generator_suspensions.js && node test/testwrapper.js && node test/testunit.js && node test/testunit.js --python3",
     "start": "node support/run/runfile.js",
     "profile": "node --prof --no-logfile-per-isolate --log-internal-timer-events support/run/runfile.js -o",
     "postprofile": "node --prof-process v8.log"
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -1144,19 +1144,7 @@ Sk.builtin.next_ = function next_(iter, default_) {
             if (default_) {
                 return default_;
             }
-            const v = iter.gi$ret;
-            // gi$ret is the original value that was thrown by the StopIteration
-            // (or returned by a generator)
-            // We've now lost the original StopIteration
-            // We make the assumption that it was probably called like
-            // raise StopIteration # or  StopIteration()
-            // rather than
-            // raise StopIteration(None)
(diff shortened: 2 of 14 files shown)


alexwenbj:
It seems this PR broke the class inheritance mechanism, the following code failed when I merge this PR in my project:
```
class A:
    def __init__(self):
        self.x = 0

class B(A):
    def __init__(self):
        super().__init__()  #-->here failed with error message:RuntimeError: super(): no arguments
        self.y = 1
    def run(self):
        print(self.x, self.y)

b = B()
b.run()
```

s-cork:
@alexwenbj - I don't think it's this branch alone. If you just run that code in this branch there's no problems.
That code is also well tested in the unittests `test_super.py`.

I guess the combination of merges might have caused a conflict somewhere. (I don't have a branch with all of these prs combined anymore). As they get picked off I'll do the necessary rebasing and merging for each branch to avoid that.

alexwenbj:
@s-cork Yeah,maybe. So  I listed the PRs in order I merged for your information.

PR #1310, opened 2021-05-12
Add Github actions config to deploy to NPM on version tag

timmartin:
This sets up GitHub actions to deploy to NPM automatically on a tag prefixed with "v".

This isn't complete in itself; it will require an NPM auth token to be set up and deployed as a GitHub secret called `NPM_TOKEN`.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,20 @@
+name: deploy
+on:
+  push:
+    tags: v*
+jobs:
+  publish-npm:
+    runs-on: ubuntu-latest
+    environment: CI
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+          registry-url: https://registry.npmjs.org/
+      - run: npm ci
+      - run: npm run build
+      - run: npm test
+      - run: npm publish
+        env:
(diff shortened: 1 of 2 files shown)


bnmnetp (approved):

albertjan:
@timmartin do we not also have to update the version in package.json?

albertjan:
It would be really cool if we could automatically push the latest master build to npm as a beta package.

timmartin:
If you want to use NPM tooling, you can create the version tag in Git and update `package.json` in one go with `npm version patch` / `npm version minor` / etc. We can hook additional behaviour into this to do other things we want using the `preversion`, `version` and `postversion` NPM hooks. However, from the discussion on the issue it seemed like we wanted to skip NPM tooling entirely and have non-NPM tooling update the `package.json`, in which case you're on your own.

I don't know about the tooling for pushing every build of `master` to the NPM registry. I think it's required that every new version has a distinct version number, in which case we'd need to find a way to generate that version number (possibly we can use the GitHub actions build number, which should be monotonic). This will mean that the patch number of the version number is outside our control, which doesn't seem desirable. I haven't seen any other packages that do this, so it might not be widely supported / easy to set up.

bnmnetp:
Hi @albertjan - I would like to merge this and start to publish more frequently.  It looks like you are the owner of skulpt on the npm package archive.  We would need to set up an access token in the secrets for this repo.  Can you help with that?

PR #1313, opened 2021-05-22
implement __annotations__ for functions

s-cork:
We already parse annotations from the function definition. 
But they aren't stored in the `__annotations__` descriptor.

This pr implements that mechanism.

The code in compile.js and symtable.js is adapted from cpython's implementation and is fairly close.

I've added a few tests and included the mechanism for posonly args (although we don't yet parse those)

The lack of `__annotations__` was a reason I couldn't include `single_dispatch` in the functools implementation.


meredydd (approved):
Well, this is exciting. Approved!

meredydd:
This is small enough for a unilateral merge

PR #1320, opened 2021-06-09
tweak exec which fails when importing modules in repl

s-cork:
After #1212 I encountered a problem when trying to import a module in the repl. 
The code in this pr fixes it
I couldn't find a failing test for the unittest but in the repl importing any module failed.

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -798,7 +798,10 @@ Sk.builtin.exec = function (code, globals, locals) {
     globals = globals || tmp;
     return Sk.misceval.chain(
         code,
-        (co) => eval(co.code)(globals, locals),
+        (co) => {
+            co.code += "\n" + co.funcname + ";";
+            return eval(co.code)(globals, locals);
+        },
         (new_locals) => {
             Sk.globals = tmp;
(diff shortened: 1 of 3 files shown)


albertjan (approved):

src/builtin.js
```diff
@@ -798,7 +798,10 @@ Sk.builtin.exec = function (code, globals, locals) {
     globals = globals || tmp;
     return Sk.misceval.chain(
         code,
-        (co) => eval(co.code)(globals, locals),
+        (co) => {
+            co.code += "\n" + co.funcname + ";";
```
albertjan, line 802: what does this do? can you add a comment?

s-cork: This ensures that the main function scope within the compiled code string is returned from eval. 

What do you think about moving this line to the end of the Sk.compile instead?

The line also appears in import.js and so we can then only append this line in import.js if we're importing a $builtinmodule. 

That might cleaner?

albertjan: yeah, makes sense to me keep all the code producey bits in there

PR #1321, opened 2021-06-09
make keyword a js module

s-cork:
Only so that it doesn't need to be compiled. It's used in collections.

I also change the constructor for `frozenset` to allow a python iterable to be used as the argument. (the change matches the constructor for `set` which also allows an iterable).

The keywords are copied from python (3.9).


albertjan (approved):

s-cork:
Small change - merging

PR #1322, opened 2021-06-09
Make ordered dict use dict since it's now ordered

s-cork:
Our current implementation of ordered dict can break if you try to do something like

```python
from collections import OrderedDict

>>> a = OrderedDict()
>>> a.set_default('foo', 'bar')
>>> a
# error dict changed size during iteration
```

That's because not all the methods are overridden for ordered dict 
so things in `ordered_keys` are not set for certain methods
We could extend the ordered dict to override all the necessary methods (basically all of them).
or we could just remove `ordered_keys` and let ordered dict use dict (which is now ordered).

The benefits of this approach is reducing the code size.

We then only need to override the 3 methods that are different `$r`, `tp$richcompare`, `move_to_end`.


albertjan (approved):
looks good to me makes sense not maintaining two dicts

test/unit3/test_ordereddict.py
```diff
@@ -85,8 +118,8 @@ def test_delitem(self):
         od = OrderedDict(pairs)
         del od['a']
         self.assertNotIn('a', od)
-        #with self.assertRaises(KeyError):
-        #    del od['a']
+        with self.assertRaises(KeyError):
+           del od['a']
```
albertjan, line 122: whoa when did you fix the `del` statement?

bnmnetp (approved):

PR #1330, opened 2021-06-25
reduce parsenumber complexity

s-cork:
When the skulpt tokenizer produces a `NUMBER` token we know it's positive and:
- a valid int literal
- or float
- or purely imaginary complex number

We don't need to do quite as much work to turn it into the correct python type.

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -2677,94 +2677,39 @@ function parsestrplus (c, n) {
     }
 }
 
-const invalidSyntax = /_[eE]|[eE]_|\._|j_/;
-const invalidDecimalLiteral = /_\.|[+-]_|^0_\D|_j/;
-const validUnderscores = /_(?=[^_])/g;
-function parsenumber (c, s, lineno) {
-    var neg;
-    var val;
-    var tmp;
-    var end = s.charAt(s.length - 1);
-    
-    if (s.indexOf("_") !== -1) {
-        if (invalidSyntax.test(s)) {
-            throw new Sk.builtin.SyntaxError("invalid syntax", c.c_filename, lineno);
-        }
-    
-        if (invalidDecimalLiteral.test(s)) {
-            throw new Sk.builtin.SyntaxError("invalid decimal literal", c.c_filename, lineno);
-        }
-        
-        s = s.replace(validUnderscores, "");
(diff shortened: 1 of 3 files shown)


src/ast.js
```diff
@@ -2677,95 +2677,41 @@ function parsestrplus (c, n) {
...
-const invalidDecimalLiteral = /_\.|[+-]_|^0_\D|_j/;
-const validUnderscores = /_(?=[^_])/g;
-function parsenumber (c, s, lineno) {
-    var neg;
-    var val;
-    var tmp;
-    var end = s.charAt(s.length - 1);
-    
-    if (s.indexOf("_") !== -1) {
-        if (invalidSyntax.test(s)) {
```
albertjan, line 2690: so we can't get these syntax errors because our tokeniser is more specific then cpythons?

s-cork: Yeah - so because we're using the `tokenize.py` rather than `tokenizer.c` the tokens are generated in a different way and these SyntaxErrors were never thrown because our tokens were already valid numeric literals - because the regex expressions that generate the tokens ensure they are already valid...

E.g. (in python)
```python
>>> from tokenize import generate_tokens
>>> tok = generate_tokens(readline("123__456"))
>>> next(tok)
TokenInfo(type=2 (NUMBER), string='123', start=(1, 0), end=(1, 3), line='123__456')
>>> next(tok)
TokenInfo(type=1 (NAME), string='__456', start=(1, 3), end=(1, 8), line='123__456')
```

So we're not catching the bad token at the tokenize stage.
We only get a syntax error at the parsing stage because you can't follow a `NUMBER` with a `NAME`.

Whereas - `tokenizer.c` does something different to determine a number token and will catch the bad numeric literal at that stage.


There wasn't anything obvious I could see to get these syntax errors to work. parse_number doesn't know about the next token - and I don't think it should.
I guess the Regex in `tokenize.js` could be adjusted to encompass bad numeric patterns.
But that would mean deviating away from `tokenize.py`.

albertjan: I figured as much 😄 thanks.

albertjan (approved):

meredydd (approved):

PR #1337, opened 2021-07-09
Fix bug in function astFromParse in ast.js

pingskills:


s-cork (approved):
Looks good to me - equivalent line in cpython
https://github.com/python/cpython/blob/3.7/Python/ast.c#L794

bnmnetp (approved):

PR #1338, opened 2021-07-13
Adjust the round function in datetime.js

s-cork:
Because in python2 rounding a float returns a float - it becomes a problem when we assert that certain values are python integers if we happen to be in python2 mode.

`timedelta(minutes=0)` will fail in python2 mode...


albertjan (approved):

s-cork:
small change merging

PR #1339, opened 2021-07-16
implement json

s-cork:
1 commit ahead of #1191.
(Only look at the final commit in this pr to see the implementation)
I wanted to use the ffi branch for this because it allows hook functions into `toJs`
The hook functions are used in this implementation of `dumps`.

Based largely on the python json implementation.
This PR implements `dumps`, `loads` and `JSONDecodeError` with tests from CPython.

Some differences to python
`dump`, `load`, `JSONDecoder` and `JSONEncoder` are not implemented.
`strict` is ignored when passed to `loads`.
`ensure_ascii` is ignored when passed to `dumps`.
using `cls` or other keyword arguments throws.


s-cork:
@bnmnetp this PR should be good for a review and merge

bnmnetp (approved):

PR #1343, opened 2021-07-21
Errors and imports

### LINKED ISSUE #1341: ImportError for imports
I wish I had been here seven years ago to say:

Ahhh! Don't change the raised error type from `ImportError` to anything else! That will break literally any code which gracefully degrades in the absence of a module! (Unless you subclass `ImportError` and raise that subclass - that's harmless.)

The whole point of having import errors be typed as import errors is so that code can do stuff like this:

```python
try:
    from threading import local
     _threadlocal_state = local()
except ImportError:
    # degrade to works in single-threaded environments
    class _Object(object):
        pass
    _threadlocal_state = _Object()

...
```
or
```python
try:
    from reprlib import recursive_repr
    _recursive_repr_if_available = recursive_repr()
except ImportError:
    def _recursive_repr_if_available(decorated_function):
        return decorated_function

...

class SomeClass(object):
    ...
    @recursive_repr_if_available
    def __repr__(self):
        ...
```

If the error type is changed, all code already written to handle missing modules goes from working to unusably broken.

Because of this change, maximally portable code needs to catch two possible errors on every import instead of just one, if it wants to run without modification on Skulpt. Though I imagine almost every self-respecting developer solves that problem by deciding "my module just isn't going to work on Skulpt then, oh well" - I'm the only person I know neurotic enough to feel compelled to make sure users of my stuff can run it on Python implementations like Skulpt without modification.

Anyway, if it were me, I'd keep the very helpful error messages as-is, but change the error types back to `ImportError`. If someone actually has a use-case for _programmatically_ responding to these import errors differently than other ones, I'd `class ModuleNotImplemented(ImportError): pass`. But since by now some Skulpt code out there might have been written to rely on catching the current error type, I'd `class ModuleNotImplemented(ImportError, NotImplementedError): pass` - that way you restore compatibility with the standard and retain compatibility with how Skulpt's been doing it for seven years.

_Originally posted by @mentalisttraceur in https://github.com/skulpt/skulpt/issues/276#issuecomment-850090989_

### LINKED ISSUE #1342: Multiple inheritance for builtin classes doesn't work when it should
Most of the time it shouldn't work

e.g. you can't just inherit from `str` and `dict`

```python
>>> class A(str, dict): pass
TypeError: multiple bases have instance lay-out conflict
```

But at the moment we tar all builtin classes with the same brush. Even when it should work.
Specifically for Errors classes
```python
>>> class A(TypeError, ValueError): pass
TypeError: multiple bases have instance lay-out conflict
# This should be allowed - it is in cpython
````

We'd need to fix this before addressing #1341

s-cork:
This pr does a few things
1. changes the flag `sk$acceptable_as_base_class` to `sk$unacceptableBase` 
2. Allows Errors to support multiple inheritance close #1342 
3. Uses the cpython approach to construct python Error classes
4. Replaces all `NotImplementedError`s with `NotImplementedImportError` for our stdlib stubs close #1341 
5. Don't cache modules in sys.modules that raise exceptions

### 1. `sk$acceptable_as_base_class` -> `sk$unacceptableBase`
it seems better to have a positive flag here based on the context it's checked
It's a relatively new flag and internal so this shouldn't be breaking
See `type.js` for when this is checked.

### 2. Allows Errors to be support multiple inheritance
adds a flag `sk$solidBase` which all builtin native classes are assigned by default
layout conflicts in the `type.js` `best_base` function are checked against this flag.
This can be overridden when building a native class. Which is what the Exception classes now do.

### 3. Uses the cpython approach to construct python Error classes
See cpython's exceptions.c
They have 3 subclass creator functions `ComplexExtendsException`, `SimpleExtendsException` `MiddlingExtendsException`.
We take a similar approach and construct the Exceptions this way.
The exceptions constructed by `simpleExtends` can be subclassed with other similar exceptions.
The major change in this pr is `errors.js`.

### 4. Replaces all `NotImplementedError`s with `NotImplementedImportError` for our stdlib stubs
We might break code if we change the Error type. But using multiple inheritance works well.
(as suggested in this https://github.com/skulpt/skulpt/pull/276#issuecomment-850090989)
It's a very common pattern to catch an `ImportError` and use a fallback so we should support that.

### 5. Don't cache modules in sysmodules that raise exceptions
This is consistent with cpython's importlib
At the moment in skulpt you can import a not implemented module but only get the error the first time you try to import.
All subsequent times you get the cached import
This could be bad if you have fallback imports
The next time you try to import the module you'd get the cached module rather than the fallback 😬


s-cork:
For any @skulpt/committers it's probably best to look through commits individually. Since looking at all files changed gives you too many automatic changes across the stdlib which hides the detail.

meredydd (approved):
Solid stuff that solves a real problem.
I confess to being a little woolly on the multiple-inheritance stuff - it's not recently loaded into cache for me - but this is a strict expansion of capabilities and only affects Exceptions, which were already extremely hacked up, so I see no reason to put the brakes on this. Ship it!

PR #1347, opened 2021-08-05
throw python syntax errors not javascript ones

s-cork:
a couple of places we throw a js SyntaxError when we should throw a python SyntaxError.


albertjan (approved):

s-cork:
Small change - merging

PR #1348, opened 2021-08-10
Fix datetime regression - can no longer be copied

s-cork:
Moving datetime.js to a native class meant it could no longer be copied or deepcopied.

This pr fixes that.

I adjusted copy.py to only complain about special methods if we are using the custom copy method to do so.
I adjusted bytes.js to support a `valueOf` js method, which returns it's `.v` property.

I implemented the `__reduce__` methods based on python's datetime.py module which takes an identical approach.
The `tp$new` methods are adjusted inline with the equivalent `__new__` methods in datetime.py.

For testing I was able to uncomment one set of tests.
I added a test at the end of `test_datetime.py` since copying was primarily tested via pickling...
And skulpt doesn't do pickling.


meredydd (approved):
Solid improvement! I'm going to say this is enough to insta-merge.

PR #1349, opened 2021-08-10
Ellipsis

s-cork:
I didn't realise that Ellipsis were part of `__builtins__`
This pr puts `Ellipsis` in the skulpt builtindict.

It also changes the name of `Sk.builtin.ellipsis` -> `Sk.builtin.Ellipsis`.
I can't see this as breaking since it was only introduced 4 weeks ago.
Reason: In python the constant is `Ellipsis` and `type(Ellipsis)` is `ellipsis`.


albertjan (approved):

s-cork:
small change

PR #1356, opened 2021-09-05
fix __contains__ can't suspend

s-cork:
the compile code assumed `Sk.builtin.bool` could suspend but i don't think that's ever been the case.

This pr fixes this by only calling `Sk.builtin.bool` after the call to `__contains__` has completed any suspensions.

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -623,9 +623,9 @@ Compiler.prototype.ccompare = function (e) {
 
     for (i = 0; i < n; ++i) {
         rhs = this.vexpr(e.comparators[i]);
-        out("$ret = Sk.builtin.bool(Sk.misceval.richCompareBool(", cur, ",", rhs, ",'", e.ops[i].prototype._astname, "', true));");
+        out("$ret = Sk.misceval.richCompareBool(", cur, ",", rhs, ",'", e.ops[i].prototype._astname, "', true);");
         this._checkSuspension(e);
-        out(fres, "=$ret;");
+        out(fres, "=Sk.builtin.bool($ret);");
         this._jumpfalse("$ret", done);
         cur = rhs;
     }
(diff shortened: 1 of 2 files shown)


src/compile.js
```diff
@@ -623,9 +623,9 @@ Compiler.prototype.ccompare = function (e) {
 
     for (i = 0; i < n; ++i) {
         rhs = this.vexpr(e.comparators[i]);
-        out("$ret = Sk.builtin.bool(Sk.misceval.richCompareBool(", cur, ",", rhs, ",'", e.ops[i].prototype._astname, "', true));");
+        out("$ret = Sk.misceval.richCompareBool(", cur, ",", rhs, ",'", e.ops[i].prototype._astname, "', true);");
         this._checkSuspension(e);
-        out(fres, "=$ret;");
+        out(fres, "=Sk.builtin.bool($ret);");
         this._jumpfalse("$ret", done);
```
meredydd, line 629: `$ret` is now a different data type - this line should probably update to match

meredydd (changes requested):
One nit, then looks good to go

src/compile.js
s-cork: `_jumpfalse` takes a javascript bool or python object.
So the code here works fine. I can add a comment or replace line 629 with
`this._jumpfalse(fres, done);`

meredydd: Omg it can take both?! Gross.

If you can confirm that we have tests for multi-comparison expressions (I think that's just `a < b< c`) that would break if this assumption were (or became) incorrect, I'm happy to merge.

s-cork: test/run/t268.py and t273.py test this as well as `test/unit3/test_types.py (TestTypes.test_comparisons)` 👍 

We should really add all the tests from cpython's `test_grammar.py` which also includes these sorts of tests.
I'll make a separate pr for that at some point.

meredydd (approved):

PR #1357, opened 2021-09-05
fix float formatting when the width is is specified for negative values

s-cork:
The following code:
```python
print(f'{-0.003:6.3f}')
print(f'{0.003:6.3f}')
print(f'{-1.234:6.3f}')
print(f'{1.234:6.3f}')
```
Produces the following output:
```python
0.003
 0.003
1.234
 1.234
```
Instead of.
```python
-0.003
 0.003
-1.234
 1.234
```

Test taken from cpython.
I took the whole test class and commented out the tests that are beyond the scope of this pr.
This bug is covered by `FormatTestCase.test_issue35560`.

### DIFF
--- a/src/formatting.js
+++ b/src/formatting.js
@@ -28,7 +28,7 @@ let handleWidth = function (m, r, prefix, isNumber) {
         let nFill = fieldWidth - (r.length + (prefix ? prefix.length : 0));
 
         if (nFill <= 0) {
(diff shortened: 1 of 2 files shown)


meredydd (approved):
Looks good! (And ten points for getting us on the CPython tests so we can clean up the rest.)

PR #1358, opened 2021-09-05
time.time() should be increasing

s-cork:
it's possible to get negative timestamp differences in skulpt
But all timestamps should really be monotonically increasing 
(unless the system time was changed between calls)

```python
from time import time
t = time()
for _ in range(1000):
  if time() < t:
    print('back in time!')
  t = time()
```

This pr fixes that with an implementation that entirely depends on the browser `performance` api.
(falling back to the `Date.now()` api if the `performance` api is not available)
The previous implementation combined the two apis which led to negative timings since the fractional offset was not accurate with respect to `Date.now()`.

I've added comments for browser compatibility links to mdn.
I haven't added additional tests since node doesn't support `performance` without a `require('perf_hooks')`.

---

I also replace `this` with `Sk.global` in a few places which I think is safer.

### DIFF
--- a/src/lib/time.js
+++ b/src/lib/time.js
@@ -10,6 +10,23 @@
 var $builtinmodule = function (name) {
     var mod = {};
 
+    var getTimeStamp = (function () {
+        const performance = Sk.global.performance;
+        // all modern browsers support performance (IE10+) but node does not without the use of require('perf_hooks');
+        const timeOrigin =
+            performance && (performance.timeOrigin || (performance.timing && performance.timing.navigationStart));
+        // timeOrigin is experimental but supported by all modern browsers except safari - https://developer.mozilla.org/en-US/docs/Web/API/Performance/timeOrigin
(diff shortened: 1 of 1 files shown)


meredydd:
Er...I'm not sure about this! On a few minutes' reading, it looks as though `performance` is a monotonic-clock API for performance measurements, not a date/time API, which means that using it like one invites weird effects (such as _not_ updating on a system clock change, or [weird readings](https://github.com/open-telemetry/opentelemetry-js/issues/852) over suspend/resume, off the top of my head).

s-cork:
The current skulpt implementation of `time.time()` already uses `performance`...
But it combines `performance` and `Date` in a way that causes negative time intervals.

I'm happy to just remove `performance` from `time.time()` and go with the `Date` api... But then it's a different pr.
thoughts @meredydd ?

meredydd:
I think relying entirely on `Date` makes sense for a datetime module. Let's do that instead.

meredydd (changes requested):
Per @s-cork, let's switch to using `Date` throughout

meredydd (approved):
Looking good now!

PR #1362, opened 2021-09-09
include test_long and adjust accordingly

### LINKED ISSUE #1325: % operator incorrect for large negative integers
e.g.

```
>>> x = -int('1' * 100)
>>> x%4
# skulpt says -3
1
```

This is because of the difference between the % operator in javascript vs python
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder

It shouldn't be too difficult to fix.
Inside the scope of a related pr might be to add `test_long.py` from cpython (we don't currently include it)
Adding it will likely expose some other int related bugs 😬

s-cork:
This pr implements `test_long.py` and updates `test_pow.py`.

Adding the tests exposed some problems (mostly) with BigInts and I adjust our implementations accordingly.
close #1325

---

The **[final commit](https://github.com/skulpt/skulpt/commit/993e19bdfe1511b7191c128bbfc25ee5baf95112)** in this pr is the one that's worth looking at, since that's where the changes are made. The bulk of the changes are in int.js.
- commit 1 - minor formatting and a small tweak to the int slots (minor cosmetic changes)
- commit 2 - copy and paste cpython tests into our test suite.
- commit 3 - update JSBI (our BigInt polyfill) to the latest version, which includes bug fixes
- commit 4 - adjust the cpython tests for skulpt and fix the issues that arose.

---

**Some of the bug fixes:**
- pow didn't work for negative bigints
- round wasn't correctly implemented for bigints
- division for bigints things like `9007199254740992/2`
- comparisons with floats and bigints
- shift slots for integers larger than `2**31` might give the js value rather than the python value.
- domain error not thrown for math log functions when `x = 0`.

---

Note - we have two versions of the `JSBI.BigInt` polyfill. We use native `BigInt`s under the hood if they are available and if we're in an older browser we use the `JSBI` library as is.
To ensure that the changes work in both cases I also ran tests locally after deleting `window.BigInt`.


src/formatting.js
```diff
@@ -82,15 +85,20 @@ let handleInteger = function(m, n, base){
         }
     }
 
-    if (m[FMT.CONVERSION_TYPE] === "X") {
-        r = r.toUpperCase();
+    const converstionType = m[FMT.CONVERSION_TYPE];
```
meredydd, line 88: speling

meredydd (approved):
Sensible changes, with CPython tests! What's not to like?
(apart from one spelling error)

PR #1363, opened 2021-09-13
Fix time tests for Australian Time Zones

pingskills:


s-cork:
Could you add a description about what this solves?

The change in `_strptime` seems right and a mistake in the original code.
See cpython: https://github.com/python/cpython/blob/main/Lib/_strptime.py#L152-L167

But I'd rather not change the `test_strptime.py` if we can avoid it - it's directly taken from cpython:
https://github.com/python/cpython/blob/main/Lib/test/test_strptime.py#L82-L102
https://github.com/python/cpython/blob/main/Lib/test/test_strptime.py#L258-L265

test/unit3/test_strptime.py
```diff
@@ -83,8 +83,8 @@ def test_date_time(self):
         #  output.
         magic_date = (1999, 3, 17, 22, 44, 55, 2, 76, 0)
         strftime_output = time.strftime("%c", magic_date)
```
pingskills, line 85: On my local machine strftime_output is set to:
`Wed 17 Mar 1999 10:44:55 PM Australian Eastern Daylight Time`

test/unit3/test_strptime.py
```diff
@@ -83,8 +83,8 @@ def test_date_time(self):
...
-        self.assertEqual(time.strftime(self.LT_ins.LC_date_time, magic_date),
```
pingskills, line 86: And time.strftime(self.LT_ins.LC_date_time, magic_date) is set to:
`Wed 17 Mar 1999 10:44:55 PM australian eastern daylight time`

hence the test fails

test/unit3/test_time.py
```diff
@@ -113,8 +113,11 @@ def test_strftime(self):
         timestamp_to_test = (
             days_to_20020203 * 24 * 60 * 60  # Unix timestamps ignore leap seconds
             + seconds_within_day
-            + time.timezone
         )
+        if time.localtime(timestamp_to_test).tm_isdst:
```
pingskills, line 117: In Australia, the 3rd of February is in Daylight Savings Time and hence this test fails by 1 hour without this fix.

pingskills:
Sure, I'll update the PR to only update _strptime. 

I added some comments to the changed files to show where and why the tests are failing when I run them on my local machine.

test/unit3/test_time.py
s-cork: We should definitely fix this. I just think we should fix the root cause rather than changing the tests.

I have feeling that the problem is that `time.daylight` is returning the incorrect result for the southern hemisphere.

I think if you can fix that then the rest of the tests will pass. Let us know if you need help chasing that bug down.
You might find the pypy/cpython implementation helpful.

https://github.com/mozillazg/pypy/blob/e765b5194c0962ba921a501f59c3011feb7fcf68/pypy/module/time/interp_time.py#L282-L292
https://github.com/python/cpython/blob/main/Modules/timemodule.c#L1781-L1796

test/unit3/test_strptime.py
pingskills: These tests seem to be failing only because of the case, not because of any issues with daylight savings.

test/unit3/test_time.py
pingskills: I think time.daylight is correct for the southern hemisphere. If I `print(time.daylight) `it displays 0 which is correct for today in Australia. The problem is the way the test is written. It assumes that the 3rd of February 2002 is not daylight savings and hence always adds time.timezone (non DST offset). The test needs to add time.timezone in the northern hemisphere and time.altzone in the southern. That's why I updated the test.

Here is how skulpt currently determines if it is daylight savings:

https://github.com/skulpt/skulpt/blob/master/src/lib/time.js#L88-L102

This works because the offset is positive for the southern hemisphere and negative for the northern hemisphere.

s-cork: Really - if I run python on my laptop and switch my timezone to Sydney i get a value of 1 for `time.daylight`

pingskills: Ahh, so it's whether or not we have daylight! I'll fix the time.daylight then. When I print time.daylight in Skulpt I get 0, however on Python I get 1 just like you.

pingskills: All tests now pass for me except this one. :smiley: 

I believe it's because this specific test relies on time.timezone which assumes the chosen date is not currently in daylight savings time, however because we are using localtime(), for me it is.

s-cork: Great - definitely include a fix for this test.

It's not directly from cpython, from what I can gather. And has failed for sometime until a recent pr fixed it (for most people).

test/unit3/test_time.py
```diff
@@ -113,10 +113,9 @@ def test_strftime(self):
         timestamp_to_test = (
             days_to_20020203 * 24 * 60 * 60  # Unix timestamps ignore leap seconds
             + seconds_within_day
-            + time.timezone
         )
         self.assertEqual(
-            time.strftime("%b %d %Y %H:%M:%S", time.localtime(timestamp_to_test)),
+            time.strftime("%b %d %Y %H:%M:%S", time.gmtime(timestamp_to_test)),
```
pingskills, line 118: Is there any requirement to use localtime() for this test? If not then using gmtime() instead should resolve the issues.

test/unit3/test_time.py
pingskills: I've updated the test to use gmtime() instead of localtime() so we don't need to worry about any offsets.

s-cork (approved):
LGTM - small change - merging.

PR #1365, opened 2021-09-18
fix: #931

### LINKED ISSUE #931: With new Skulpt,turtle register_shape() will raise error:wrapperFn() takes 0 positional arguments but 1 was  given
With this new version of Skulpt,the turtle register_shape() function will raise an error:`wrapperFn() takes 0 positional arguments but 1 was  given`.It is OK before the July version.The register_shape() invocation is as below:
`register_shape("football.png")`

viewerw:
fix #931

### DIFF
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -1321,7 +1321,8 @@ function generateTurtleModule(_target) {
             }
         };
         proto.$register_shape.minArgs = 1;
-
+        proto.$register_shape.co_varnames = ["name", "points"];
+        
         proto.$getshapes = function() {
             return Object.keys(SHAPES);
         };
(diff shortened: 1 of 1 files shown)


src/lib/turtle.js
```diff
@@ -1321,7 +1321,8 @@ function generateTurtleModule(_target) {
             }
         };
         proto.$register_shape.minArgs = 1;
-
+        proto.$register_shape.co_varnames = ["name", "points"];
+
```
s-cork, line 1325: looking at the definition in cpython - isn't the second argument called shape?

viewerw: maxArgs is not necessary, the defaultValue of maxArgs is the func's length

s-cork (approved):

PR #1367, opened 2021-09-19
feat: add lcm function to math module

### LINKED ISSUE #1366: [feature] Support for arbitrary number of arguments for builtin functions
The question is how it could be implemented and if it is even possible?

I know very little of this library internals but from a quick look it seems that it would be enough to remove hardcoded arguments and Sk.builtin.pyCheckArgsLen from lib js files.

Reason I'm asking is that I wan't to add some math functions from latest versions and they are documented with arbitrary amount of arguments (ex.: lcm).

BioMedTech:
docs: https://docs.python.org/3/library/math.html#math.lcm

closes #1366

### DIFF
--- a/src/lib/math.js
+++ b/src/lib/math.js
@@ -334,30 +334,30 @@ const $builtinmodule = function (name) {
 
         let i;
         for (i = 1; i < nargs; ++i) {
-          const argument = args[i];
+            const argument = args[i];
 
-          Sk.builtin.pyCheckType(
-            "*integers",
-            "integer",
-            Sk.builtin.checkInt(argument)
-          );
-
-          // If any of the arguments is zero, then the returned value is 0
-          if (argument === 0) return 0;
-
-          let _argument = Sk.builtin.asnum$(argument);
-
-          if (typeof result === "number" && typeof _argument === "number") {
-            result = (result / gcd(result, _argument)) * _argument;
-          } else {
-            result = JSBI.BigInt(result);
-            _argument = JSBI.BigInt(_argument);
-
-            result = JSBI.multiply(
-              JSBI.divide(result, gcd(result, _argument)),
-              _argument
+            Sk.builtin.pyCheckType(
+                "*integers",
+                "integer",
+                Sk.builtin.checkInt(argument)
             );
-          }
+
+            // If any of the arguments is zero, then the returned value is 0
+            if (argument === 0) return 0;
+
+            let _argument = Sk.builtin.asnum$(argument);
+
+            if (typeof result === "number" && typeof _argument === "number") {
+                result = (result / gcd(result, _argument)) * _argument;
+            } else {
+                result = JSBI.BigInt(result);
+                _argument = JSBI.BigInt(_argument);
+
+                result = JSBI.multiply(
+                    JSBI.divide(result, gcd(result, _argument)),
+                    _argument
+                );
+            }
         }
 
         return new Sk.builtin.int_(result);
--- a/src/lib/math.js
+++ b/src/lib/math.js
@@ -333,7 +333,7 @@ const $builtinmodule = function (name) {
     function lcm(...args) {
         function abs(n) {
             if (typeof n === "number") {
-                return Math.abs(n);
+                return new Sk.builtin.int_(Math.abs(n));
             }
 
             return JSBI.lessThan(n, JSBI.__ZERO)
--- a/src/lib/math.js
+++ b/src/lib/math.js
@@ -346,21 +346,26 @@ const $builtinmodule = function (name) {
         // lcm() without arguments returns 1
         if (nargs === 0) return new Sk.builtin.int_(1);
 
-        let result = Sk.misceval.asIndexOrThrow(args[0]);
+        // Test & convert all arguments
+        let i;
+        for (i = 0; i < nargs; ++i) {
+            args[i] = Sk.misceval.asIndexOrThrow(args[i]);
+        }
 
+        let result = args[0];
         if (nargs === 1) {
             return abs(result);
         }
 
-        let i;
+        let arg;
         for (i = 1; i < nargs; ++i) {
-            let argument = Sk.misceval.asIndexOrThrow(args[i]);
+            arg = args[i];
 
             // If any of the arguments is zero, then the returned value is 0
-            if (argument === 0) return new Sk.builtin.int_(0);
+            if (arg === 0) return new Sk.builtin.int_(0);
 
-            if (typeof result === "number" && typeof argument === "number") {
-                let tmp = (result / _gcd_internal(result, argument)) * argument;
+            if (typeof result === "number" && typeof arg === "number") {
+                let tmp = (result / _gcd_internal(result, arg)) * arg;
                 tmp = Math.abs(tmp);
 
                 // check the size - if we're too big reset the result and then fall through
@@ -371,10 +376,10 @@ const $builtinmodule = function (name) {
 
             // allow fall through - if result gets too big we'll need to redo the calculation with BigInts
             if (typeof result !== "number") {
-                argument = JSBI.BigInt(argument);
+                arg = JSBI.BigInt(arg);
                 result = JSBI.multiply(
-                    JSBI.divide(result, _gcd_internal(result, argument)),
-                    argument
+                    JSBI.divide(result, _gcd_internal(result, arg)),
+                    arg
                 );
             }
         }
--- a/src/lib/math.js
+++ b/src/lib/math.js
@@ -314,13 +314,13 @@ const $builtinmodule = function (name) {
         throw new Sk.builtin.NotImplementedError("math.isqrt() is not yet implemented in Skulpt");
     };
 
-    function lcm() {
-        const nargs = arguments.length;
+    function lcm(...args) {
+        const nargs = args.length;
 
         // lcm() without arguments returns 1
         if (nargs === 0) return new Sk.builtin.int_(1);
 
-        let result = arguments[0];
+        let result = args[0];
         Sk.builtin.pyCheckType("*integers", "integer", Sk.builtin.checkInt(result));
         result = Sk.builtin.asnum$(result);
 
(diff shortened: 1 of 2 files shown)


src/lib/math.js
```diff
@@ -314,6 +314,55 @@ const $builtinmodule = function (name) {
...
+            Sk.builtin.checkInt(argument)
+          );
+
+          // If any of the arguments is zero, then the returned value is 0
+          if (argument === 0) return 0;
+
+          let _argument = Sk.builtin.asnum$(argument);
+
+          if (typeof result === "number" && typeof _argument === "number") {
+            result = (result / gcd(result, _argument)) * _argument;
```
s-cork, line 351: for the `gcd` function: i'd consider extracting the logic that just does the javascript part into a separate function.
Otherwise we end up going through all the type checks and conversions to javascript again.

src/lib/math.js
```diff
@@ -943,6 +992,12 @@ const $builtinmodule = function (name) {
             $textsig: "($module, x, /)",
             $doc: "Return True if x is a NaN (not a number), and False otherwise.",
         },
+        lcm: {
```
s-cork, line 995: Bonus - python 3.9 also added arbirtray arguments to the `gcd` function.
Feel free to make that outside the scope of this PR. But it might be a nice feature to add.

test/unit3/test_math.py
```diff
@@ -1481,6 +1481,14 @@ def testIsinf(self):
         self.assertFalse(math.isinf(2**1024))
         self.assertFalse(math.isinf(2**2048))
 
+    def testLcm(self):
+        self.assertEqual(math.lcm(), 1)
+        self.assertEqual(math.lcm(2), 2)
+        self.assertEqual(math.lcm(0), 0)
+        self.assertEqual(math.lcm(576559230871654959816130551884856912003141446781646602790216406874), 576559230871654959816130551884856912003141446781646602790216406874)
+        self.assertEqual(math.lcm(1, 2, 0, 5, 6), 0)
+        self.assertEqual(math.lcm(2, 3, 4, 5, 6), 60)
```
s-cork, line 1490: It would be good to see the cpython lcm tests here
(commenting out the ones that don't work)
https://github.com/python/cpython/blob/main/Lib/test/test_math.py#L1051-L1093

If you end up touching `gcd` likewise:
https://github.com/python/cpython/blob/main/Lib/test/test_math.py#L705-L745

s-cork (changes requested):
Nice feat branch! I've made some suggestions. Hope it helps.

BioMedTech:
I accidentally pushed code from https://github.com/skulpt/skulpt/pull/1368

src/lib/math.js
```diff
@@ -314,6 +314,55 @@ const $builtinmodule = function (name) {
...
+
+        let result = arguments[0];
+        Sk.builtin.pyCheckType("*integers", "integer", Sk.builtin.checkInt(result));
+        result = Sk.builtin.asnum$(result);
+
+        if (nargs === 1) {
+            if (typeof result === "number") {
+                return new Sk.builtin.int_(Math.abs(result));
+            } else {
+                return new Sk.builtin.int_(result);
```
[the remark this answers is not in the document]
BioMedTech: I added this check twice, also at the end. Not sure it it is required though.

src/lib/math.js
```diff
@@ -314,6 +314,55 @@ const $builtinmodule = function (name) {
...
     };
 
+    function lcm() {
+        const nargs = arguments.length;
+
+        // lcm() without arguments returns 1
+        if (nargs === 0) return new Sk.builtin.int_(1);
+
+        let result = arguments[0];
+        Sk.builtin.pyCheckType("*integers", "integer", Sk.builtin.checkInt(result));
```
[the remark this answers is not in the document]
BioMedTech: Adding same thing in a loop (`const argument = Sk.misceval.asIndexOrThrow(args[i]);`). It breaks tests with `Error: Assertion failure: JS object passed to richCompareBool`. Should I even use it there? It has same checks as for first argument.

src/lib/math.js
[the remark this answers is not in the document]
s-cork: Your `abs` function will need to behave different depending on whether you have a number or a JSBI.BigInt.

Another thing to note in this implementation is that - since lcm will keep growing in size - you will need to keep checking that the result hasn't grown larger than `Number.MAX_SAFE_INTEGER`. As soon as the result starts getting bigger than that you'll need to start working with JSBI.BigInts.

src/lib/math.js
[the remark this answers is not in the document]
s-cork: Yes you should. This means that a javascript object is being leaked into python.

It's because you're returning `0` on line 348 (rather than a python integer). 

In the original version that code path never hit because at that stage `argument` was a python object so the test
`if (argument === 0)` was always falsey.

src/lib/math.js
BioMedTech: After finishing with this pr I plan to do other missing math methods. If I will have spare time I will try to get to updating gcd)

src/lib/math.js
BioMedTech: Something like this?

```js
    function _gcd_internal(_a, _b) {
        function _gcd(a, b) {
            if (b == 0) {
                return a;
            }
            return _gcd(b, a % b);
        }
        
        function _biggcd(a, b) {
            if (JSBI.equal(b, JSBI.__ZERO)) {
                return a;
            }
            return _biggcd(b, JSBI.remainder(a, b));
        }

        let res;
        if (typeof _a === "number" && typeof _b === "number") {
            _a = Math.abs(_a);
            _b = Math.abs(_b);
            res = _gcd(_a, _b);
            res = res < 0 ? -res : res; // python only returns positive gcds
            return new Sk.builtin.int_(res);
        } else {
            _a = JSBI.BigInt(_a);
            _b = JSBI.BigInt(_b);
            res = _biggcd(_a, _b);
            if (JSBI.lessThan(res, JSBI.__ZERO)) {
                res = JSBI.multiply(res, JSBI.BigInt(-1));
            } 
            return new Sk.builtin.int_(res.toString()); // int will convert strings
        }
    };

    function gcd(a, b) {
        // non ints not allowed in python 3.7.x
        Sk.builtin.pyCheckType("a", "integer", Sk.builtin.checkInt(a));
        Sk.builtin.pyCheckType("b", "integer", Sk.builtin.checkInt(b));

        let _a = Sk.builtin.asnum$(a);
        let _b = Sk.builtin.asnum$(b);

        return _gcd_internal(_a, _b);
    };
```

And then use _gcd_internal in lcm function.

test/unit3/test_math.py
BioMedTech: Thanks, thought of that but in the end I forgot to add it)

BioMedTech:
Very strange, it doesn't fail locally.

src/lib/math.js
s-cork: Yes although - I don't think you want the internal version to return a python object.

i.e. return the raw res from `_gcd_internal` and the final line of `gcd` should be something like 

```javascript
const res = _gcd_internal(_a, _b);
// res is positive
if (typeof res === "number") {
    return new Sk.builtin.int_(res);
} else {
    return new Sk.builtin.int_(res.toString());
    // this is a bit of a hack - we're being lazy and passing a string to int_
    // the int constructor will then decide if it should be a number or a BigInt depending on the size.
    // really we should do:
    // return  JSBI.lessThanOrEqual(res, JSBI.BigInt(Number.MAX_SAFE_INTEGER)) ? new Sk.builtin.int_(Number(res)) : new Sk.builtin.int_(res);
}
```

src/lib/math.js
```diff
@@ -314,6 +314,47 @@ const $builtinmodule = function (name) {
         throw new Sk.builtin.NotImplementedError("math.isqrt() is not yet implemented in Skulpt");
     };
 
+    function lcm(...args) {
+        function abs(n) {
+            return JSBI.lessThan(result, JSBI.__ZERO)
+                ? new Sk.builtin.int_(JSBI.unaryMinus(result))
+                : new Sk.builtin.int_(result);
+        }
```
s-cork, line 322: We need to apply this differently if `n` is a `number`.

```suggestion
        function abs(n) {
            if (typeof n === "number") {
                return Math.abs(n);
            }
            return JSBI.lessThan(n, JSBI.__ZERO) ? new Sk.builtin.int_(JSBI.unaryMinus(n)) : new Sk.builtin.int_(n);
        }
```

src/lib/math.js
```diff
@@ -314,6 +314,47 @@ const $builtinmodule = function (name) {
...
+                result = (result / gcd(result, argument)) * argument;
+            } else {
+                result = JSBI.BigInt(result);
+                argument = JSBI.BigInt(argument);
+
+                result = JSBI.multiply(
+                    JSBI.divide(result, gcd(result, argument)),
+                    argument
+                );
+            }
```
s-cork, line 352: gcd really shouldn't work here - 
gcd returns a python object and the fact that division is working here is really just accidental.
(JS converts the python objects to strings which and then tries to do division 🤷 )

I'd use the `_gcd_internal` function you suggested but returning a javascript object so that the division works like it should.


Another thing we should account for is the multiplication getting too large which means we have to stop working with  numbers and start working with bigints.

Something like this will probably work:

```suggestion
            if (typeof result === "number" && typeof argument === "number") {
                let tmp = (result / _gcd_internal(result, argument)) * argument;
                tmp = Math.abs(tmp);
                // check the size - if we're too big reset the result and then fall through
                result = tmp > Number.MAX_SAFE_INTEGER ? JSBI.BigInt(result) : tmp;
            } else {
                result = JSBI.BigInt(result);
            }

            // allow fall through - if result gets too big we'll need to redo the calculation with BigInts
            if (typeof result !== "number") {
                argument = JSBI.BigInt(argument);
                result = JSBI.multiply(JSBI.divide(result, _gcd_internal(result, argument)), argument);
            }
```

s-cork (commented):
Some more suggestions.

I think when you implement the `_gcd_internal` the fails we're currently getting won't happen. At the moment it's failing because we're passing BigInts to `gcd` which is expecting python objects.

BioMedTech:
It was my bad that it wasn't failing locally. For testing specific thing I commented out most of the tests.

s-cork (approved):
Lgtm - has tests, adds functionality we don't have and only touches math.js. Ready for @skulpt/committers

albertjan:
Thanks @BioMedTech!

PR #1368, opened 2021-09-19
feat: select module to run unit tests on

BioMedTech:
While developing for this lib I found that testing could be hard since it runs tests on all modules.

This pr adds option to choose which module to test.

Usage: `node test/testunit.js --python3 --module math`


pingskills:
This PR definitely makes testing an individual module much easier. I think it would be a nice feature to add.

s-cork (approved):

s-cork:
I'm calling this a small change and merging - won't effect the current workflow only augments it

PR #1369, opened 2021-09-23
Add prod function to math module

pingskills:


test/unit3/test_math.py
```diff
@@ -1527,6 +1527,98 @@ def test_exceptions(self):
...
+        self.assertIsNaN(prod([1, 2, 3, float('inf'), -3, 0, 3]))
+        self.assertIsNaN(prod([1, 2, 3, float('-inf'), -3, 0, 2]))
+
+        # Type preservation
+
+        self.assertEqual(type(prod([1, 2, 3, 4, 5, 6])), int)
+        self.assertEqual(type(prod([1, 2.0, 3, 4, 5, 6])), float)
+        self.assertEqual(type(prod(range(1, 10000))), int)
+        self.assertEqual(type(prod(range(1, 10000), start=1.0)), float)
+        # Decimal not implemented in Skulpt yet
```
pingskills, line 1614: Commented out these tests as Decimal is not yet implemented in Skulpt

src/lib/math.js
```diff
@@ -361,8 +361,42 @@ const $builtinmodule = function (name) {
...
+
+        let result = Sk.ffi.remapToJs(start);
+        result = JSBI.BigInt(result);
+        prod_iterable = Sk.abstr.iter(iterable);
+        for (let x = prod_iterable.tp$iternext(); x !== undefined; x = prod_iterable.tp$iternext()) {
+            if (!Sk.builtin.checkNumber(x)) {
+                throw new Sk.builtin.TypeError("must be a number in prod function");
+            }
+            if (!Sk.builtin.checkInt(x)) {
+                return new Sk.builtin.float_(prod_float(iterable, start));
```
s-cork, line 395: This looks like we might end up doing two iterations and creating a second iterable.

s-cork (changes requested):
Nice work implementing passing the cpython tests.
I've added some comments to be addressed.

You might want to checkout the `sum` implementation in builtin.js

It's not the easiest to read because it accounts for code that might suspend.
But it tries to follows the cpython approach of
- do the fast thing for ints while we can
- if we hit a float continue and do the fast thing for floats while we can
- if we hit anything else continue and do the slow thing

The cpython implementation is also useful as a reference:
https://github.com/python/cpython/blob/main/Modules/mathmodule.c#L3073-L3215

pingskills:
I did not know that start could be a list! It's interesting that there were no tests for that?

I've  copied the sum implementation as you suggested. I didn't fully understand everything there but it seems to work and passes all of the tests.

test/unit3/test_math.py
s-cork: re `prod` working with lists. I read the original feat request and it was discussed there.
It's more that it doesn't break with lists rather than it's intended to work with lists.

This last test seems to be the only one that tests the slow path. A shame there isn't another test or two from cpython.
The one tagged as `# start is keyword only` would have passed if `start` was a keyword.

(Sidenote: Most of the TypeError tests are a bit weird - they don't fail because of multiplying weird types - they only fail because `start` isn't a keyword - 🤷 )

s-cork (approved):
LGTM - ready for @skulpt/committers

albertjan:
🎉

PR #1371, opened 2021-09-26
Add perm function to math module

pingskills:


s-cork (changes requested):
Looks good - I think we can tidy the iteration.

There's no need to create a skulpt style iteration here.
The reason `sum` was weird was because a python iterator might suspend.

But we're working with python integers, 
so can just use the same approach as cPython and do normal javascript iteration.
https://github.com/python/cpython/blob/main/Modules/mathmodule.c#L3303
i.e.
```
for (let i = 1; i < factors; i++) {
   // do the fast thing
}
```
(And since we know `n` and `tot` are always python integers there's no need for a slow version.)

Cpython also restricts the value of `k` to be less than `Number.MAX_SAFE_INTEGER`.
It's the line where they throw an OverflowError
https://github.com/python/cpython/blob/main/Modules/mathmodule.c#L3281
I think it'll mean you can uncomment the OverflowError test.

test/unit3/test_math.py
```diff
@@ -1571,6 +1574,72 @@ def test_exceptions(self):
...
+
+        # Returns zero if k is greater than n
+        self.assertEqual(perm(1, 2), 0)
+        self.assertEqual(perm(1, 2**1000), 0)
+
+        n = 2**1000
+        self.assertEqual(perm(n, 0), 1)
+        self.assertEqual(perm(n, 1), n)
+        self.assertEqual(perm(n, 2), n * (n-1))
+        # if support.check_impl_detail(cpython=True):
```
pingskills, line 1631: If I uncomment this test I get: Test threw exception in testPerm (name 'support' is not defined)

s-cork: Sorry - I mean comment the if statement and unindent the test.

pingskills: Ah, that makes sense. Thanks :)

s-cork (approved):
Ready for @skulpt/committers

albertjan (approved):

albertjan:
@s-cork merged the tests, feel free to merge the PR if they pass. Thanks again @pingskills. 😄

PR #1372, opened 2021-09-27
Add comb function to math module

pingskills:


src/lib/math.js
```diff
@@ -18,8 +18,40 @@ const $builtinmodule = function (name) {
...
+        if (Sk.ffi.remapToJs(k) === 0) {
+            return new Sk.builtin.int_(1);
+        }
+        if (Sk.ffi.remapToJs(k) > Number.MAX_SAFE_INTEGER) {
+            throw new Sk.builtin.OverflowError("k must not exceed " + Number.MAX_SAFE_INTEGER);
+        }
+
+        const one = new Sk.builtin.int_(1);
+        let tot = n;
+        for (let i = 1; i < k; i++) {
```
s-cork, line 48: Be careful with your types here. `k` is a python object. We shouldn't do this type of comparison between a python object and a javascript object.

src/lib/math.js
```diff
@@ -18,8 +18,40 @@ const $builtinmodule = function (name) {
...
+        if (k < 0) {
+            throw new Sk.builtin.ValueError("k must be a non-negative integer");
+        }
+        if (k > n) {
+            return new Sk.builtin.int_(0);
+        }
+        n = new Sk.builtin.int_(n);
+        k = new Sk.builtin.int_(k);
+        let temp = n.nb$subtract(k);
+        if (Sk.ffi.remapToJs(temp) < Sk.ffi.remapToJs(k)) {
```
s-cork, line 36: you've got a lot of `remapToJs` going on. It's worth noting that `remapToJs` is a slow operation. So doing it several times on the same object is unnecessary.

src/lib/math.js
```diff
@@ -18,8 +18,40 @@ const $builtinmodule = function (name) {
...
+            k = temp
+        }
+        if (Sk.ffi.remapToJs(k) === 0) {
+            return new Sk.builtin.int_(1);
+        }
+        if (Sk.ffi.remapToJs(k) > Number.MAX_SAFE_INTEGER) {
+            throw new Sk.builtin.OverflowError("k must not exceed " + Number.MAX_SAFE_INTEGER);
```
s-cork, line 43: Check your error message against here.

s-cork (changes requested):
Some javascript/python type conversions to consider but otherwise looks good.

src/lib/math.js
pingskills: Ahh, yes. It's so easy to introduce a bug!

src/lib/math.js
```diff
@@ -18,8 +18,41 @@ const $builtinmodule = function (name) {
         return new Sk.builtin.float_(Math.ceil(_x));
     };
 
-    function comb(x, y) {
-        throw new Sk.builtin.NotImplementedError("math.comb() is not yet implemented in Skulpt");
+    function comb(n, k) {
+        n_js = Sk.misceval.asIndexOrThrow(n);
+        k_js = Sk.misceval.asIndexOrThrow(k);
```
s-cork, line 23: `let`

src/lib/math.js
```diff
@@ -18,8 +18,41 @@ const $builtinmodule = function (name) {
...
+        if (k_js > n) {
+            return new Sk.builtin.int_(0);
+        }
+        n = new Sk.builtin.int_(n_js);
+        k = new Sk.builtin.int_(k_js);
+        let temp = n.nb$subtract(k);
+        if (Sk.ffi.remapToJs(temp) < k_js) {
+            k = temp;
+            k_js = Sk.ffi.remapToJs(k);
+        }
```
s-cork, line 39: consider `let temp = Sk.ffi.remapToJs(n.nb$subtract(k));`
We don't use the `k` python object after this line.

s-cork (changes requested):

src/lib/math.js
pingskills: Oh man, is it possible to automatically run eslint over these modules as part of the build process?

src/lib/math.js
pingskills: Yeah, that's much nicer :)

src/lib/math.js
s-cork: You can probably add an eslint plugin to your editor?

s-cork (approved):
Another one ready for @skulpt/committers

albertjan:
amazing thanks @pingskills

PR #1373, opened 2021-09-27
Add isqrt function to math module

### LINKED ISSUE #1201: Implemented function Math.isqrt
Also ready to open a pr.

pingskills:
close #1201

### DIFF
--- a/src/lib/math.js
+++ b/src/lib/math.js
@@ -358,8 +358,50 @@ const $builtinmodule = function (name) {
         }
     };
 
-    function isqrt(x) {
-        throw new Sk.builtin.NotImplementedError("math.isqrt() is not yet implemented in Skulpt");
+    function bigint_isqrt(n_js) {
+        // Algorithm taken from:
+        // https://github.com/python/cpython/blob/main/Modules/mathmodule.c#L1598-L1618
+        let c = n_js.toString(2).length; // bitlength of a positive bigint
+        c = Math.floor((c - 1) / 2);
+        let s = c.toString(2).length; // bitlength
+        const one = JSBI.BigInt(1);
+        const two = JSBI.BigInt(2);
+        const cBigInt = JSBI.BigInt(c);
+        const twoC = JSBI.multiply(two, cBigInt);
+        let a = one;
+        let d = JSBI.BigInt(0);
+        while (s > 0) {
+            s--;
+            let e = d;
+            d = JSBI.signedRightShift(cBigInt, JSBI.BigInt(s));
+            const leftShiftAmount = JSBI.subtract(JSBI.subtract(d, e), one);
+            const leftPart = JSBI.leftShift(a, leftShiftAmount);
+            const rightShiftAmount = JSBI.add(JSBI.subtract(JSBI.subtract(twoC, e), d), one);
+            const rightPart = JSBI.signedRightShift(n_js, rightShiftAmount);
+            a = JSBI.add(leftPart, JSBI.divide(rightPart, a));
+        }
+        let res = a;
+        if (JSBI.greaterThan(JSBI.multiply(res, res), n_js)) {
+            res = JSBI.subtract(res, one);
+        }
+        if (JSBI.lessThanOrEqual(res, JSBI.BigInt(Number.MAX_SAFE_INTEGER))) {
+            res = Number(res);
+        }
+        return new Sk.builtin.int_(res);
+    }
+
+    function isqrt(n) {
+        let n_js = Sk.misceval.asIndexOrThrow(n);
+        if (n_js < 0) {
+            throw new Sk.builtin.ValueError("isqrt() argument must be nonnegative");
+        }
+        if (n_js == 0) {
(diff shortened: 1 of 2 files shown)


src/lib/math.js
```diff
@@ -326,8 +326,40 @@ const $builtinmodule = function (name) {
...
+        }
+        const one = new Sk.builtin.int_(1);
+        const two = new Sk.builtin.int_(2);
+        let c = bit_length(n);
+        c = c.nb$subtract(one).nb$floor_divide(two);
+
+        let a = new Sk.builtin.int_(1);
+        let d = new Sk.builtin.int_(0);
+        let s;
+        if (Sk.ffi.remapToJs(c) > 0) {
```
pingskills, line 350: I needed this because the following gives different values in Python and in Skulpt
```
i = 0
print(i.bit_length())
```
Returns 0 in Python
Returns 1 in Skulpt

s-cork: Fixed in #1362

src/lib/math.js
```diff
@@ -326,8 +326,40 @@ const $builtinmodule = function (name) {
...
 
-    function isqrt(x) {
-        throw new Sk.builtin.NotImplementedError("math.isqrt() is not yet implemented in Skulpt");
+    function bit_length(i) {
+        return new Sk.builtin.int_(Sk.builtin.bin(i).sq$length() - 2);
+    }
+
+    function isqrt(n) {
+        n_js = Sk.misceval.asIndexOrThrow(n);
+        n = new Sk.builtin.int_(n_js);
```
s-cork, line 335: I'm not sure I like this approach of working with Python objects for `isqrt`.
It's seems a little obtuse.

What about working exclusively with javascript objects?
If `n_js` is a number we can just do the quick thing with `Math.` functions.

And if it's not a number we know it's a JSBI.BigInt.
we can then follow a sqrt algorithm using `JSBI.` calls.

pingskills: I've had a go at doing this. Let me know of any suggested improvements (especially for the BigInt part).

Cheers,
Jeff.

src/lib/math.js
```diff
@@ -326,8 +326,44 @@ const $builtinmodule = function (name) {
...
+        while (s > 0) {
+            s--;
+            e = d;
+            d = JSBI.signedRightShift(JSBI.BigInt(c), JSBI.BigInt(s));
+            a = JSBI.add(JSBI.leftShift(a, JSBI.subtract(JSBI.subtract(d, e), JSBI.BigInt(1))), JSBI.divide(JSBI.signedRightShift(n_js, JSBI.add(JSBI.subtract(JSBI.subtract(JSBI.multiply(JSBI.BigInt(2), JSBI.BigInt(c)),e), d), JSBI.BigInt(1))), a));
+        }
+        if (JSBI.greaterThan(JSBI.multiply(a, a), n_js)) {
+            return new Sk.builtin.int_(JSBI.subtract(a, JSBI.BigInt(1)));
+        }
+        return new Sk.builtin.int_(a);
```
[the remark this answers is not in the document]
pingskills: If it's not less than Number.MAX_SAFE_INTEGER do we throw a value error?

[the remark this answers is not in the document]
pingskills: It seems this change causes some tests to fail. If the number is bigger than Number.MAX_SAFE_INTEGER, can't we still represent it as an int_?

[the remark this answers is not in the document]
s-cork: No `ValueError` we just need to determine what the internal representation of the python `int_` should be before calling the `int_` constructor.

If the value is less than or equal to `Number.MAX_SAFE_INTEGER` the internal representation should be a number
whereas if it's greater than `Number.MAX_SAFE_INTEGER`, the internal representation should be a `JSBI.BigInt`.
(See the description above the `int_` constructor in `int.js` for more details)

e.g. if `n = 20282409603651670423947251286016` the internal representation starts as a bigint
But when we sqrt we get `4503599627370496` which is less than `Number.MAX_SAFE_INTEGER` and so the internal representation should become a number.

src/lib/math.js
```diff
@@ -358,8 +358,46 @@ const $builtinmodule = function (name) {
...
+        let s = c.toString(2).length; // bitlength
+        const one = JSBI.BigInt(1);
+        const two = JSBI.BigInt(2);
+        const cBigInt = JSBI.BigInt(c);
+        const twoC = JSBI.multiply(two, cBigInt);
+        let a = one;
+        let d = JSBI.BigInt(0);
+        while (s > 0) {
+            s--;
+            e = d;
```
s-cork, line 375: I'm not sure `e` has been declared. Likewise `n_js` in `isqrt`

pingskills: Is there a way of having the build run these types of checks for us?

s-cork: I'm sure there is - would be a really useful pr.
Eslint with the right configuration would probably work.

s-cork (approved):
Ready for @skulpt/committers

PR #1376, opened 2021-09-30
Fix nb$ispositive for int

pingskills:


s-cork (approved):
clearly a mistake - merging

PR #1378, opened 2021-09-30
fix: missing $len from dict iterators

s-cork:
The value is never instantiated and since we're about to throw anyway just get the value at runtime

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -373,6 +373,9 @@ const pyExc = {
 
 Object.assign(Sk.builtin, pyExc);
 
+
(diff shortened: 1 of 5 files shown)


meredydd:
Erm...what does this fix? The commit message is pretty vague, and the original doesn't fail any tests...was the previous behaviour broken, and if so how?

s-cork:
.$len was never instantiated in the constructor for the operator so the first test was always falsely. Which meant the second error message was thrown always. 

This pr means the first error actually gets thrown with the correct error message.

PR #1382, opened 2021-10-06
Fix: accidental removal of SuspensionError

s-cork:
Oops - when injecting python Exceptions in #1343 SuspensionError got lost.
Added a test.


meredydd (approved):
Small, important fix. Ship it.

PR #1386, opened 2021-10-10
Adjust the Name tokenize regex

### LINKED ISSUE #637: Any idea how hard it would be to support unicode characters in variables ?
I've had this request from a Swedish teacher.

s-cork:
in tokenize.py the `Name` variable is `\w+`.
But a Name in python is not `\w+`.

This pr adjusts our implementation of tokenize so that the `Name` variable matches a the regex for a python identifier.

close #637


bjorn3:
I just tested this PR. I can confirm this works.

cc @Felienne This will remove the need for hashing of unicode identifiers in [Hedy](https://github.com/Felienne/hedy).

Felienne:
> I just tested this PR. I can confirm this works.
> 
> cc @Felienne This will remove the need for hashing of unicode identifiers in [Hedy](https://github.com/Felienne/hedy).

Nice find @bjorn3!! Can you make an issue in our repo to update this when it comes available?

Felienne:
Hi folks! Is there any progress on this? Our project Hedy would very much love to use this! Is there anything we can do to help it move forward?

meredydd (approved):
LGTM

PR #993, opened 2019-12-20
turtle.Turtle add the shape keyword.

Dustyposa:
turtle's feature.
A small change for #988 
😄

PR #1398, opened 2021-11-22
To fix a python run error: 

leafdown:
Sample Code as : 

> import turtle
> 
> class MyTurtle(turtle.Turtle):
>   def __init__(self, screen = turtle.Screen()):
>     turtle.Turtle.__init__(self, screen)
>     self.hideturtle()
>     
> def create_turtles(screen, n = 10):
>   for i in range(n):
>     MyTurtle(screen)
>     
> def move_turtles(screen, dist=10, angle = 4):
>   for i, turtle in enumerate(screen.turtles()):
>     turtle.left(angle*(1+i))
>     turtle.forward(dist)
>     x, y = turtle.pos()
>     try:
>       turtle.color(abs(x), abs(y), abs(x+y))
>     except:
>       pass
>     
> writer = MyTurtle()
> writer.penup()
> writer.goto(0,100)
> writer.write("Click Me!", font=("Arial",30), align = "center")
> 
> 
> number_of_turtles = 10
> 
> screen = turtle.Screen()
> 
> def draw_shape(x, y, n = 20, clear = True):
>   if clear:
>     writer.clear()
>   screen.tracer(0)
>   for turtle in screen.turtles():
>     turtle.penup()
>     turtle.goto(x,y)
>     turtle.pendown()
>   screen.tracer(1)
>   for i in range(n):
>     screen.tracer(0)
>     move_turtles(screen)
>     screen.tracer(1)
> 
> create_turtles(screen, number_of_turtles)
> 
> draw_shape(0,-150, clear = False)
> 
> screen.onclick(draw_shape)
> 
> screen.listen()
> turtle.done()

### DIFF
--- a/src/lib/turtle.js
+++ b/src/lib/turtle.js
@@ -2340,20 +2340,6 @@ function generateTurtleModule(_target) {
     addModuleMethod(Screen, _module, "$window_width", getScreen);
     addModuleMethod(Screen, _module, "$window_height", getScreen);    
     addModuleMethod(Screen, _module, "$title", getScreen);
-    addModuleMethod(Screen, _module, "$resetscreen", getScreen);
-    addModuleMethod(Screen, _module, "$setup", getScreen);
-    addModuleMethod(Screen, _module, "$onkey", getScreen);
-    addModuleMethod(Screen, _module, "$listen", getScreen);
-    addModuleMethod(Screen, _module, "$register_shape", getScreen);
-    addModuleMethod(Screen, _module, "$clearscreen", getScreen);
-    addModuleMethod(Screen, _module, "$bgcolor", getScreen);
-    addModuleMethod(Screen, _module, "$bgpic", getScreen);
-    addModuleMethod(Screen, _module, "$setworldcoordinates", getScreen);
-    addModuleMethod(Screen, _module, "$ontimer", getScreen);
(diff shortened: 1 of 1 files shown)


s-cork:
I don't think this is the right fix for this issue.
The issue is really that the call signature to the class Turtle from commit #993, isn't quite right.
The first argument should be the screen/canvas rather than the shape.

See cpython code for the __init__ method to RawTurtle:
https://github.com/python/cpython/blob/main/Lib/turtle.py#L2521-L2524

You might be able to have a look at pr #993 and see if you can adjust it to include the canvas argument as the first argument.

leafdown:
> I don't think this is the right fix for this issue. The issue is really that the call signature to the class Turtle from commit #993, isn't quite right. The first argument should be the screen/canvas rather than the shape.
> 
> See cpython code for the `__init__` method to RawTurtle: https://github.com/python/cpython/blob/main/Lib/turtle.py#L2521-L2524
> 
> You might be able to have a look at pr #993 and see if you can adjust it to include the canvas argument as the first argument.

I think this issue was caused by the code be removed as you mentioned. 
Do you think that we should add protection or rollback https://github.com/skulpt/skulpt/pull/993 fixed?

s-cork:
I would vote for adding the the canvas argument to the init method.

PR #1401, opened 2021-12-09
Fix/finally raises

### LINKED ISSUE #1400: try finally with a return and an exception
```python
def foo():
  try:
    return
  finally:
    print("finally")
    1/0

foo()
```
In the above example I'd expect finally to get printed only once.
But in skulpt finally is printed twice.

A similar thing happens in a context manager's `__exit__` method.
(if there is a return in the with statement and the `__exit__` block raises, it will get called again).

The bug is caused in compile code in the `Return` case.
We jump to the finally block before we've cleaned up the exception blocks.
Which means we don't pop the `$exc` object correctly.
So when the finally block raises, we still have an exception block in the `$exc` array.
And the final handler says oh wait, we still have an exception block to visit, let's go visit it.

It's worse if we have Exception handlers
```python
def foo():
  try:
    return
  except:
    print("except")
    return
  finally:
    print("finally")
    1/0

foo()
# finally, except, finally, finally
```

s-cork:
close #1400 

We weren't popping `$exc` array when jumping to a finally block from a return statement.
This meant that if a finally block (or an __exit__ function) raised an error they would get revisited.

See #1400 for examples as well as the tests added to this pr

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -1613,10 +1613,14 @@ Compiler.prototype.ctry = function (s) {
 
     if (handlers.length != 0) {
         this.setupExcept(handlers[0]);
+        this.pushExceptBlock(handlers[0]);
     }
     this.vseqstmt(s.body);
     if (handlers.length != 0) {
+        // note that if the try body jumped to a new blk (in the case of a return statement)
+        // then the output of this.endExcept() will be dropped by the compiler
         this.endExcept();
+        this.popExceptBlock();
     }
     this._jump(orelse);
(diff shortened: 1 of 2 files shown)


s-cork:
reverting to draft:
realised this approach breaks for a nested try finally, like:

```python
def foo():
  try:
    try:
       return 1
    finally:
      print("finally")
      raise Exception("foo")
  except:
    print("except")

foo()
```

the except doesn't get called

PR #1409, opened 2022-02-05
dir: make it match cpython

s-cork:
If you like reading c - here are the implementations that I was following

`type__dir__`:
https://github.com/python/cpython/blob/fea7290a0ecee09bbce571d4d10f5881b7ea3485/Objects/typeobject.c#L4280-L4291

`object__dir__`:
https://github.com/python/cpython/blob/fea7290a0ecee09bbce571d4d10f5881b7ea3485/Objects/typeobject.c#L5564-L5609

`merge_class_dict`:
https://github.com/python/cpython/blob/fea7290a0ecee09bbce571d4d10f5881b7ea3485/Objects/typeobject.c#L4215-L4267

It's probably a slower implementation - but looking up the `__dir__` is mostly for introspection.
It's also more correct - I was able to uncomment a test that would have previously failed.
It came up when overriding the `__dict__` property and discovered that cpython uses the return value from `__dict__`, rather than the internal dict attribute, when resolving the dir.

### DIFF
--- a/src/abstract.js
+++ b/src/abstract.js
@@ -921,6 +921,16 @@ Sk.abstr.lookupSpecial = function (obj, pyName) {
 };
 Sk.exportSymbol("Sk.abstr.lookupSpecial", Sk.abstr.lookupSpecial);
 
+Sk.abstr.lookupAttr = function (obj, pyName) {
+    try {
+        return obj.tp$getattr(pyName);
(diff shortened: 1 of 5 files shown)


meredydd (approved):
I applaud @s-cork's commitment to compatibility nit-picking. LGTM.

PR #1414, opened 2022-03-08
Fix float and int don't support __index__

### LINKED ISSUE #1413: `int()` and `float()` doesn't fall back to `__index__()`
## Code:
```python
class Demo:
    def __index__(self):
        return 1


instance = Demo()

print(int(instance))
print(float(instance))
```

## Expected result:
1
1.0

## Real result:
`<TypeError: int() argument must be a string, a bytes-like object or a number, not 'Demo' on line 8>`

## CPython documentation:
https://docs.python.org/3/reference/datamodel.html?highlight=__index__#object.__index__
> If [__int__()](https://docs.python.org/3/reference/datamodel.html?highlight=__index__#object.__int__), [__float__()](https://docs.python.org/3/reference/datamodel.html?highlight=__index__#object.__float__) and [__complex__()](https://docs.python.org/3/reference/datamodel.html?highlight=__index__#object.__complex__) are not defined then corresponding built-in functions [int()](https://docs.python.org/3/library/functions.html#int), [float()](https://docs.python.org/3/library/functions.html#float) and [complex()](https://docs.python.org/3/library/functions.html#complex) fall back to [__index__()](https://docs.python.org/3/reference/datamodel.html?highlight=__index__#object.__index__).

s-cork:
close #1413 

tests added from cpython - i found the ones that seemed relevant with `__index__` and added them.

### DIFF
--- a/src/int.js
+++ b/src/int.js
@@ -421,6 +421,7 @@ Sk.builtin.int_ = Sk.abstr.buildNativeClass("int", {
         valueOf() {
             return this.v;
         },
+        sk$int: true,
     },
 });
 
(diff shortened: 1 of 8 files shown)


src/int.js
```diff
@@ -327,6 +327,7 @@ Sk.builtin.int_ = Sk.abstr.buildNativeClass("int", {
         valueOf() {
             return this.v;
         },
+        sk$int: true,
```
meredydd, line 330: Could we have a brief comment on why this is required please?

meredydd (approved):
Makes sense! Per conversation with @s-cork, the new `sk$int` attribute is justified (to pass CPython tests), but probably wants a comment before we merge!

PR #1416, opened 2022-03-09
Fix: __eq__ set with no __hash__ makes a class unhashable

s-cork:
See python docs:
https://docs.python.org/3/reference/datamodel.html#object.__hash__

Uncomments a test that previously failed.
Plus another one that was already passing on master


meredydd (approved):
A correctness fix with tests. We like!

s-cork:
small change merging

PR #1419, opened 2022-03-24
make circular import errors work with current import implementation

s-cork:
In cpython import lib checks a module's spec to see if it's initializing
If it's initializing then a `"most likely due to a circular import"` error message is thrown.

We don't have specs in skulpt.
Instead, I add a `.$initializing` attribute to modules before we eval them.
Since a `relativePackage` is responsible for importing its leaf modules, I add the same attribute to the `relativePackage` (if it's not already in the initializing state).

I added the Circular Import tests from Cpython and tweaked them slightly.
Most tests failed because skulpt thought there were circular imports going on that weren't
That seemed to be because the `relativePackage` didn't know anything about the module it was importing early enough.
To fix this I add the module as an attribute to the `relativePackage` at the same time we put it in `sys.modules`
If the import fails, at the same time as removing the module from `sys.modules`, I also remove the module from the `relativePackage`'s attributes.

Since our internal `tp$getattr` implementations are supposed to return `undefined`,
To ensure that the `"most likely due to a circular import"` error message is thrown, I adjust the `sk$attrError()` callback.
This gets called from compile code to throw a somewhat correct `AttributeError`.

In the compile code I also removed a fixReseverd call, which was no longer needed (and actually a little bit broken) since the names for __import__ get converted to `Sk.builtin.str` when  we do `tp$getattr` and `tp$setattr`. Keeping it around meant relative modules would have a `_$rw$` in their python name.

Finally I adjust the `testunit.js` file - since it didn't support packages.

😅

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -1837,7 +1837,7 @@ Compiler.prototype.cfromimport = function (s) {
         level = -1;
     }
     for (i = 0; i < n; ++i) {
-        names[i] = "'" + fixReserved(s.names[i].name.v) + "'";
+        names[i] = "'" + s.names[i].name.v + "'";
     }
     out("$ret = Sk.builtin.__import__(", s.module["$r"]().v, ",$gbl,$loc,[", names, "],",level,");");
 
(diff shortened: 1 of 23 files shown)


meredydd (approved):
Hurrah! Major quality of life improvement. Ship it.

PR #1420, opened 2022-03-24
Fix nesting import chaining

s-cork:
fix regression from #1419 

Javascript scoping rules meant that the formName was not scoped correctly
This Pr fixes that

(prior to #1419 we nested the call at each iteration, which meant we didn't need to worry about scoping in the same way)


meredydd (approved):
Oops! Merging the fix.

PR #1422, opened 2022-04-01
Implement __import__

s-cork:
tests taken from cpython

I think the the `__import__` implementation is fairly well tested since it's part of our compile code,
so really this pr just supports actually calling `__import__` from python code.
Which currently just breaks because it expects javascript types.

I do some type checking but also do the lazy thing of just remapping the objects to javascript since:
- from list should be a tuple/list of strings, and
- level should be an integer


meredydd (approved):
Looks good! Merge it :)

PR #1429, opened 2022-04-19
fix reserved words don't need to be mangled in annotations

### LINKED ISSUE #1428: __annotations__ has mangled keys
I believe that this is an issue with `__annotations__`:

```python
class Dog:
    name: str
```

The current code will make it so that the key is `name_$rw$` in the dictionary.

https://github.com/skulpt/skulpt/blob/5f5d40d973de02b0d8316557e4a7d48ad8616f5b/src/compile.js#L1090-L1092

But the names shouldn't be mangled here, should they? They should just be placed in the __annotations__ dictionary directly?

s-cork:
close #1428


test/unit3/test_skulpt_bugs.py
```diff
@@ -24,5 +24,14 @@ def test_bug_1345(self):
             self.fail("this shouldn't fail")
 
 
+class An:
+    name: str
+    __foo: int
+
+class TestAnnotaitons(unittest.TestCase):
```
meredydd, line 31: speling

meredydd (approved):
Yep, looks good! Have approved, but it would be nice to fix the typo in the test name.

s-cork:
typo fixed 😄 (small change - merging)

PR #1434, opened 2022-06-09
Use weakmap instead of map

s-cork:
Probably better to use a WeakMap in these two places so that we don't hold onto references unnecessarily


meredydd (approved):
Ew, yes, we don't want to be leaking anything you ever called `id()` on. Browser compat is surprisingly high - even IE has it!

PR #1435, opened 2022-06-09
make skulpt "is" equivalent to object identity

s-cork:
We currently have a hack for object identity with floats and ints
in python only the ints `-5 <= x <= 256` are interned
floats are not interned so we shouldn't expect object identity there

```python
>>> x = 300; y = 300
>>> x is y
False
```

In python you will often get identity because the compiler does interning of constants
This is also the case in skulpt
But we don't have constant folding like cpython does
```python
print(3.0 is 3.0)
# True for both skulpt and cpython
print(-2.0 is -2.0)
# False for skulpt (this branch) True for cpython
# in skulpt this becomes Sk.abstr.numberUnaryOp($scope0.$const6, 'USub');
# because we don't have constant folding 
```

_(Note that @albertjan and I have implemented constant folding in https://github.com/skulpt/skulpt_parser/pull/109)_

---

This PR means we can inline the "Is" and "IsNot" operator to just be `===` and `!==` in compiled code.


meredydd (approved):
So, I would merge this - it's a reduction in complexity and `is` really _should_ be object identity.

**BUT:** Do we really want to sprain our backs going for exact CPython compatibility with regard to integer interning? If it gives us a speed boost, sure, by all means, but otherwise I'm not convinced we need to ape CPython so exactly. What code will rely on this, apart from stunt code that's explicitly demonstrating implementation details of CPython?

s-cork:
I guess the options:

- A: stick with current behaviour
  - possibly not the right option, since identity of floats isn't really a thing in python
  - might encourage bad coding habits if users can rely on identity of floats and ints which would fail in normal python
- B: merge this PR i.e. intern ints from `-5`-`256`
  - +: same behaviour as cpython
- C: adjust this PR without interning ints
  - +: we get the same benefits using `===` and `!==`
  - -: ints will only ever be identical if they are interned as part of the constant interning during compilation
  - -: it won't match cpython and there is a test for this:
    https://github.com/skulpt/skulpt/blob/4c834635442ec91f3e1129d8f8901dc3ca0ab8db/test/unit3/test_long.py#L979-L995

(my preference is for merging)

rixner:
I don't have any problem with merging this.  However, I will say that I also do not think that we should "sprain our backs going for exact CPython compatibility" on things like this that are not part of the spec.  This would only matter as @meredydd said for "stunt code".  Anyone who is relying on this behavior shouldn't be.

rixner:
("This behavior" meaning that -5 through 256 are interned)

PR #1436, opened 2022-06-11
Added the JoBase Python Game Library - Implemented in JavaScript

Grey41:
Hello, I hope this is the place to contribute new libraries. I have just built a JavaScript implementation of [JoBase](https://jobase.org) in Skulpt. You can see it in action at [jobase.org/Browser](https://jobase.org/Browser). It is currently compatible with JoBase 1.6 and also supports some new parts of JoBase 1.7.

Because the online version requires a canvas, you may pass an `HTMLCanvasElement` to the `Sk.JoBase` variable.

```
Sk.JoBase = document.createElement("canvas")
```

Please let me know if anything needs to be changed.


bnmnetp:
Hi @Grey41 thanks for the PR.

I tried to run the example you linked to, but I cannot move the man with my arrow keys.

This does seem like a nice addition that would be useful for some classes and students!

Grey41:
It should work now. The canvas is focused so that the user can interact with it.

Below is a minimal html example that runs a small piece of JoBase code. It is based on the Turtles example on the Skulpt main website.

```
<!DOCTYPE html>
<html>
    <head>
        <script src = https://jobase.org/Browser/skulpt/skulpt.min.js></script>

        <!-- This links to the compiled Skulpt libraries with JoBase -->
        <script src = https://jobase.org/Browser/skulpt/skulpt-stdlib.js></script>
    </head>

    <body>
        <button type = button onclick = run()>Run</button> 
        <canvas id = canvas></canvas> 

        <script>
            function run() {
                Sk.JoBase = canvas
                Sk.configure({output: text => console.log(text)})

                Sk.misceval.asyncToPromise(
                    () => Sk.importMainWithBody("<stdin>", false, `
from JoBase import *

text = Text("Hello World", color = (1, 0, 0))

def loop():
    text.draw()

run()`, true))
            } 
        </script>
    </body> 
</html>
```

meredydd:
Hi @Grey41,

Thanks for the PR, and apologies for being a bit late getting back!

The library looks really cool, and it's an example of the kind of thing we want to enable people to do with Skulpt! However, I'm not sure it belongs in our standard library. Skulpt is a Python implementation, which aims towards CPython compatibility. This does mean that, for historical reasons, libraries like `turtle` are part of our standard library. But additional libraries, however cool,  should probably be distributed separately. (I say this particularly from the perspective of Anvil, which serves Skulpt as part of every app we serve, and we already have to work quite hard to keep our page weight down!)

I would recommend packaging `JoBase` as a third-party library, as a separate GitHub repository (perhaps as a `.js` file you can include after `skulpt-stdlib.js`, which adds itself to `Sk.builtin_files`? @albertjan might have an opinion on the best way to do this). Then you can submit a PR to `gh-pages` to add a link to your project from our website!

Grey41:
Thank you. I understand that Skulpt wants to keep its standard libraries minimal and in harmony with the standard Python libraries. I will try out the suggestions and maybe package JoBase as a separate library.

bnmnetp:
@Grey41 I grabbed this PR for my Fork of skulpt.  I think there would be some real interest in it for our introductory textbooks on https://runestone.academy  But it appears that maybe there is some bit rot that has caused this version to not work.  

If you are interested in reviving this effort, let me know and we can try to figure out a way forward.

The repo is https://github.com/RunestoneInteractive/skulpt

Brad

Grey41:
Your project looks good, I'll see what I can do. I have recently migrated to WebAssembly instead, but I'll create a Skulpt version for your repo.

PR #1438, opened 2022-06-23
Fixed #1437

### LINKED ISSUE #1437: turtle.colormode(255) Cannot be used with t.color(150,150,0)
for example:
```
import turtle
turtle.colormode(255)
t=turtle.Turtle()
t.pensize(5)
col=[(150,150,0),(0,150,150),(180,0,180),(220,100,100)]
t.up()
t.goto(-170,150)
t.down()
for i in range(4):
    t.color(col[i])
    t.fd(12)
    t.up()
    t.fd(10)
    t.down()
```
Must be changed to：

```
import turtle
t=turtle.Turtle()
t.colormode(255)
t.pensize(5)
col=[(150,150,0),(0,150,150),(180,0,180),(220,100,100)]
t.up()
t.goto(-170,150)
t.down()
for i in range(4):
    t.color(col[i])
    t.fd(12)
    t.up()
    t.fd(10)
    t.down()
```
The modified version can be used on the platform, but this is not allowed in Python

alexwenbj:
colormode() only works in Screen Class. fixed #1437


bnmnetp (approved):

s-cork (approved):

alexwenbj:
> I'm happy to try to review this, but I'm not very familiar with the Turtle module. Could you please explain a bit more about what this change does, and why it makes us better/more correct?

Sorry for my late reply.Simply say, this PR only move the colormode() from Turtle class to Screen class, to align with the Python implement.

PR #1439, opened 2022-06-24
Pr/namedtuple

s-cork:
namedtuples in CPython are python classes rather than C or Native classes
in Skulpt they're Native classes

On a regular Python class you can dynamically add attributes to your class
```python
class A: pass
A.foo = 'bar'
```

But in a native class this is disallows
```python
dict.foo = 'bar'
# TypeError
```

There's a flag for this in `type.protoype.tp$setattr`
We check if the class is an `sk$klass`. If it is, then we consider the class a python class for attribute setting
If not then it's a native class and we throw a `TypeError` error:

https://github.com/skulpt/skulpt/blob/1103d4a475276ca325599fae68e68344c3e3f88e/src/type.js#L252-L259

By adding this flag to the namedtuple class we support setting class attributes dynamically


bnmnetp (approved):

s-cork:
Small change

PR #1440, opened 2022-06-28
Fix: multiple class decorators applied in wrong order

pjdewitte:
This fixes the order in which decorators are applied to classes.

Previously this was fixed for functions in #1149. The same fix is applied in this PR.

I also added a test for this, and extended the existing test for functions.


s-cork (approved):
LGTM, small change, with tests. merging. thanks @pjdewitte

PR #1442, opened 2022-07-04
fix property and super should be suspension aware

s-cork:
Not particularly easy to add tests for this since python attribute access, when compiled to javascript, checks for suspensions

But it will be an issue for any internal code that expects `tp$gettattr` or `Sk.abstr.gattr` to not suspend unexpectedly.


meredydd (approved):
LGTM, fixes a bug. Ship it.

PR #1447, opened 2022-07-15
fix bad variable declaration in compile code.

s-cork:
Fix a bug from recent PRs

In a previous PR we declared a variable for the instance that would be passed to `super()`
(previous to that PR the instance to `super()` was always assumed to be variable named `self` but it could easily be `cls` for a classmethod).

But the compile code ended up with a bug
```js
var $sup = self = args[0], ...;
```
Now fixed to be
```js
var self = args[0], ..., $sup = self;
```

The former overrides the global self - i.e. window.self! 😬
The latter uses the locally scoped self ☑️


meredydd (approved):
Oops. Good fix.

PR #1452, opened 2022-08-19
Fix/suspension dunders

s-cork:
A couple of dunders were not entirely suspension aware.

This is mostly ok, because skulpt doesn't use suspensions for these internally - e.g. a skulpt dictionary never suspends.

However it's not great for user code with a `@property` decorator
and anything that calls `object.__getattribute__` in a way that might lead to a suspension.
I added tests for these two cases that failed.

I also added suspension awareness to the other dunders that can take a `canSuspend` argument.
These can't be tested internally because we don't have native classes that care about suspension for say the `__len__`/`sq$length`.
But it's feasible others might want to make use of suspensions in this way.

### DIFF
--- a/src/slotdefs.js
+++ b/src/slotdefs.js
@@ -464,11 +464,13 @@ slots.__getattribute__ = {
         if (!Sk.builtin.checkString(pyName)) {
             throw new Sk.builtin.TypeError("attribute name must be string, not '" + Sk.abstr.typeName(pyName) + "'");
         }
-        const res = this.call(self, pyName);
-        if (res === undefined) {
-            throw new Sk.builtin.AttributeError(Sk.abstr.typeName(self) + " has no attribute " + pyName.$jsstr());
-        }
-        return res;
+        const res = this.call(self, pyName, true);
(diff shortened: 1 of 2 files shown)


src/slotdefs.js
```diff
@@ -99,10 +123,10 @@ function wrapperRichCompare(self, args, kwargs) {
     return new Sk.builtin.bool(res);
 }
 
-function wrapperCallBack(wrapper, callback) {
+function wrapperCallBack(wrapper, callback, canSuspend) {
     return function (self,args, kwargs) {
         const res = wrapper.call(this, self, args, kwargs);
-        return callback(res);
+        return canSuspend ? Sk.misceval.chain(res, callback) : callback(res);
```
meredydd, line 129:
```suggestion
        return canSuspend ? Sk.misceval.chain(res, callback) : callback(retryOptionalSuspensionOrThrow(res));
```

meredydd (approved):

src/slotdefs.js
meredydd: Approval contingent on this change (or on explaining while I'm wrong)

PR #1454, opened 2022-09-02
func.__defaults__ should have a set method

s-cork:
I found a use case for being able to override `__defaults__`.
This PR makes `__defaults__` a writable attribute.

I added the tests from the cpython test module `test_funcattrs.py`
Any other tests commented out in that file are beyond the scope of this PR
But should probably be addressed at some point


meredydd (approved):
Obviously a net addition, and an incremental one. Ship it.

PR #1456, opened 2022-09-03
Always provide the second argument to `tp$descr_get`

s-cork:
For internal calls to `tp$descr_get` we should send both the instance and the type argument.

Whilst most builtins that use `__get__` cope without the second argument, it's possible that a custom `__get__` will expect it to be there.
(I found this recently with another library).

Quite a small change, I don't think it really needs any additional tests since the call signature to `__get__` is typically `def __get__(self, __obj: Any, __type: type | None = None) -> Any:`


meredydd (approved):
These are all obviously correct changes; waiving tests for each one in the interest of getting the fix in!

PR #1473, opened 2022-11-08
partial implementation of uuid

s-cork:
implements `uuid4()` which is probably the more common use of the uuid module

partial support for `int.from_bytes()` and `int.to_bytes()` (`signed=True` not supported)

Tests added for the above

The tests also highlighted an issue with our parsing of numbers - see ast.js for the change
Skulpt fails with a number like `0x000102030405060708090a0b0c0d0e0f` where it gets compiled as `float(0.0)` 😬 
This is tested in the uuid module but happy to add an additional test - and or make this fix a separate pr.


src/int.js
```diff
@@ -213,6 +219,51 @@ Sk.builtin.int_ = Sk.abstr.buildNativeClass("int", {
...
+                let [bytes, byteorder, signed] = Sk.abstr.copyKeywordsToNamedArgs(
+                    "from_bytes",
+                    ["bytes", "byteorder", "signed"],
+                    args,
+                    kws,
+                    [Sk.builtin.bool.false$]
+                );
+                const s_lit = new Sk.builtin.str("little");
+                const s_big = new Sk.builtin.str("big");
+                if (byteorder !== s_big && byteorder !== s_lit) {
```
meredydd, line 235: This relies on interned strings in a way that freaks me out! Can I at least request a comment?

meredydd (approved):

src/int.js
s-cork: Tweaked and extracted the logic so we don't rely on interning of strings here

PR #1482, opened 2022-11-23
fractions module

s-cork:
Implement fractions module in javascript with tests

required changes:
- adjust hashing of ints/floats
- sys.hash_info
- implement -float ** fractional power (comment out python 2 test that says this is not allowed)


bnmnetp:
@s-cork -- Any big risks with this one?   Its mostly a library function, and a few changes to core, which I don't follow 100%.

bnmnetp (approved):

PR #1505, opened 2023-06-08
Re - improve python to js regex

s-cork:
Obviously the first library I wanted to use the new re module on wouldn't compile
This PR fixes the issue and adds a test case in a `test_re_custom` file
Future issues like this can be added to that file.

### DIFF
--- a/src/lib/re.js
+++ b/src/lib/re.js
@@ -285,11 +285,22 @@ function $builtinmodule(name) {
         }
     })();
 
+    /*
+     * Python docs:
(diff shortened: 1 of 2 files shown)


meredydd (approved):
LGTM. Small change, full tests, could have self merged!

s-cork:
Was waiting the 48 hours 😉.

But yeah - probably a small enough change to just merge. But it's regex. And regex always makes my brain hurt. 

Thanks.

bnmnetp:
Maybe the 48 hours isn't necessary if its a bug fix PR?  I know it would drive me nuts to know that I had fixed an obvious bug but had to wait to merge.

PR #1507, opened 2023-06-09
fix some re compile errors

s-cork:
more fails from using it in the wild

bug fix and small change


s-cork:
merging as bug fix and small change

PR #1509, opened 2023-07-14
re: fix more javascript escape fails

s-cork:


s-cork:
small change - bug fix merging

PR #1513, opened 2023-09-21
property fixes

s-cork:
allows property to be subclassed correctly
and fixes doc issues from the test, uncommenting out some tests from the test suite


s-cork:
small change, enhancement with tests - merging

PR #1520, opened 2023-10-15
ffi: preserve arrays on javascript objects

s-cork:
Currently when we interop with javascript objects on window,
if we get an array, we return a python copy of this array

I've found this to be problematic since,
- we lose mutability
- some javascript libraries do crazy things with arrays
  (See below)

When returning an Array from a proxied javascript object
This PR instead returns a ProxyList
Which is very similar to a python list (it's even a subclass)
But, the internal Array is a Proxy of the original Javascript Array
Now we essentially have the same reference and mutating the ProxyList mutates the original list

I've added tests to demonstrate this interopability

---

Example in an app we had failing:

```javascript
// initial page script
function(){var analytics=window.analytics=window.analytics||[];//...}

```
The script used an array as a placeholder while the real `window.analytics` object was loaded.
It then preceded to do things like analytics.track=()=>{...}
But interop converted `window.analytics` to a pure python list so it wasn't possible to use the placeholder in python.
With this PR the `window.analytics` would be a `proxylist` and so accessing the `track` method in python would work.

---

Considerations

Why not two way?
i.e. sending a python list to a proxied javascript object should hold a reference to the original python list.
I'm happy to add support for this
But I felt it was less of a concern

### DIFF
--- a/src/ffi.js
+++ b/src/ffi.js
@@ -761,8 +761,6 @@ const arrayHandler = {
         if (rv === undefined && !(attr in target)) {
             return rv;
         }
-        console.log(attr, target);
-        debugger;
         // attributes on the list instance;
         return toPy(rv, pyHooks);
     },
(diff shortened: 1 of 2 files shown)


s-cork:
I think this is just about small enough to merge
Seems a sensible enhancement and should be backwards compatible
Spoke to @meredydd about the concept and he thought it was sensible
(merging)

PR #1534, opened 2023-11-29
functools: add cached_property implementation with tests

s-cork:
python 3.12 removed the lock implementation of cached_property, maybe that's why it didn't get added with the initial functools implementation.

Also adds support for assertRaisesRegex in unittest
which was simple now that the re module has been updated


s-cork:
new feature with tests, doesn't touch the core, small change, merging

PR #1537, opened 2024-01-21
ffi: fix converting cross origin windows to python

s-cork:
Advanced use case of the ffi module is to have an iframe in python and post messages to the iframe's `contentWindow`

When we access the `contentWindow` we try to convert this object to a python object.
But we break the browser's same origin policy when we access properties like `sk$object` or `$isPyWrapped` when determining what type of object we are converting to python.
https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#window

This PR adds a check in the Proxy object lookup to determine if the attribute is a cross origin window object
If it is, we shortcut the code and return a Proxy object rather than going through `toPy`.

I added tests that can be run in the browser
But these tests won't run in a node environment

### DIFF
--- a/src/ffi.js
+++ b/src/ffi.js
@@ -312,7 +312,7 @@ function toPyDict(obj, hooks) {
 
 function isCrossOriginWindow(obj) {
     // based on https://github.com/weizman/is-cross-origin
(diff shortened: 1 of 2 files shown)


s-cork:
small change - with tests - merging

PR #1539, opened 2024-02-08
fix: a couple of bugs in formatting with tests

s-cork:
We leak some errors in the implementation of `str.format` that should throw python errors.
the function assumes that there will be a valid value in the args or kws
but if there isn't, we end up trying to access internal skulpt attributes of undefined
rather than throwing KeyErrors/IndexErrors

This PR fixes those errors
and adds some relevant tests from cpython

It also adds:
- support for `!a` formatter
- error when mixing between automatic and manual numbering

And I hoist out the regex from the function as a performance improvement

### DIFF
--- a/src/formatting.js
+++ b/src/formatting.js
@@ -279,11 +279,14 @@ function formatString(format_spec) {
 };
 
 // str.format() implementation
+const isDigit = /^\d+$/;
+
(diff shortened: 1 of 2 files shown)


s-cork:
bug fix, relatively small change, with tests (merging)

PR #1541, opened 2024-02-24
bytes: support decode encodings that the browser supports #noa

s-cork:
while encoding in the browser is only valid for utf8,
the browser's TextDecoder supports various decode encodings

https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API/Encodings

This PR uses the browsers decode implementation if available
adds a couple of tests


s-cork:
This seems small enough to merge
doesn't break existing functionality, only enhances what we already have
has tests

PR #1552, opened 2024-06-05
Line numbers are invalid after a function call

s-cork:
This seems like an old piece of code that was superseded by other code
And doesn't cause issues generally since the line no is corrected when we visit the next expression

The test case is an edge case where a function call inside an f string exposes the incorrect line no

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -915,10 +915,7 @@ Compiler.prototype.vexpr = function (e, data, augvar, augsubs) {
         case Sk.astnodes.Compare:
             return this.ccompare(e);
         case Sk.astnodes.Call:
-            result = this.ccall(e);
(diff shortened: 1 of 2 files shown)


meredydd:
I trust the test, but can you briefly describe why annotateSource() doesn't do what the original author clearly expected it to here? (From the symptom I'm guessing it was somehow resetting the line number to one...)

meredydd (approved):

s-cork:
Closing this - looking more into the incorrect line numbers
It's because fstring parsing creates a new tokenizer/parser and never adjusts the starting lineno/column offset

This PR might cause a regression somewhere else
The work on the python 3.9 parser fixes this issue
Probably better to wait for that to eventually land, rather then plaster over the issue here

PR #1556, opened 2024-07-21
Unittest magic mock

Duncanr-glitch:
To be able to more effectively test web python projects with mocks to prevent running non-frontend code like server calls, I have made a MagicMock class and the necessary call class that handles many magic methods and all the assertions found in the standard library's MagicMock.


s-cork:
thanks for the contribution @Duncanr-glitch 

- how did you create the tests?
- how does this implementation differ from that of cpython?

For testing, it's always good to include tests from cpython where possible.
It's fine to comment out tests that are not appropriate/passable.
(let us know if you need help finding the test files in cpython source code)

Duncanr-glitch:
Thanks for pointing out the already built out std lib tests, I have added those along with my custom tests already there.

Here are the differences from the standard library:

- magic methods don't work when running type() on an instance of MagicMock
- No spec added -- I couldn't see what use it really had
- a number of private propertes, such as the _mock_children property, because I chose different implementation paths with the functionality being mostly the same
- __fspath__ isn't implemented since the os module is currently entirely unimplemented
- same with __aiter__ and asyncio
- errors with >/>=/</<= are from the MagicMock's side always (right or left); the standard library puts the error from the left hand side (but the mock_calls addition in both is from the MagicMock's side)
- __bool__ doesn't log it's calls to the parent MagicMock's mock_calls list because it causes infinite recursion (for the life of me I don't understand why, but if you see a solution I am very open to it.)
- The call object is merely a light wrapper of the standard python version without a lot of depth into matching all its functionality. It merely does enough to make the MagicMock function and be able to be debugged in an equivalent (though superficial) way. 
- This includes that there's an error when you try to get a random attribute on an uninitiated call object, but I didn't see this as worth the effor to fix.
- Also, my version of call can be initiated with _mock_name to instantiate it with that name, whereas the standard library passes that to the args of the call.

test/unit3/test_magic_mock.py
```diff
@@ -0,0 +1,352 @@
+__author__ = 'Duncan Richards'
+import unittest, math
+from unittest.mock import MagicMock, call, _magics
+
+
+
+class TestMockingMagicMethods(unittest.TestCase):
```
s-cork, line 7: Thanks for adding these extra tests

This seems to be a mixture of cpython tests and custom tests?

For tests here could we do the following:
- Take the `testsmagicmethods.py` file from cpython (which is the test file that this is based on)
- copy the file exactly (same name too)
- comment out tests that are not passing/not appropriate and make any other necessary tweaks

Create a separate file for custom tests, and name it appropriately e.g. `test_skulpt_magic_mock.py` (or can keep it in the same file but create a separate unittest.TestCase class)

This makes it easier to distinguish what is custom and what is from cpython for future developers.

doc/static
```diff
@@ -0,0 +1 @@
....getGreen,[t])),e.imagedata.data[r+2]=Sk.builtin.asnum$(Sk.misceval.callsimArray(t.getBlue,[t])),e.imagedata.data[r+3]=255,a(e,i,u)};n.set_pixel_at=new Sk.builtin.func(setpixelat),n.setPixelAt=new Sk.builtin.func(setpixelat);var updatepixel=function(e,n){var t,i,u;return Sk.builtin.pyCheckArgsLen(\"updatepixel\",arguments.length,2,2),t=Sk.builtin.asnum$(Sk.misceval.callsimArray(n.getX,[n])),i=Sk.builtin.asnum$(Sk.misceval.callsimArray(n.getY,[n])),l(e,t,i),u=4*i*e.width+4*t,e.imagedata.data[u]=Sk.builtin.asnum$(Sk.misceval.callsimArray(n.getRed,[n])),e.imagedata.data[u+1]=Sk.builtin.asnum$(Sk.misceval.callsimArray(n.getGreen,[n])),e.imagedata.data[u+2]=Sk.builtin.asnum$(Sk.misceval.callsimArray(n.getBlue,[n])),e.imagedata.data[u+3]=255,a(e,t,i)};n.update_pixel=new Sk.builtin.func(updatepixel),n.updatePixel=new Sk.builtin.func(updatepixel);var getheight=function(e){return Sk.builtin.pyCheckArgsLen(\"getheight\",arguments.length,1,1),new Sk.builtin.int_(e.height)};n.get_height=new Sk.builtin.func(getheight),n.getHeight=new Sk.builtin.func(getheight);var getwidth=function(e,n){return Sk.builtin.pyCheckArgsLen(\"getwidth\",arguments.length,1,1),new Sk.builtin.int_(e.width)};n.get_width
```
s-cork, line 1: can we remove this file from the PR

s-cork:
Thanks for cleaning up the tests.

When i run test_skulpt_magic_methods in cpython i get 2 failures

```
FAIL: test_mock_calls_propagation (__main__.TestMagicMockBasics)
----------------------------------------------------------------------
  File "/Users/scork/Desktop/Projects/anvil/foo.py", line 62, in test_mock_calls_propagation
    self.assertEqual(parent_mock.mock_calls, [call(1, 2, _mock_name="a", key='value')])
AssertionError: [call.a(1, 2, key='value')] != [call(1, 2, _mock_name='a', key='value')]

======================================================================
FAIL: test_return_side_effect (__main__.TestMagicMockBasics)
----------------------------------------------------------------------
  File "/Users/scork/Desktop/Projects/anvil/foo.py", line 27, in test_return_side_effect
    self.assertEqual(mock(), [1, 2, 3])
AssertionError: 1 != [1, 2, 3]

----------------------------------------------------------------------
Ran 2 tests in 0.001s
```

Duncanr-glitch:
> Thanks for cleaning up the tests.
> 
> When i run test_skulpt_magic_methods in cpython i get 2 failures
> 
> ```
> FAIL: test_mock_calls_propagation (__main__.TestMagicMockBasics)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/Users/scork/Desktop/Projects/anvil/foo.py", line 62, in test_mock_calls_propagation
>     self.assertEqual(parent_mock.mock_calls, [call(1, 2, _mock_name="a", key='value')])
> AssertionError: [call.a(1, 2, key='value')] != [call(1, 2, _mock_name='a', key='value')]
> 
> ======================================================================
> FAIL: test_return_side_effect (__main__.TestMagicMockBasics)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/Users/scork/Desktop/Projects/anvil/foo.py", line 27, in test_return_side_effect
>     self.assertEqual(mock(), [1, 2, 3])
> AssertionError: 1 != [1, 2, 3]
> 
> ----------------------------------------------------------------------
> Ran 2 tests in 0.001s
> ```

It passes in skulpt though right? I'm not sure what the issue is nor the resolution path tbh.

s-cork:
Because the tests are not from CPython it's important to also test against CPython. 

It suggests that the implementation doesn't match CPython's expected implementation. It looks like the implementation of side effect is not correct when given an iterable. The docs aren't very clear on this, but there is an example where it'll iterate through the iterable when being called. It might be certain types of iterable, so worth fixing the implementation here. And I'm not sure about the other test, but it would be worth working out what's going on. If it's not important then we don't need to test it. 

But it should generally be the case that running custom tests in CPython will pass.

PR #1564, opened 2024-10-29
Feat: add lineno, col_offset support for alias

s-cork:
This was added in python 3.10
I overrode the astnodes.js file manually
I think that's fine for now

Reason for the change, if we know the alias position then it makes a jump to definition implementation easier from the ast.


s-cork:
small change - merging

PR #1567, opened 2024-11-29
Logging.py

s-cork:
Adds logging support

mostly from the cpython library with some adaptions to skulpt

Not supported:
- String templates style support
- No error checking in string format style
- No error handling support
- No warning support
- No file handler support
- no logging.config module or logging.handlers module

adds sys.stderr - which is the same as sys.stdout


bnmnetp (approved):
These seem very straightforward, and self contained.  Very mergeable.

PR #1572, opened 2025-02-11
Intern compiled strings

s-cork:
builds on top of #1571 

(see last commit)

I mistakenly thought that compiled strings were already interned
but actually this happens on a per scope basis
I think it's probably worth including this PR so that something like this works

```python
def check(x):
    for x in ("123", "abc"):
        print(x is y)

check("123") # True, False
check("abc") # False, True
```

In Skulpt, after #1481  we get False values because constants in different scopes are not identical
But in CPython we get True values for str literals interned at compile time

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -951,7 +951,7 @@ Compiler.prototype.vexpr = function (e, data, augvar, augsubs) {
             }
             // else fall through and make a string instead
         case Sk.astnodes.Str:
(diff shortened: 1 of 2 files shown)


meredydd (approved):
This looks fine to me.

I'm not particularly sure why all the single-character strings should also be singled out for interning, but I have no objection to it.

PR #1580, opened 2025-05-09
Ensure we check the value is true when doing ffi.toPy

s-cork:
amendment to #1579


s-cork:
small change and small fix

PR #1585, opened 2025-07-28
Improve behavior of input() when using non-default sys.stdin

skiadas:
Currently input() is behaving in a non-expected manner in two ways:

- If the `input` function takes a prompt, then `Sk.builtin.file.$readline` is hard-coded to be used, even if `sys.stdin` has been replaced by a `StringIO` object. This function expects the internal structure of a `file` and results in errors if the `sys.stdin` has been redirected to a `StringIO`.
- When `input` uses `readline` to read, it does not remove the possible newline character at the end of the line. This results in somewhat inaccurate behavior in multi-line input cases.

This PR addresses these issues by only using `$readline` in the case where `sys.stdin` has not been replaced, and by trimming the newline character if present on the normal `readline` call.

### DIFF
--- a/src/builtin.js
+++ b/src/builtin.js
@@ -711,7 +711,8 @@ Sk.builtin.raw_input = function (prompt) {
                     return Sk.misceval.callsimOrSuspendArray(sys["$d"]["stdout"]["write"], [sys["$d"]["stdout"], new Sk.builtin.str(lprompt)]);
                 },
                 function () {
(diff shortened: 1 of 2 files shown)


bnmnetp:
@s-cork I have merged this into my fork.  Happy for any feedback you may have. Otherwise I think this can be merged.

s-cork:
There's one suspension chaining issue, if the file write suspends on line 714, then we'll be trying to convert a suspension into a string. I should be able to push changes to the branch so I'll do that, have a play, and then merge.

s-cork (approved):

PR #1592, opened 2026-01-08
Feat: decimal module

s-cork:
Partial implementation of the decimal module with tests

This is a branch i had around for a long time
in the penultimate commit this where i used Opus 4.5 to help me finish of something that's been hanging around for long time


bnmnetp:
This looks fine to me.  As a library module I think you are good to merge whenever.

PR #1599, opened 2026-06-16
Fix: Free variables in class bodies

### LINKED ISSUE #1178: scoping of classes within functions
examples
```
def f():
    x = 'bar'
    class A: 
        foo = x
f()
```

It get's compiled as `$loc.foo = $free.x;`  
but `x` is not a `$free` it's part of the `$cell` (I think)
and so gets assigned to `undefined`.

My knowledge of the compiler is not too strong so I'm not quite sure how to go about tackling this one... 

it comes up in this test in skulpt - and i've come across it before in the wild
https://github.com/skulpt/skulpt/blob/3155b77c6c88806061ebb9592cf60a7a2ed12979/test/unit/test_dict.py#L116

bennorth:
While working on something involving descriptors, I came across a problem which I boiled down to the behaviour of this code:

```python
def f():
    s = "hello"
    class _inner:
        print(s.upper())

f()
```

In CPython this prints `HELLO` as expected.  In Skulpt it generates a JavaScript error:

```test
ExternalError: TypeError: Cannot read properties of undefined (reading 'tp$getattr') on line 4
```

I used Claude to help investigate, and refined what it produced into this PR.  The root cause was that cell variables weren't being made available to the `class` body evaluation.

With the PR, all tests pass, including some new ones (`test/unit3/test_unbound_free.py`) which fail under current `master` Skulpt.  The comments are mostly Claude's and seem reasonable to me, but let me know if they should be made briefer.

Thanks!

Close #1178

### DIFF
--- a/src/compile.js
+++ b/src/compile.js
@@ -2832,6 +2832,22 @@ Compiler.prototype.nameop = function (name, ctx, dataToStore) {
         case OP_DEREF:
             switch (ctx) {
                 case Sk.astnodes.Load:
+                    // The cell may exist but hold no value yet (e.g. a free variable
+                    // read before the enclosing binding has executed). A bound Python
+                    // value is never JS undefined (None is Sk.builtin.none.none$), so
+                    // an undefined slot means the name is unbound. Match CPython: a free
+                    // variable raises NameError, a local cell raises UnboundLocalError.
+                    out("if (", dict, ".", mangledNoPre, "===undefined) {");
+                    if (scope === Sk.SYMTAB_CONSTS.FREE) {
+                        out("throw new Sk.builtin.NameError(\"cannot access free variable '",
+                            mangledNoPre,
+                            "' where it is not associated with a value in enclosing scope\");");
+                    } else {
+                        out("throw new Sk.builtin.UnboundLocalError(\"cannot access local variable '",
+                            mangledNoPre,
+                            "' where it is not associated with a value\");");
+                    }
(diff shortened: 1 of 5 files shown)


s-cork:
Hi @bennorth  i think it's also fixed as part of #1591 

can you check that branch works for you?

bennorth:
Behaviour of the new tests under the different versions of the code as follows.  All tests pass under this PR's code.

| test | master | 1591 |
|--------|---------|-----------|
| test_class_captures_function_local | FAIL | ok |
| test_deep_class_captures_function_locals | FAIL | FAIL |
| test_class_body_binding_not_visible_to_nested_class | ok | ok |
| test_free_var_read_during_class_construction | FAIL | FAIL |
| test_cell_relayed_through_intermediate_functions | ok | ok |
| test_no_binding_falls_through_to_global | ok | ok |
| test_free_var_read_before_binding | FAIL | FAIL |
| test_local_cell_read_before_binding | FAIL | FAIL |

So one test goes from failing to passing in #1591.

s-cork (approved):
LGTM thanks @bennorth

bennorth:
Thanks for linking #1178.  I should have looked for existing issues when creating this PR.

s-cork:
Went through this with codex astra
added a couple of additional tests that failed, plus the fixes
merging

bennorth:
Thanks for catching that additional situation.

